PR #256, opened 2023-04-21
Propagate exceptions

kumaranvpl:


fastkafka/_application/app.py
```diff
@@ -579,6 +584,9 @@ async def _shutdown_consumers(
 ) -> None:
     if self._kafka_consumer_tasks:
         await asyncio.wait(self._kafka_consumer_tasks)
+        for task in self._kafka_consumer_tasks:
+            if self.test_mode and task.exception():
```
abodacs, line 588: @kumaranvpl 
Thanks for the PR.
I think we can move the if condition for test_mode before the for loop

davorrunje:
We refactored heavily stream processing loop in releases 0.6.0. One of the supported features is exception handling and this is no longer needed.

PR #431, opened 2023-07-07
Use create_subprocess_exec to execute commands instead of multiple options

kumaranvpl:


davorrunje:
what is this?

kumaranvpl:
@davorrunje Nevermind. This is not needed.

davorrunje:
Still, what is it?

PR #447, opened 2023-07-20
Draft: 442 fix semgrep issues at githubworkflows

kumaranvpl:


davorrunje:
Are you sure you want to do this? Isn't it better just to use comments to ignore these warnings.

PR #484, opened 2023-08-29
feat: add AsyncAPI endpoints to FastAPI plugin

Lancetnik:
There two features in this PR

The first one can be described by this test: https://github.com/airtai/fastkafka/blob/FastAPI-AsyncAPI/tests/asyncapi/base/fastapi.py#L17

This is the FastAPI compatible way to setup AsyncAPI information with the FastStream plugin. Also, it adds 3 routes `/asyncapi`, `/asyncapi.json` and `/asyncapi.yaml` to interact with the AsyncAPI scheme right from the FastAPI application

The next feature is much simpler custom decoder/parser support

For now, the following examples are valid decoders/encoders too

```python
def decode(msg):
    ...

async def decode(msg):
    ...

async def decode(msg, original):
    return await original(msg)
```

In the old version, only the last one was valid function

### DIFF
--- a/faststream/_compat.py
+++ b/faststream/_compat.py
@@ -13,6 +13,7 @@ from pydantic import BaseModel
 #     from typing_extensions import override as override
 # else:
 #     from typing import override
+from typing_extensions import Required as Required
 from typing_extensions import TypedDict as TypedDict
 from typing_extensions import override as override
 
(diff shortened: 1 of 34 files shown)


Lancetnik:
Don't merge, asyncapi information storage changes required

davorrunje (approved):

PR #498, opened 2023-09-01
feat: add FastAPI Router include_router method, fix #497

### LINKED ISSUE #497: Producer not working if there is a subscriber with batch set to True on the same topic
Why is this failing?

```python
from typing import List

from faststream.kafka import KafkaBroker, TestKafkaBroker

app = FastStream(broker)


@broker.subscriber("test")
async def handle(msg: str, logger: Logger):
    logger.info(msg)

@broker.subscriber("test_batch", batch=True)
async def handle_batch(msg: List[str], logger: Logger):
    logger.info(msg)

import pytest

@pytest.mark.asyncio
async def test_me():
    async with TestKafkaBroker(broker) as test_broker:
        # This works
        await test_broker.publish("123", "test")

        # why is this failing?
        await test_broker.publish("I", "test_batch")

```

Lancetnik:
Support FastAPI routers nesting:

```python
router = self.router_class()
router2 = self.router_class()
router.broker = self.broker_test(router.broker)

app = FastAPI(lifespan=router.lifespan_context)

@router.subscriber(queue)
async def hello():
    return "hi"

@router2.subscriber(queue + "1")
async def hello_router2():
    return "hi"

router.include_router(router2)
```

Fix `TestKafkaBroker.publish` and `publish_batch` methods to support batch consumers

### DIFF
--- a/faststream/broker/fastapi/router.py
+++ b/faststream/broker/fastapi/router.py
@@ -1,4 +1,5 @@
 import json
+from abc import abstractmethod
 from contextlib import asynccontextmanager
 from enum import Enum
 from typing import (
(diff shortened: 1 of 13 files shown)


Sternakt:
Closes #497

PR #563, opened 2023-09-06
docs: create documentation navigation structure and some examples

### LINKED ISSUE #511: Side effects in pytest
When we change the order of test in pyproject.toml, some tests fail with weird errors:

```
E       pydantic_core._pydantic_core.ValidationError: 1 validation error for on_input_data
E       logger
E         Field required [type=missing, input_value={'msg': {'data': 0.2}}, input_type=dict]
E           For further information visit https://errors.pydantic.dev/2.3/v/missing

/work/davor/projects/airt/fastkafka/.venv/lib/python3.11/site-packages/fast_depends/core/model.py:202: ValidationError

```
Change:
```
testpaths = [
    "docs_src",
    "tests",
]
```
to 
```
testpaths = [
    "tests",
    "docs_src",
]
```

Lancetnik:
We need to refactor AsyncAPI export guide as a part of AsyncAPI articles or include it in the new schema
close #511


Lancetnik:
All examples are checked manually, will create a tests tomorrow

Lancetnik:
1) create the documentation plan in the `mkdocs.yaml`
2) create some examples for this plan in the `docs/docs_src/getting_started`
3) fix some TestClient errors
4) remove all tests from `docs_src` to `tests/doc` directory

Lancetnik:
Please, use only the `tests` directory to write tests sunce this PR
Also, try to not move working directory with `os.chdir` and use long sleeps
Mark all slow tests by `pytest.mark.slow` and tests require real kafka connection by `pytest.mark.kafka`

docs/docs_src/kafka/testing_with_real_broker/test_with_real_broker.py
```diff
@@ -16,7 +14,8 @@ async def on_output_data(msg: Data):
     async with TestKafkaBroker(broker, with_real=True) as tester:
         await tester.publish(Data(data=0.2), "input_data")
 
-        await asyncio.sleep(2.0)
+        await on_input_data.wait_call(3)
```
davorrunje, line 17: There is no documentation for wait_call() in Propan. Why do we have this function?

faststream/kafka/producer.py
```diff
@@ -35,16 +35,12 @@ async def publish(
 
         headers_to_send = {
             "content-type": content_type or "",
+            "correlation_id": correlation_id or str(uuid4()),
```
davorrunje, line 38: correlation_id is still in the signature of the publish(), but it is not used anywhere.

faststream/kafka/test.py
```diff
@@ -320,6 +320,7 @@ async def _fake_close(
         p.mock.reset_mock()
         if getattr(p, "_fake_handler", False):
             self.handlers.pop(p.topic, None)
+            p._fake_handler = False
```
davorrunje, line 323: what is this change? is this to make it reenterent?

davorrunje (approved):

docs/docs_src/kafka/testing_with_real_broker/test_with_real_broker.py
Lancetnik: It's a testing purposes function added in the FastStream
But, if you are using TestClient, you need no wait any time, publishing working synchronous with consuming this case. So, It is a function helpful in the real broker tests

davorrunje: It was not documented anywhere and it is very useful. We have to add it to docs about testing with real broker.

faststream/kafka/test.py
Lancetnik: It fixes the part of re-enter problems, but not all of them

davorrunje: Can you write tests that fail on re-entrance? We can skip them for now and include them later when we fix the problem.

faststream/kafka/producer.py
davorrunje: I was just wondering if that was a bug. The parameter is specified in the signature of the function and it is not used anywhere in the body.

PR #666, opened 2023-09-17
Fix gen

Lancetnik:
Should fix CI


harishmohanraj (approved):

davorrunje:
These files are autogenerated and used only for code generation. They will be overwritten by docs.py script anyway.

PR #673, opened 2023-09-18
Fix docs img

Sternakt:


Lancetnik:
Should we create a new release with all minor changes (even in the docs)?

Sternakt:
> Should we create a new release with all minor changes (even in the docs)?

This is pretty visible on pypi, and I had to release to test the fix, hence this release candidate
I think it will not be necessary to have release candidates for such minor issues in the future

The docs don't need a release, they are updated when merged into main

davorrunje:
We will probably make a new release today or tomorrow with a few fixes. Frequent releases are common with new packages, later we'll stabilize the release schedules.

davorrunje (approved):

PR #700, opened 2023-09-21
fix: correct StreamRouter broker annotation

Lancetnik:
Now `StreamRouter().broker` return abstract `AsyncBrokerUsecase` annotation, so all type hints and code suggestions are broken. This is the fix for it.
Version is increased to deliver this change to users.


faststream/__about__.py
```diff
@@ -1,5 +1,5 @@
 """Simple and fast framework to create message brokers based microservices"""
-__version__ = "0.1.1"
+__version__ = "0.1.2"
```
davorrunje: we are releasing 0.1.1 today

davorrunje (approved):

PR #722, opened 2023-09-23
fix: correct top-level errors logging

Lancetnik:


faststream/broker/handler.py
```diff
@@ -317,12 +310,6 @@ async def consume(self, msg: MsgType) -> SendableMessage:  # type: ignore[overri
                         await self.close()
                         return None
 
-                    except Exception as e:
```
Lancetnik, line 320: For now this exception hides all Exceptions, but we need to catch them in the CriticalLogMiddleware

Lancetnik:
replaced by #723

PR #825, opened 2023-10-10
chore: split real broker tests to independant runs

Lancetnik:
# Description

Much faster CI runs and less random erros occures


Lancetnik:
@davorrunje can u merge it, plz?

davorrunje (approved):

Issue #811, opened 2023-10-06
Documentation: add custom AsyncAPI information


Lancetnik:
1) description can be Markdown text, replace and show example with this feature
2) add `consumer(title=...)` and `producer(title=...)` examples
3) add `producer(schema=...)` example
4) consumer function docstring can be used as a description too
5) add query params description to [hosting page](https://faststream.airt.ai/0.2/getting-started/asyncapi/hosting/)
Replace all AsyncAPI example screenshots after #735
6) `asyncapi_url` argument description

Lancetnik:
After #800 and #735 merged

PR #903, opened 2023-11-01
feat: add `asyncapi_url` broker arg

### LINKED ISSUE #578: Support security in RabbitBroker
- [x] Add parsing for security classes in rabbit/security.py
- [x] Pass parsed attrs to rabbit.connect and init call
- [x] Prepare tests for passing of attrs
- [x] Fix rabbit/fastapi.pyi and rabbit broker/pyi annotations
- [x] Add AsyncAPI schema generation tests

### LINKED ISSUE #895: Feature: schema generate enable connection arguments rather than connection string url
# Current behaviour

At the time of writing (faststream==0.2.8) the RabbitBroker constructor signature reads:

```python
 def __init__(
        self,
        url: Union[str, URL, None] = "amqp://guest:guest@localhost:5672/",
        *,
        virtualhost: str = "/",
        max_consumers: Optional[int] = None,
        protocol: str = "amqp",
        protocol_version: Optional[str] = "0.9.1",
        security: Optional[BaseSecurity] = None,
        **kwargs: Any,
    )
```

which implies that an empty url is allowed. `**kwargs` arguments are correctly passed down to `connect_robust`:

```python
async def connect_robust(
    url: Union[str, URL, None] = None,
    *,
    host: str = "localhost",
    port: int = 5672,
    login: str = "guest",
    password: str = "guest",
    virtualhost: str = "/",
    ssl: bool = False,
    loop: Optional[asyncio.AbstractEventLoop] = None,
    ssl_options: Optional[SSLOptions] = None,
    ssl_context: Optional[SSLContext] = None,
    timeout: TimeoutType = None,
    client_properties: Optional[FieldTable] = None,
    connection_class: Type[AbstractRobustConnection] = RobustConnection,
    **kwargs: Any,
)
```
Thus we can use ``**kwargs``rather than the url to provide connection parameters. This works all fine. However, `faststream/asyncapi/generate.py:get_app_broker_server` will fail at line 140:

```python
elif len(broker.url) == 1:
```
because broker.url is `None`. 

# Desired behaviour

It is understood that every underlying broker technology uses different **kwargs, to build their internal actual connection string. The issue here is, that the url used for the schema generation is not exactly the same thing as the connection string. I actually ran into this issue because I did not want to display the connection string password in the asyncapi schema and decided to use **kwargs for providing the connection parameters. It is definitely not desired to display the full connection string in the api documentation.

One way to get around this issue is the introduce a new argument to the Broker constructor:
```python
server_url: list[str]|str|None 
```
used in in `get_app_broker_server` in case no url is specified.

Another way is to simply not display any server in the schema if url is None.

Yet another idea is to render the connection without username:password and parameters.

A combination of these approaches would also be feasible an fit all needs of customizability.

# Current workaround

Here is a monkey patch resolving the issue for RabbitBroker:

```python
def get_app_broker_server(*argv, **kwargs):
    url = make_url(
        login="****",
        password="****",
        host=args["host"],
        port=args["port"],
        virtualhost=args["virtualhost"],
        ssl=args["ssl"],
    )
    return {
        "development": faststream.asyncapi.generate.Server(
            url=str(url),
            protocol=url.scheme,
            protocolVersion=args["protocol_version"],
        )
    }

setattr(
    faststream.asyncapi.generate, "get_app_broker_server", get_app_broker_server
)
broker = faststream.rabbit.fastapi.RabbitBroker(**args)

```

Lancetnik:
# Description

Now user can overrides the original broker url in AsyncAPI schema only with `asyncapi_url` broker argument

It can be extremelly hepful to hide connection credentials in a public documentation

```python
broker = RabbitBroker(
    "amqp://guest:guest@localhost:5672/",
    asyncapi_url="amqp://****:****@localhost:5672/",
)
```

Also, in the **RMQ** case, keyword arguments are overrides the original url, so

```python
broker = RabbitBroker(
    "amqp://guest:guest@localhost:5672/",
    host="192.168.1.1",
)

# amqp://guest:guest@192.168.1.1:5672/ - the final url
```

Also, **RMQ** AsyncAPI protocol now defines automatically from "amqp"/"amqps"

Also, **RMQ** supports `BaseSecurity` and `SASLPlaintext` as a `security` option for now

Also, mv `faststream.broker.security` to `faststream.security` with deprecation notice

Fixes #895, Close #578

## Type of change

## Checklist


Lancetnik:
@kumaranvpl can you please use the Issue description for detail release notes?
Also, we need to add `asyncapi_url` to AsyncAPI custom docs in the #811 scope

davorrunje (approved):

Lancetnik:
@kumaranvpl can you please use the PR description to update release note by some details? Peoples should know about changes explicitly, not only by PRs list

PR #1413, opened 2024-05-01
Fixes  #1412 with `TestKafkaBroker` behaviour where Consumer Groups weren't being respected

### LINKED ISSUE #1412: Kafka Test Subscriber doesn't match Consumer Group behaviour
Currently the Kafka Test Publisher / Subscriber combo returns early upon hitting a subscriber / topic combination.

If an application has two subscribers, across two `consumer-group`s listening on a single topic, FastStream does not respect that both subscribers should receive a message each from the topic. In testing only the first-matched topic would run.

https://github.com/airtai/faststream/blob/ed270add879eb48f5a0f6ce07564ab5625be5049/faststream/kafka/testing.py#L94-L102

Removing the `return` and letting each Subscriber handle the alert seems to get me closer to working behaviour (essentially we will have to create a list of subscribers attached to consumer groups and run `handle` on the first one in each consumer group – not sure how filters should come into play here).

Will work on a PR soon ☺️

sifex:
# Description

This fixes bug #1412 where distribution of messages on a topic does not get distributed to each consumer group when running the `TestKafkaBroker`. 

Fixes #1412 

**Example:** <br/>This example should have 2 messages received to 2 subscribers.

```python
@test_broker.subscriber(queue, group_id="group1")
async def subscriber1(): ...

@test_broker.subscriber(queue, group_id="group2")
async def subscriber2(): ...

assert subscriber1.mock.call_count == 1
assert subscriber2.mock.call_count == 1
```

Where as if we have the same group_id, we only expect to receive it once.

```python
@test_broker.subscriber(queue, group_id="same_group")
async def subscriber1(): ...

@test_broker.subscriber(queue, group_id="same_group")
async def subscriber2(): ...

assert subscriber1.mock.call_count == 1
assert subscriber2.mock.call_count == 0
```

## Other fixes

- It also fixes a small bug in `DOCKER_COMPOSE_PROJECT` project name generation where the users' username field contains a `.` character (which docker believes to be invalid naming).
- Typo fix in `.github/PULL_REQUEST_TEMPLATE.md`

## Type of change

## Checklist

- [ ] Both new and existing unit tests pass successfully on my local environment by running `scripts/test-cov.sh` **(No. Currently Segfaults on Local M1 MBP, will rely on GHA)**


sifex:
Currently facing an issue where the RPC test assumes you can get a response back from a publisher? Not sure why this assumption was made.

```python
@pytest.mark.asyncio()
    async def test_rpc(self, queue: str, rpc_broker: BrokerUsecase):
        @rpc_broker.subscriber(queue)
        async def m(m):  # pragma: no cover
            return "1"
    
        async with rpc_broker:
            await rpc_broker.start()
            r = await rpc_broker.publish("hello", queue, rpc_timeout=3, rpc=True)
    
>       assert r == "1"
E       AssertionError

```

Lancetnik:
> Currently facing an issue where the RPC test assumes you can get a response back from a publisher? Not sure why this assumption was made.
> 
> ```python
> @pytest.mark.asyncio()
>     async def test_rpc(self, queue: str, rpc_broker: BrokerUsecase):
>         @rpc_broker.subscriber(queue)
>         async def m(m):  # pragma: no cover
>             return "1"
>     
>         async with rpc_broker:
>             await rpc_broker.start()
>             r = await rpc_broker.publish("hello", queue, rpc_timeout=3, rpc=True)
>     
> >       assert r == "1"
> E       AssertionError
> ```
> 
> imo I think publishers by definition can't return anything other than a ReceiveAck / similar response from the message queue – also not sure if I misunderstand the use case for RPC in this context.

RPC in TestClient allows you to validate your handler result without special publisher creation. I think, we should save this behavior. In you case, if user has the only last result - it should be OK for most cases: (the following pseudocode represents assuming bahavior)

```python
result = None
for sub in subscribers:
    result = call_subscriber(...)
return result
```

faststream/kafka/testing.py
```diff
@@ -91,17 +91,21 @@ async def publish(  # type: ignore[override]
...
-                    handler=handler,
-                    message=[incoming]
-                    if isinstance(handler, AsyncAPIBatchSubscriber)
-                    else incoming,
-                    rpc=rpc,
-                    rpc_timeout=rpc_timeout,
-                    raise_timeout=raise_timeout,
-                )
+        for consumer_group in set(
+            {sub.group_id for sub in self.broker._subscribers.values()}
```
Lancetnik, line 95: I don't like O(n**2) in this case. Can we find any decision without iteration throw all subscribers twice?

Lancetnik (changes requested):
Can you please fix the comments and upload main changes to make it mergeable?

sifex:
Should be good to go, applied ruff formatting + linting.

Lancetnik:
I'll check it and solve conflicts with main soon. Thank you for the work!

Lancetnik (dismissed):

Lancetnik (changes requested):
@sifex looks really great, but I now can't understand, why it respects consumer group😆 

Anyway, can you, please add test for this behavior for `publish_batch` method? Then we can merge it

Lancetnik (approved):

PR #1484, opened 2024-05-29
Resolve Issue 1386, Add rpc_prefix

### LINKED ISSUE #1386: Feature: more stronger unique RPC subject names
We should use smth more stronger, than just `str(uuid4())` to generate unique reply-to topic for RPC requests in Redis (and other future brokers): https://github.com/airtai/faststream/blob/main/faststream/redis/publisher/producer.py#L71

I suggest to copy `nats.NUID` class to **FastStream** and use it like in the NatsBroker do: https://github.com/nats-io/nats.py/blob/main/nats/nuid.py

aKardasz:
# Description

Adds nuid from nats broker to generation unique ids.
Adds rpc_prefix, to allow support for nats streams inbox_. functionality via permissions, for consistency added the same rpc_prefix to the redis implementation. 

Fixes #1386 

## Type of change

- [X] New feature (a non-breaking change that adds functionality)

## Checklist

- [X] My code adheres to the style guidelines of this project (`scripts/lint.sh` shows no errors)

### DIFF
--- a/faststream/nats/publisher/producer.py
+++ b/faststream/nats/publisher/producer.py
@@ -73,7 +73,6 @@ class NatsFastProducer(ProducerProto):
 
             reply_to = client.new_inbox()
 
-
             future: asyncio.Future[Msg] = asyncio.Future()
             sub = await client.subscribe(reply_to, future=future, max_msgs=1)
             await sub.unsubscribe(limit=1)
@@ -133,7 +132,6 @@ class NatsJSFastProducer(ProducerProto):
         stream: Optional[str] = None,
         timeout: Optional[float] = None,
         rpc: bool = False,
-        rpc_prefix: str = "",
         rpc_timeout: Optional[float] = 30.0,
         raise_timeout: bool = False,
     ) -> Optional[Any]:
@@ -148,9 +146,7 @@ class NatsJSFastProducer(ProducerProto):
         if rpc:
             if reply_to:
                 raise WRONG_PUBLISH_ARGS
-            nuid = NUID()
-            rpc_nuid = str(nuid.next(), "utf-8")
-            reply_to = f"{rpc_prefix}{rpc_nuid}"
+            reply_to = self._connection._nc.new_inbox()
             future: asyncio.Future[Msg] = asyncio.Future()
             sub = await self._connection._nc.subscribe(
                 reply_to, future=future, max_msgs=1
--- a/faststream/nats/publisher/producer.py
+++ b/faststream/nats/publisher/producer.py
@@ -135,7 +135,7 @@ class NatsJSFastProducer(ProducerProto):
         stream: Optional[str] = None,
         timeout: Optional[float] = None,
         rpc: bool = False,
-        reply_to_prefix: str = "",
+        rpc_prefix: str = "",
         rpc_timeout: Optional[float] = 30.0,
         raise_timeout: bool = False,
     ) -> Optional[Any]:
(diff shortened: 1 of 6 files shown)


aKardasz:
Statis analysis error:

```

Test results:
>> Issue: [B311:blacklist] Standard pseudo-random generators are not suitable for security/cryptographic purposes.
   Severity: Low   Confidence: High
   CWE: CWE-330 (https://cwe.mitre.org/data/definitions/330.html)
   More Info: https://bandit.readthedocs.io/en/1.7.8/blacklists/blacklist_calls.html#b311-random
   Location: faststream/utils/nuid.py:40:22
39          def __init__(self) -> None:
40              self._prand = Random(randbelow(max_int))
41              self._seq = self._prand.randint(0, MAX_SEQ)

--------------------------------------------------

Code scanned:
        Total lines of code: 34801
        Total lines skipped (#nosec): 0

Run metrics:
        Total issues (by severity):
                Undefined: 0
                Low: 1
                Medium: 0
                High: 0
        Total issues (by confidence):
                Undefined: 0
                Low: 0
                Medium: 0
                High: 1
Files skipped (0):
```

Lancetnik:
We can no worry about this warning due we doesn't use random for cryptography
You can just suppress it

faststream/nats/publisher/producer.py
```diff
@@ -134,6 +134,7 @@ async def publish(  # type: ignore[override]
         stream: Optional[str] = None,
         timeout: Optional[float] = None,
         rpc: bool = False,
+        rpc_prefix: str = "",
```
Lancetnik, line 137: Can we do not add an extra parameter, but reuse this nats-py code: https://github.com/nats-io/nats.py/blob/main/nats/aio/client.py#L1041 ?

Looks, like user should setup this option via `NatsBroker(inbox_prefix=...)` now and I want to save this behavior

Lancetnik (changes requested):
I don't like an idea to add an extra option if nats-py already has the mechanism we can reuse

faststream/nats/publisher/producer.py
aKardasz: I did it this way because your original implementation doesn't use nats inbox creation, so to keep true to that I didn't want to change it up. when you use inbox the reply_to will be some form of _inbox.<id> when you use the broker inbox prefix, which can be passed in if I am not mistaken in a few spots, you get <prefix>_inbox.<id>, while your reply to uses only an id, which doesn't have enough consistency for nats permissions setup. To make this functionality optional is why I used rpc_prefix. Hopefully that explains, otherwise you will want to use nats features all the way through ( in essence I am making an rpc prefix and not broker inbox prefix as they don't necessarily work the same)

aKardasz: To add more info, I would replace the rpc id generation with with " nats.NewInbox() "

aKardasz:
@Lancetnik with the removal of rpc_prefix, realistically that would mean no new tests are needed, would that be a safe assumption since the inbox update would just fall under regular tests.

Also, having a hard time running the tests in github workspaces, maybe we can do a setup guide for others or like a make file what are your thoughts?

Lancetnik:
@aKardasz I think, we should update CONTRIBUTING file to allow users to contribute easily

About testing - I think we should check `NatsBroker(inbox_prefix=...)` correct behavior

Can you please add any test for it and then we can merge the branch?

davorrunje:
> Also, having a hard time running the tests in github workspaces, maybe we can do a setup guide for others or like a make file what are your thoughts?

We could create devcontainers with the complete setup for testing particular MQ. That would probably be the best solution.

aKardasz:
> @aKardasz I think, we should update CONTRIBUTING file to allow users to contribute easily
> 
> About testing - I think we should check `NatsBroker(inbox_prefix=...)` correct behavior
> 
> Can you please add any test for it and then we can merge the branch?

Let me know if the added test is sufficient.

Lancetnik (approved):

PR #1592, opened 2024-07-11
Feat: init ping method

### LINKED ISSUE #1582: Feature: add `broker.ping()` method
We should add a mechanism to check real connection to broker liveness as a step of #1181 impl

First of all, we should add abstract method to [`BrokerUsecase` class](https://github.com/airtai/faststream/blob/main/faststream/broker/core/usecase.py#L48)

Then we should add this method impl to **KafkaBroker**/**RabbitBroker**/etc (using `self._connection` or `self._producer` object)

Also, we should add tests for that functional smth there: https://github.com/airtai/faststream/blob/main/tests/brokers/base/connection.py

Flosckow:
# Description

Close #1582

## Type of change

## Checklist


faststream/broker/core/usecase.py
```diff
@@ -342,3 +342,8 @@ async def publish(
             publish = partial(m(None).publish_scope, publish)
 
         return await publish(msg, **kwargs)
+    
+    @abstractmethod
+    async def ping(self, timeout: int | None) -> bool:
```
Lancetnik, line 347: `int | None` is not compatible with lt python3.10. Use `Optional[int]` instead (btw, should it be `float` instead of `int`?)

faststream/rabbit/broker/broker.py
```diff
@@ -655,3 +655,13 @@ async def declare_exchange(
         """Declares exchange object in **RabbitMQ**."""
         assert self.declarer, NOT_CONNECTED_YET  # nosec B101
         return await self.declarer.declare_exchange(exchange)
+    
+    async def ping(self, timeout: int | None) -> bool:
+        await super().ping(timeout)
```
Lancetnik, line 660: Why are u calling super here? It raises `NotImplementedError`

faststream/rabbit/broker/broker.py
```diff
@@ -655,3 +655,13 @@ async def declare_exchange(
...
+        try:
+            channel = await self._connection.channel()
+            await channel.declare_queue(name='test_queue', durable=True)
```
Lancetnik, line 663: We should no create side-effects inside broker. Please, use native `ping` method (if it exists) or check underlying library flag

faststream/rabbit/broker/broker.py
```diff
@@ -655,3 +655,13 @@ async def declare_exchange(
         """Declares exchange object in **RabbitMQ**."""
         assert self.declarer, NOT_CONNECTED_YET  # nosec B101
         return await self.declarer.declare_exchange(exchange)
+    
+    async def ping(self, timeout: int | None) -> bool:
+        await super().ping(timeout)
+        try:
```
Lancetnik, line 661: Also, I didn't see `timeout` logic here. You can set use it in native methods or `anyio.move_on_after` method

tests/brokers/base/connection.py
```diff
@@ -46,3 +46,21 @@ async def test_connect_by_url_priority(self, settings):
         await broker.connect(**kwargs)
         assert await self.ping(broker)
         await broker.close()
+
+    @pytest.mark.asyncio()
+    async def test_connection_alive(self, settings):
```
Lancetnik, line 51: I don't think, that we should create an extra tests. We already have `self.ping` method in the suite. Just use `broker.ping` inside.
But, we should add a test to check ping timeout

tests/brokers/rabbit/test_connect.py
```diff
@@ -58,3 +58,19 @@ async def test_connect_merge_args_and_kwargs_native(self, settings):
         broker = self.broker("fake-url")  # will be ignored
         assert await broker.connect(url=settings.url)
         await broker.close()
+
+    @pytest.mark.asyncio()
+    async def test_connection_alive(self, settings):
```
Lancetnik, line 63: Why are you duplicate these tests in the inheritor? They will be runned already due parent declaration

faststream/rabbit/broker/broker.py
Flosckow: In RMQ not ping method, add check flags

Flosckow:
In redis ping method was already added in the past, can i remove it on tests/broker/test_connect.py and add new method in broker and called it from parent class(brokers/base/connection.py)? As a result, the code will be unified

Lancetnik:
> In redis ping method was already added in the past, can i remove it on tests/broker/test_connect.py and add new method in broker and called it from parent class(brokers/base/connection.py)? As a result, the code will be unified

For sure

Flosckow:
@Lancetnik, can you view it? I added ping for all brokers

faststream/confluent/broker/broker.py
```diff
@@ -523,3 +524,12 @@ async def publish_batch(
             reply_to=reply_to,
             correlation_id=correlation_id,
         )
+
+    @override
+    async def ping(self, timeout: Optional[float]) -> bool:
```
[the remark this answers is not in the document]
kumaranvpl: @Lancetnik Sorry for the late reply.
Confluent does not provide any out of the box solution to check the connection. 
I wrote a ping method in my async confluent wrapper and calling it in the `broker.ping` method.
Tests passed locally, pushed it and waiting for tests to pass in CI.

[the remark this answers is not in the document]
kumaranvpl: @Lancetnik All tests are passing including confluent tests

Lancetnik:
@kumaranvpl ty. Great point, but I think, we can reuse `self._procuder._producer` (it has the same method) and ask for concrete topic from already using (if it has sence)

faststream/confluent/broker/broker.py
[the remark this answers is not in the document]
Lancetnik: @kumaranvpl can you please check my implementation?

faststream/confluent/client.py
```diff
@@ -227,21 +227,27 @@ async def send_batch(
...
-            admin_client = AdminClient(
-                {
-                    x: self.config[x]
-                    for x in ADMINCLIENT_CONFIG_PARAMS
-                    if x in self.config
-                }
-            )
             cluster_metadata = await call_or_await(
-                admin_client.list_topics, timeout=timeout
+                self.producer.list_topics,
```
kumaranvpl, line 245: Changes looks good to me @Lancetnik

Lancetnik (approved):

PR #1607, opened 2024-07-20
Feat: extend response api

### LINKED ISSUE #1587: feature: extend Response API
We should specify concrete broker Response classes (like [KafkaResponse](https://github.com/airtai/faststream/blob/main/faststream/kafka/response.py#L4)) to make them supporting all broker features ([Publisher.publish](https://github.com/airtai/faststream/blob/main/faststream/kafka/publisher/usecase.py#L100) options)

Another words, **KafkaResponse** should has the following signature:

```python
class KafkaResponse(Response):
    def __init__(
        self,
        body: "SendableMessage",
        *,
        headers: Optional["AnyDict"] = None,
        correlation_id: Optional[str] = None,
        # Kafka specific
        timestamp_ms: Optional[int] = None,
        key: Optional[bytes] = None,
        ...
    ) -> None:
        ...
```

Another brokers Response classes should be extended the same way

Flosckow:
# Description

Fixes #1587 

## Type of change

## Checklist


faststream/kafka/response.py
```diff
@@ -1,5 +1,19 @@
...
 
 class KafkaResponse(Response):
-    pass
+    def __init__(
+        self,
+        body: "SendableMessage",
+        *,
+        headers: Optional["AnyDict"] = None,
+        correlation_id: Optional[str] = None,
+        partition: Optional[int] = None,
```
Lancetnik, line 16: Remove `partition`

faststream/nats/response.py
```diff
@@ -1,5 +1,20 @@
...
 
 class NatsResponse(Response):
-    pass
+    def __init__(
+        self,
+        body: "SendableMessage",
+        *,
+        headers: Optional["AnyDict"] = None,
+        correlation_id: Optional[str] = None,
+        timeout: Optional[float] = None,
```
Lancetnik, line 16: Remove `timeout` and all RPC options

faststream/rabbit/response.py
```diff
@@ -1,5 +1,22 @@
...
+
 
 class RabbitResponse(Response):
-    pass
+    def __init__(
+        self,
+        body: "AioPikaSendableMessage",
+        *,
+        correlation_id: Optional[str] = None,
+        queue: Optional["RabbitQueue"] = None,
```
Lancetnik, line 16: Remove `queue`, `routing_key` and RPC options. Add `headers`

faststream/redis/response.py
```diff
@@ -1,5 +1,19 @@
...
 
 class RedisResponse(Response):
-    pass
+    def __init__(
+        self,
+        body: Optional["SendableMessage"] = None,
+        *,
+        headers: Optional["AnyDict"] = None,
+        correlation_id: Optional[str] = None,
+        rpc: bool = False,
```
Lancetnik, line 16: Remove all RPC options

faststream/confluent/response.py
```diff
@@ -1,5 +1,19 @@
...
 
 class KafkaResponse(Response):
-    pass
+    def __init__(
+        self,
+        body: "SendableMessage",
+        *,
+        headers: Optional["AnyDict"] = None,
+        correlation_id: Optional[str] = None,
+        partition: Optional[int] = None,
```
Lancetnik, line 16: Remove `partition` here

Lancetnik (changes requested):

faststream/rabbit/response.py
Lancetnik: Also, you forgot about RMQ extra options: there are a lot of them - https://github.com/Flosckow/faststream/blob/feat-extend-response-api/faststream/rabbit/publisher/usecase.py#L220

Lancetnik:
@Flosckow don't forget to initiate classes. Now you have only `__init__(): ...`, but it should use super and setup custom fields as well

Lancetnik:
And finally, don't forget to override `Response.as_publish_kwargs` for each class

faststream/rabbit/response.py
Flosckow: Yes, indeed, I missed that. We are talking about Publish Kwargs, which also need to be passed to method init?

Lancetnik: Yep

Flosckow:
If we have equal argument in specific woker response class, can we not override as_publish_kwargs, or is it better to do it right away, in case something comes up later?

Flosckow:
Can we use AioPikaSendableMessage(need add to base response class inside union), or more correctly strategy use SendableMessage inside rmq response?

Lancetnik (dismissed):

Lancetnik (approved):

Lancetnik (approved):

Lancetnik (approved):

PR #1683, opened 2024-08-13
Replace while-sleep with Event

Olegt0rr:
# Description

Instead of sleeping in a while loop, and waiting for a condition to become true, it's preferable to await on an Event object such as: `anyio.Event`

This way you don't bother event loop every `0.1` seconds (used as default) and handle condition change without waiting additional sleep time. 

## Type of change

## Checklist


faststream/app.py
```diff
@@ -236,3 +238,17 @@ async def _shutdown(self, log_level: int = logging.INFO) -> None:
...
+        return self._should_exit.is_set()
+
+    @should_exit.setter
+    def should_exit(self, value: Literal[True]) -> None:
+        """Set should exit flag state."""
+        if value is not True:
+            msg = "'should_exit' can receive only 'True' value"
+            raise ValueError(msg)
+
+        self._should_exit.set()
```
Olegt0rr, line 254: I'm not sure if you want to save `should_exit` public boolean or not

faststream/app.py
```diff
@@ -154,7 +153,13 @@ async def run(
         self,
         log_level: int = logging.INFO,
         run_extra_options: Optional[Dict[str, "SettingField"]] = None,
-        sleep_time: float = 0.1,
+        sleep_time: Annotated[
+            float,
+            deprecated(
+                "Deprecated in **FastStream 0.5.0**. "
```
Olegt0rr, line 159: Should I set patch-version here? What version should I set?

Lancetnik:
Sorry, we should take a little pause with this PR. Our first implementation worked this way, but we switched to sleeping due `nats-py` bug. We should check it again

faststream/app.py
Lancetnik: No, just remove it

Lancetnik:
@Olegt0rr I think, we can merge it and come back to the problem if there is a feedback from NATS users
Can u please resolve conflict and remove property to let me merge PR?

faststream/app.py
Lancetnik: 0.5.24

faststream/app.py
Lancetnik: Just remove it :)

Lancetnik (approved):

Issue #1806, opened 2024-09-23
Bug: Injecting a custom Context field implementing __eq__ does not work


antoinehumbert:
**Describe the bug**
Since https://github.com/airtai/faststream/pull/1667/files, when trying to use a custom Context field that implements \_\_eq\_\_ returning `NotImplemented` in some cases, the dependency is not injected in subscriber.

**How to reproduce**
Having following User class:

```python
class User:
    def __init__(self, id):
        self.id = id

    def __eq__(self, other):
        if not isinstance(other, User):
            return NotImplemented
        return self.id == other.id
```

Having a middleware injecting a User object in the context:

```python
from faststream import BaseMiddleware, context


class UserMiddleware(BaseMiddleware):
    async def on_consume(self, msg):
        context.set_local("user", User(1))
        return await super().on_consume(msg)
```

Having a subscriber with `user` from context as a dependency:

```python
@broker.subscriber("authenticated-topic")
def handler(msg, user: User = Context("user")):
    ...
````

The `user` parameter is not injected leading to a pydantic validation error because field is missing.

The issue arises because this (https://github.com/airtai/faststream/blob/689f6740a67c19b70c1e488afa081fe7a5b02eb2/faststream/utils/context/types.py#L63) `if` statement always evaluates to `False`.

Here is what happens when doing the comparison with `EMPTY`:

* `User.__eq__` is called, returning `NotImplementError`
* So, the comparison `User(1) != EMPTY` evaluates to `not(NotImplementedError)` which is `False`, leading to not injecting `user` in kwargs.

I think that `_EmptyPlaceHolder.__eq__` should return `False` instead of `NotImplementedError` when compared with an object which is not an `_EmptyPlaceHolder` object, because this kind of comparison is what would most commonly happen at https://github.com/airtai/faststream/blob/689f6740a67c19b70c1e488afa081fe7a5b02eb2/faststream/utils/context/types.py#L63, and the fact is that any object that is not an `_EmptyPlaceHolder` instance should pass the test.

Additionaly, I think this comparison should be made with `EMPTY` as the lhs operand (i.e. `EMPTY != ...`), as this will ensure that `_EmptyPlaceHolder.__eq__` is called first and will always return a result (`True` if the rhs is an `_EmptyPlaceHolder` object, `False` otherwise), avoiding potential side effects depending on implementation of `__eq__` in the compared object

Lancetnik:
[...]

PR #1809, opened 2024-09-23
fix: correct dependency injection of custom context fields implementing partial __eq__/__ne__

antoinehumbert:
This PR fixes issue #1806

It includes 2 regression tests related to both corrections in `faststream.utils.context.types` for the 2 use cases where value is compared to EMPTY.

Note that the issue description is incorrect as the custom field class must implement `__ne__` and not only `__eq__` to trigger the problem. This is the reason for the `User` implementation provided in regression test.

Please, let me know if PR needs some improvements.


Lancetnik (dismissed):

Lancetnik:
LGTM, thank you!

Lancetnik:
> Sorry for not executing the pre-commit check before submitting PR. Do you want me to correct those errors ?

It will be really kind

examples/e12_custom_context.py
```diff
@@ -0,0 +1,69 @@
+from typing import TYPE_CHECKING, Any
```
Lancetnik, line 1: Can you move your example to just a `tests/utils/context/test_main.py` 2 tests? It is too complex to be an example

Lancetnik (changes requested):

Lancetnik (approved):

Lancetnik:
@antoinehumbert can you run precommit please? 😄

antoinehumbert:
Will do tomorrow. Sorry for the incontinence.

antoinehumbert:
Hi! Everything looks good now.

Lancetnik (approved):

PR #1811, opened 2024-09-24
do not assume discriminator is not a property

lecko-cngroup:
# Description

We happen to have property named `discriminator` in few of our models. This PR prevents `KeyError: 'propertyName'` since the schema generation code incorrectly assumes `discriminator` is keyword used to specify `propertyName`.

## Type of change


Lancetnik (changes requested):
@@lecko-cn group thank you for the PR? Can you please add simple test with your case to `tests/async api` to be sure that this behavior fixed for the future refactoring? Anybody can remove id by accident without tests restriction

Lancetnik:
Btw, I am planning to make a release tomorrow evening. It will be great to see your fix in it.

lecko-cngroup:
@Lancetnik I guess these 2 new tests should be enough, right?

Lancetnik (approved):

PR #1860, opened 2024-10-18
docs: fix public api directory

Lancetnik:


Lancetnik:
@kumaranvpl can u please fix public api? Seems like the link broken, but I can't undestand, how it should looks like

PR #2038, opened 2025-01-13
Feat: add dto in subscribers and publishers constructor

Flosckow:
# Description

Fixes # (issue number)

## Type of change

## Checklist


faststream/_internal/publisher/specified.py
```diff
@@ -9,10 +9,11 @@
     P_HandlerParams,
     T_HandlerReturn,
 )
+from faststream.rabbit.schemas.publishers import SpecificationOptions
```
Lancetnik, line 12: We shouldn't import RMQ class here. It should be create here in opposite

faststream/_internal/publisher/specified.py
```diff
@@ -28,14 +29,19 @@ class SpecificationPublisher(EndpointSpecification[MsgType, PublisherSpec]):
...
-
-        super().__init__(*args, **kwargs)
+        self.schema_ = init_options.schema_
+        params = AsyncAPIOptions(
+            include_in_schema=init_options.include_in_schema,
+            title_=init_options.title_,
+            description_=init_options.description_,
+        )
+        # Call next base class parent init
+        super().__init__(*args, init_options=params, **kwargs)
```
Lancetnik, line 44: There is no reason to pass `params` to parent. Originally, we were passing just an `*args, **kwargs`

faststream/_internal/publisher/specified.py
```diff
@@ -28,14 +29,19 @@ class SpecificationPublisher(EndpointSpecification[MsgType, PublisherSpec]):
     def __init__(
         self,
         *args: Any,
-        schema_: Optional[Any],
+        init_options: SpecificationOptions,
```
Lancetnik, line 32: Probably, `specification_options` is a better name?

faststream/_internal/subscriber/specified.py
```diff
@@ -23,12 +23,12 @@ class SpecificationSubscriber(EndpointSpecification[MsgType, SubscriberSpec]):
     def __init__(
         self,
         *args: Any,
+        async_api_options: AsyncAPIOptions,
```
Lancetnik, line 26: It should be the same name with publisher class

faststream/_internal/subscriber/specified.py
```diff
@@ -9,7 +9,7 @@
 from faststream.specification.asyncapi.message import parse_handler_params
 from faststream.specification.asyncapi.utils import to_camelcase
 from faststream.specification.proto import EndpointSpecification
-from faststream.specification.schema import SubscriberSpec
+from faststream.specification.schema import AsyncAPIOptions, SubscriberSpec
```
Lancetnik, line 12: We don't use **AsyncAPI** naming in 0.6.0 anymore. Please, use `Specification` instead

faststream/_internal/subscriber/specified.py
```diff
@@ -23,12 +23,12 @@ class SpecificationSubscriber(EndpointSpecification[MsgType, SubscriberSpec]):
...
         self,
         *args: Any,
+        async_api_options: AsyncAPIOptions,
         **kwargs: Any,
     ) -> None:
         self.calls = []
-
         # Call next base class parent init
-        super().__init__(*args, **kwargs)
+        super().__init__(*args, init_options=async_api_options, **kwargs)
```
Lancetnik, line 31: We shouldn't pass `asyncapi` to parent as well

faststream/_internal/subscriber/usecase.py
```diff
@@ -88,23 +88,14 @@ class SubscriberUsecase(SubscriberProto[MsgType]):
...
-        self,
-        *,
-        no_reply: bool,
-        broker_dependencies: Iterable["Dependant"],
-        broker_middlewares: Sequence["BrokerMiddleware[MsgType]"],
-        default_parser: "AsyncCallable",
-        default_decoder: "AsyncCallable",
-        ack_policy: AckPolicy,
-    ) -> None:
+    def __init__(self, *, init_options: InternalOptions) -> None:
```
Lancetnik, line 91: We should choose an another name

faststream/rabbit/publisher/specified.py
```diff
@@ -37,39 +25,17 @@ class SpecificationPublisher(
...
         super().__init__(
-            title_=title_,
-            description_=description_,
-            include_in_schema=include_in_schema,
-            schema_=schema_,
+            init_options=specification_options,
             # propagate to RMQSpecificationMixin
-            queue=queue,
-            exchange=exchange,
+            queue=logic_options.queue,
```
Lancetnik, line 34: This class should consume DTO itself, isn't it?

Lancetnik (changes requested):

roma-frolov (approved):

faststream/confluent/publisher/usecase.py
```diff
@@ -25,26 +26,13 @@ class LogicPublisher(PublisherUsecase[MsgType]):
...
-            middlewares=middlewares,
-        )
+    def __init__(self, *, base_options: ConfluentPublisherBaseOptions) -> None:
+        super().__init__(publisher_options=base_options.internal_options)
 
-        self.topic = topic
-        self.partition = partition
-        self.reply_to = reply_to
-        self.headers = headers or {}
+        self.topic = base_options.topic
```
vrslev, line 32: I wouldn't unpack the options, just set `self.base_options = base_options`

vrslev: Here, and in other similar cases.

faststream/confluent/schemas/subscribers.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@dataclass
+class ConfluentSubscriberBaseOptions:
+    topics: Sequence[str]
+    partitions: Sequence["TopicPartition"]
+    polling_interval: float
+    group_id: Optional[str]
+    connection_data: "AnyDict"
+    internal_options: SubscriberUsecaseOptions
```
vrslev, line 22: I don't feel nesting fits in here

faststream/kafka/subscriber/factory.py
```diff
@@ -169,57 +174,48 @@ def create_subscriber(
         connection_args["enable_auto_commit"] = True
         ack_policy = AckPolicy.DO_NOTHING
 
+    internal_options = SubscriberUsecaseOptions(
```
vrslev, line 177: I'd move building DTOs from factories to  `broker.subscriber()` and `broker.producer()`

faststream/confluent/publisher/factory.py
```diff
@@ -108,41 +105,39 @@ def create_publisher(
...
-
-    return SpecificationDefaultPublisher(
+    internal_options = PublisherUsecaseOptions(
+        broker_middlewares=cast(
+            "Sequence[BrokerMiddleware[tuple[ConfluentMsg, ...]]]",
+            broker_middlewares,
+        ),
+        middlewares=middlewares,
+    )
+    base_options = ConfluentPublisherBaseOptions(
```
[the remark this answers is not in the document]
vrslev: Agreed

faststream/confluent/schemas/subscribers.py
vrslev: Also, name `internal_options` is a bit confusing

faststream/kafka/subscriber/factory.py
Lancetnik: > I'd move building DTOs from factories to `broker.subscriber()` and `broker.producer()`

Can't agree here. `broker.subscriber` and `broker.publisher` is just a public facade and should not contains any logic - just calling factory inside it. Factory should be a real entrypoint to the FastStream internals, so we should pack arguments there as well

Lancetnik: But I am not sure about calls order.
We should
* pack arguments to DTO (should be first?)
* validate DTO for misconfiguration
* replace all sentinels to default values (should we pack arguments to DTO after this phase?)
* use DTO to build final objects

faststream/confluent/schemas/subscribers.py
Lancetnik: > I don't feel nesting fits in here

Do you think, that inheritance is better here?

faststream/confluent/publisher/factory.py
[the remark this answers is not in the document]
Lancetnik: I think, cmth like `*Config` should be OK

faststream/kafka/subscriber/factory.py
vrslev: I'd focus on making the solution as straightforward as possible and trying to avoid boilerplate complex logic that you have to copy-paste across broker implementations.

> validate DTO for misconfiguration

It makes sense to me to do this inside DTO ("won't build if misconfigured")

> replace all sentinels to default values

Majority of defaults are a part of public API, I don't see any value in adding them in DTOs. If DTO is validated on construction, then derived state can be put in `@property`s

faststream/confluent/publisher/usecase.py
Flosckow: I don't really like the idea of not passing in the constructor. This would lead to a change in all classes, but ultimately, it's not the base configurations that are used. Instead, it's the topic or some other var that seems superfluous.
It seems that we will add code for the sake of code.

faststream/kafka/subscriber/factory.py
Flosckow: Yes, this is a great suggestion to move validation inside the dto

Flosckow: Yes, this is a great suggestion to move validation inside the dto

faststream/confluent/publisher/usecase.py
Lancetnik: Also, it will be a breaking change and we are risking to break a lot of code depends of class public properties. I don't think, that we should make such changes in this PR.

faststream/confluent/schemas/subscribers.py
Lancetnik: > I think internal options can be independent from _broker specific_ subscription options, they don't seem to belong in this DTO

Probaly, you are right, but it conflicts with a decision we made for SpecifiationConfig. Now we have

```python
class SpecificationConfig: ...
class PublisherSpecificationConfig(SpecificationConfig): ...
class SubscriberSpecificationConfig(SpecificationConfig): ...

class [Broker]SubscriberSpecificationConfig(SubscriberSpecificationConfig): ...
class [Broker]PublisherSpecificationConfig(PublisherSpecificationConfig): ...

class EndpointConfig: ...
class PublisherConfig(EndpointConfig): ...
class SubscriberConfig(EndpointConfig): ...

class [Broker]SubscriberConfig(SubscriberConfig): ...
class [Broker]PublisherConfig(PublisherConfig): ...

def factory():
     return [Broker][Publisher/Subscriber](
          config=[Broker][Subscriber/Publisher]Config(),
          specification_config=[Broker][Subscriber/Publisher]SpecificationConfig(),
     )
```

But, you are suggesting to remove inheritance and it will be looks like

```python
def factory():
     return [Broker][Publisher/Subscriber](
          config=[Broker][Subscriber/Publisher]Config(),
          base_config=[Subscriber/Publisher]Config(),
          specification_config=[Broker][Subscriber/Publisher]SpecificationConfig(),
          base_specificaiont_config=[Subscriber/Publisher]SpecificationConfig(),
     )
```

I think, it is too confusing API and we should stay with inheritance here.

faststream/kafka/subscriber/factory.py
Lancetnik: LGTM

Flosckow: I reread the thread, I'm a little confused about property, how do you see it?

Flosckow:
@Lancetnik, pls check confluent, i rewrite it without internal options only use inheritance, this required the creation of a dto in the constructor for a lower level of internal  so that the details would not leak.

PR #2128, opened 2025-03-17
docs: Inlines the include files directly into the location where it is used.

### LINKED ISSUE #2109: docs: inline docs/includes to pages itself
We have a lot of `includes/` - https://github.com/airtai/faststream/tree/main/docs/includes files created to share them between different translations. But now, we don't plan to support multilanguage documentation anymore. So, we should inline such files to usages place to make documentation sources much easy to understand, support and work with.

This is a huge work, so you can take care of any number of files you want – all PRs are welcome!

PLEASE, DO NOT INLINE FILES ALREADY USING MULTIPLE TIMES

MaestroXXXVIII:
# Description
Inlines the include files directly into the location where it is used

#2109 

## Type of change

## Checklist

- [x] docs/docs/en/(kafka, nats, rabbit, redis, confluent, getting-started/contributing)
- [x] After modifications I checked the local version between production version


Lancetnik (changes requested):
Please, do not inline docker file

docs/docs/en/kafka/ack.md
```diff
@@ -60,4 +60,6 @@ If you wish to interrupt the processing of a message at any call stack level and
 
 This way, **FastStream** interrupts the current message processing and acknowledges it immediately. Similarly, you can raise `NackMessage` as well to prevent the message from being committed.
 
-{! includes/en/no_ack.md !}
+!!! tip
```
Lancetnik, line 63: Also, we shouldn't inline blocks, that uses multiple times. Eg – `no_ack.md` uses 7 times, we should duplicate such texts to be able edit them in the single source.

Please, check the blocks you inlined again

MaestroXXXVIII: в

Lancetnik: Indeed, sorry that didn't check it before

PR #2196, opened 2025-05-01
feat: add batch publish support for redis stream

### LINKED ISSUE #2195: Feature: Batch publish support for redis stream
### Summary

I'd like to propose adding **batch publish** functionality to the redis stream module. This feature would allow publishing multiple entries to the stream in a single publish operation.

### Motivation

Batch publishing is a **common and performance-critical feature** in stream-oriented systems. It helps improve throughput and reduce network latency by minimizing the number of round-trips.

- `List` operations in redis module already support batching
- redis streams are often used for high-throughput message ingestion, and batch support would significantly improve performance in such scenarios

### Proposed Solution

I propose to add a method like:
```python
pub_broker = RedisBroker("redis://127.0.0.1:6379")

# Stream without batch
queue = StreamSub("single")

# Stream with batch enabled
batch_stream = StreamSub("batch", batch=True)

@pub_broker.subscriber(stream=queue)
@pub_broker.publisher(stream=batch_stream)
async def m(msg):
    return 1, 2, 3

@pub_broker.subscriber(stream=batch_stream)
async def resp(msg):
    assert msg == [1, 2, 3]  # Should pass!
    
await pub_broker.publish("", stream=queue)
```

caozheliang:
## Description

The `batch=True` option in the stream configuration now also applies to **publisher**, enabling it to publish multiple messages to the stream in a single `publish()` call.

This improves performance and efficiency, especially in high-throughput scenarios.

closes #2195 

## Type of change

## Checklist

### DIFF
--- a/docs/docs/SUMMARY.md
+++ b/docs/docs/SUMMARY.md
@@ -1069,6 +1069,7 @@ search:
                         - [AsyncAPIListBatchPublisher](api/faststream/redis/publisher/asyncapi/AsyncAPIListBatchPublisher.md)
                         - [AsyncAPIListPublisher](api/faststream/redis/publisher/asyncapi/AsyncAPIListPublisher.md)
                         - [AsyncAPIPublisher](api/faststream/redis/publisher/asyncapi/AsyncAPIPublisher.md)
+                        - [AsyncAPIStreamBatchPublisher](api/faststream/redis/publisher/asyncapi/AsyncAPIStreamBatchPublisher.md)
                         - [AsyncAPIStreamPublisher](api/faststream/redis/publisher/asyncapi/AsyncAPIStreamPublisher.md)
                     - producer
                         - [RedisFastProducer](api/faststream/redis/publisher/producer/RedisFastProducer.md)
@@ -1077,6 +1078,7 @@ search:
                         - [ListBatchPublisher](api/faststream/redis/publisher/usecase/ListBatchPublisher.md)
                         - [ListPublisher](api/faststream/redis/publisher/usecase/ListPublisher.md)
                         - [LogicPublisher](api/faststream/redis/publisher/usecase/LogicPublisher.md)
+                        - [StreamBatchPublisher](api/faststream/redis/publisher/usecase/StreamBatchPublisher.md)
                         - [StreamPublisher](api/faststream/redis/publisher/usecase/StreamPublisher.md)
                 - response
                     - [RedisResponse](api/faststream/redis/response/RedisResponse.md)
--- a/docs/docs/en/api/faststream/redis/publisher/asyncapi/AsyncAPIStreamBatchPublisher.md
+++ b/docs/docs/en/api/faststream/redis/publisher/asyncapi/AsyncAPIStreamBatchPublisher.md
@@ -1,0 +1,11 @@
+---
+# 0.5 - API
+# 2 - Release
+# 3 - Contributing
+# 5 - Template Page
+# 10 - Default
+search:
+  boost: 0.5
+---
+
+::: faststream.redis.publisher.asyncapi.AsyncAPIStreamBatchPublisher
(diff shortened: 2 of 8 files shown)


Lancetnik:
Sorry, I am doubt a bit. Seems like it conflicts with our ideology "do not provide abstractions, that original broker does not support". But may be it doesn't... I have to think

caozheliang:
> Sorry, I am doubt a bit. Seems like it conflicts with our ideology "do not provide abstractions, that original broker does not support". But may be it doesn't... I have to think

Got it — I see what you mean.

In my day-to-day usage of Redis Streams, I frequently rely on the pipeline feature to handle batch operations efficiently.
From that perspective, this functionality could be considered Redis's built-in way of implementing a publish-batch mechanism for stream data.

Lancetnik:
> Got it — I see what you mean.
> 
> In my day-to-day usage of Redis Streams, I frequently rely on the pipeline feature to handle batch operations efficiently. From that perspective, this functionality could be considered Redis's built-in way of implementing a publish-batch mechanism for stream data.

Why you prefer built-in FastStream implementation than make the same manually in handler body?

caozheliang:
> Why you prefer built-in FastStream implementation than make the same manually in handler body?

Absolutely. I could certainly implement similar logic manually inside the handler — for example, using a Redis Lua script.
But batch publishing is a much more common use case (and one of the main reasons I regularly use pipelines in Redis).

Encapsulating this functionality would make it much easier to integrate with FastStream's existing features — such as using the publisher as a decorator, or automatically filling in metadata like correlation_id.

Also, unlike Redis Lua scripts, batch publishing is a built-in capability in Kafka. By exposing the same abstraction here, we can make it easier for users to switch from Redis to Kafka with minimal friction.

What do you think?

Lancetnik:
Sorry, I don't think, that we should merge it.
In my opinion, we can impl `publish_batch` if broker protocol allows us to publish multiple messages atomically – by one network operation or transactionally.
Redis transactions does not compress multiple request to one network operation. Also, they allows to read messages before transaction commit – "dirty read". Moreover, I am not sure about operation transactional reverting...
So, I think, cus this feature have no guarantie about smth – we shouldn't add it to the framework

Alternatively, we can add smth close to it code example to **Redis - How To** documentation section

Please, correct me if I was wrong somewhere

Lancetnik:
Well, I asked the community about this PR merhing and collect 18 (to merge) VS 27 (against merge)

Finally, I don't think, that we should do this

As an alternative I can suggest `Pipeline` feature

```python
publisher = broker.publisher("out-stream")

@broker.subscriber("in-stream")
async def handler(msg_body, pipe: Pipeline) -> None:
    async for i in range(10):
          await publisher.publish(i, pipeline=pipe)
    await pipe.execute()
```

Lancetnik:
@caozheliang have you any thoughts?

caozheliang:
> @caozheliang have you any thoughts?

Thanks for the detailed explanation, and I agree with your perspective.

Indeed, from a semantic point of view, Redis's pipeline is not truly an atomic batch publish operation. For example, it lacks transactional rollback and may lead to issues like dirty reads. I understand and respect the rationale behind not merging this PR.

I appreciate your suggestion to explore the `Pipeline` feature as an alternative. It's a practical direction, and I'll adjust the implementation accordingly.

Let me know if you have any further thoughts. Thanks again for the constructive discussion.

Lancetnik:
> I appreciate your suggestion to explore the `Pipeline` feature as an alternative. It's a practical direction, and I'll adjust the implementation accordingly.

If you would like to contribute the `Pipeline` feature, it should be easy to do – we just need to create annotated Depends smth like that:

```python
async def get_pipeline(redis: faststream.redis.annotations.Redis) -> AsyncIterator[redis.Pipeline]:
     async with redis.pipeline() as pipe:
         yielf pipe

Pipeline = Annotated[redis.Pipeline, Depends(get_pipeline)]
```

And then we should just add `pipeline` option to all publish interfaces.

Finally, I think, we can close the current Issue. Anyway, thank you for you job, ideas and interest in the project!

PR #2208, opened 2025-05-07
feat: added option to read from pending messages list in redis streams with group mode

### LINKED ISSUE #2197: Feature: Read of Pending Messages Support for Redis Stream
# Summary
I`d like to have option to work with PEL in a Redis Stream in group mode.

# Motivation
The Redis `XREADGROUP` command provides the ability to read the history of received but not confirmed Redis messages. Adding this feature will allow you to create routers that work with such messages.

# Proposed Solution

```python

@broker.subscriber(stream=StreamSub(stream, group=group, consumer=consumer, last_msg_id='>'), retry=True) # last_msg_id is default '>'
async def handler_for_new_messages(msg):
    # do something
    raise Exception
    # retry option will msg not acked

@broker.subscriber(stream=StreamSub(stream, group=group, consumer=consumer, last_msg_id='0'))
async def handler_for_no_acked_messages(msg): ...
   # do something
```
# Additional context
This scenario in redis docs  - https://redis.io/docs/latest/commands/xreadgroup/ (What happens when a message is delivered to a consumer?)

Currently, this functionality is blocked by the hardcoded value '>' in the `xreadgroup` call. 
```python
return client.xreadgroup(
    groupname=stream.group,
    consumername=stream.consumer,
    streams={stream.name: ">"},
    count=stream.max_records,
    noack=stream.no_ack,
)
```
In addition, adding this parameter to the Stream Sub hash calculation will allow this logic to be implemented within a single consumer.

```python
@broker.subscriber(stream=StreamSub(stream, group=group, consumer=consumer, msg_id='>'), retry=True)
@broker.subscriber(stream=StreamSub(stream, group=group, consumer=consumer, msg_id='0'))
async def hybrid_handler(msg): ...
```

P.S.
I have already implemented this option, and I have a working prototype. If you are interested in this improvement, I could create a PR on this topic.

Barchonok:
# Description

Added functional to provide option to change first reading message from Redis Streams in group mode.
Added  

Fixes #2197

## Type of change

## Checklist

- [x] My code adheres to the style guidelines of this project (scripts/lint.sh shows no errors)
- [ ] Both new and existing unit tests pass successfully on my local environment by running scripts/test-cov.sh (tested all, without kafka, rabbit, nats)
- [x] I have ensured that static analysis tests are passing by running scripts/static-analysis.sh

### DIFF
--- a/faststream/redis/schemas/stream_sub.py
+++ b/faststream/redis/schemas/stream_sub.py
@@ -12,6 +12,7 @@ class StreamSub(NameRequired):
         "batch",
         "consumer",
         "group",
+        "last_history_id",
         "last_id",
         "max_records",
         "maxlen",
@@ -29,6 +30,7 @@ class StreamSub(NameRequired):
         batch: bool = False,
         no_ack: bool = False,
         last_id: Optional[str] = None,
+        last_history_id: Optional[str] = None,
         maxlen: Optional[int] = None,
         max_records: Optional[int] = None,
     ) -> None:
@@ -45,6 +47,9 @@ class StreamSub(NameRequired):
         if last_id is None:
             last_id = "$"
 
+        if last_history_id is None:
+            last_history_id = ">"
+
         super().__init__(stream)
 
         self.group = group
@@ -53,6 +58,7 @@ class StreamSub(NameRequired):
         self.batch = batch
         self.no_ack = no_ack
         self.last_id = last_id
+        self.last_history_id = last_history_id
         self.maxlen = maxlen
         self.max_records = max_records
 
--- a/faststream/redis/subscriber/usecase.py
+++ b/faststream/redis/subscriber/usecase.py
@@ -666,7 +666,7 @@ class _StreamHandlerMixin(LogicSubscriber):
                 return client.xreadgroup(
                     groupname=stream.group,
                     consumername=stream.consumer,
-                    streams={stream.name: ">"},
+                    streams={stream.name: stream.last_history_id},
                     count=stream.max_records,
                     block=stream.polling_interval,
                     noack=stream.no_ack,
--- a/docs/docs/en/release.md
+++ b/docs/docs/en/release.md
@@ -12,6 +12,25 @@ hide:
 ---
 
 # Release Notes
+## 0.5.40
+
+### What's Changed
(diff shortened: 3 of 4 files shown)


faststream/redis/subscriber/usecase.py
```diff
@@ -666,7 +666,7 @@ def read(
                 return client.xreadgroup(
                     groupname=stream.group,
                     consumername=stream.consumer,
-                    streams={stream.name: ">"},
+                    streams={stream.name: stream.last_history_id},
```
Lancetnik, line 669: Seems like we will read the same message each time this way

faststream/redis/schemas/stream_sub.py
```diff
@@ -29,6 +30,7 @@ def __init__(
         batch: bool = False,
         no_ack: bool = False,
         last_id: Optional[str] = None,
+        last_history_id: Optional[str] = None,
```
Lancetnik, line 33: Why we add a new option? Could we use already exis `last_id`?

Lancetnik:
Btw, you can set current it to group manually by https://redis.io/docs/latest/commands/xgroup-setid/
I think, it's a better way

faststream/redis/schemas/stream_sub.py
Barchonok: `last_id` defines that message be "not read" for new consumer, it is not about option to read from history for group.

Barchonok:
I tested more, you are right, it is better way, because in my scenario we have 2 consumers in one group - first for "not read" messages and second for read from history. But Redis added message to history queue at one time with get this message from "not read" queue.
I learned something new. Thank you for it.

For my scenario will better list sub with `LMOVE' (https://redis.io/docs/latest/commands/lmove/) command. That about this option in faststream? I found only lpop command

Thank you for time

Lancetnik:
@Barchonok lmove has no sence in your service code due it doesn't allow to read a message. It just moves message inside Redis, you can't read, process and then move a message by this command. In this case you should read, process and republish it manually instead

We already have an Issue to document such scenarios #2210 , just wait a bit for it

Btw, what we should do with current PR? Can we close it?

Barchonok:
Hello, yes, we can close this PR.
Thank you for answer!

PR #2212, opened 2025-05-15
feat: add cluster metadata request via aiokafka admin client in ping

murzinov01:
# Description

Add request of the kafka cluster metadata via aiokafka.admin.client.AIOKafkaAdminClient in `broker.ping` method to achieve ping correct work.
For now, if you run FastStream application, connected to the Kafka, for example, in docker container and use `broker.ping` in /readiness probe you can encounter a bug:

After shutting down a Kafka docker container, broker.ping still respond as True value.
I believe that this is wrong behaviour. 
More proper and expected decision is to return False or at least raise an Exception in that case.

Fixes # (issue number)

## Type of change

## Checklist

### DIFF
--- a/faststream/kafka/broker/broker.py
+++ b/faststream/kafka/broker/broker.py
@@ -1,4 +1,5 @@
 import logging
+import warnings
 from functools import partial
 from typing import (
     TYPE_CHECKING,
@@ -605,11 +606,22 @@ class KafkaBroker(
         Consumes the same with `KafkaBroker.__init__` arguments and overrides them.
         To startup subscribers too you should use `broker.start()` after/instead this method.
         """
+        if bootstrap_servers is not EMPTY or kwargs:
+            warnings.warn(
+                "`KafkaBroker().connect(...) options were "
+                "deprecated in **FastStream 0.5.40**. "
+                "Please, use `KafkaBroker(...)` instead. "
+                "All these options will be removed in **FastStream 0.6.0**.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+
         if bootstrap_servers is not EMPTY:
             connect_kwargs: AnyDict = {
                 **kwargs,
                 "bootstrap_servers": bootstrap_servers,
             }
+
         else:
             connect_kwargs = {**kwargs}
 
--- a/faststream/kafka/broker/broker.py
+++ b/faststream/kafka/broker/broker.py
@@ -1,6 +1,6 @@
-from contextlib import suppress
 import logging
 import warnings
+from contextlib import suppress
 from functools import partial
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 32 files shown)


faststream/kafka/broker/broker.py
```diff
@@ -637,11 +643,18 @@ async def _connect(  # type: ignore[override]
         security_params = parse_security(self.security)
         kwargs.update(security_params)
 
+        self._admin_client = AIOKafkaAdminClient(
+            loop=kwargs["loop"],
+            bootstrap_servers=kwargs["bootstrap_servers"],
+            client_id=client_id,
+            **security_params,
+        )
```
murzinov01, line 651: Is it okay to pass parameters to Admin client this way? 
Is it necessary to pass other parameters, like `connection_max_idle_ms`, `request_timeout_ms`, etc ?

Lancetnik: We should pass all allowed for AdminClient options we have

faststream/kafka/broker/broker.py
```diff
@@ -908,6 +921,14 @@ async def publish_batch(
     async def ping(self, timeout: Optional[float]) -> bool:
         sleep_time = (timeout or 10) / 10
 
+        if self._admin_client is None:
+            return False
+
+        try: 
+            await self._admin_client.describe_cluster()
+        except Exception:
+            return False
```
murzinov01: Is it okey to except Exception? Or may be except some small range exception, like a `NodeNotReadyError`, which i found in functions call stack inside of the `self._admin_client.describe_cluster()`

Lancetnik: I think, it's okay this case
Also, I image smth like that:

```python
with anyio.move_on_after(timeout) as cancel_scope:
    while True:
        with suppress(Exception):
            await self._admin_client.describe_cluster()
            return True
        await anyio.sleep(sleep_time)

return False
```

murzinov01: Okey, I fixed that

faststream/kafka/broker/broker.py
```diff
@@ -908,6 +921,14 @@ async def publish_batch(
     async def ping(self, timeout: Optional[float]) -> bool:
         sleep_time = (timeout or 10) / 10
 
+        if self._admin_client is None:
+            return False
+
+        try:
```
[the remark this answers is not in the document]
murzinov01: Ok. I think that in other broker classes(like RabbitMQ and others) `the broker.ping` should be checked for an example of a similar bug. But this is beyond the scope of my PR. Just keep it mind

faststream/kafka/broker/broker.py
murzinov01: Nikita, i decided to do the similar class to `kafka.schemas.params.ConsumerConnectionParams`. I implemented AdminClientConnectionParams, which actually contains **the same list of params** of `ConsumerConnectionParams` (i verified it). But i think it's a good idea to separate them for future independent support.

P.S In 0.6.0 branch I did not find a KafkaConfig class with all parameters and properties with exact params for each consumer, producer, etc classes (we discussed this in telegram). That's why I decided to implement `AdminClientConnectionParams` class

murzinov01:
Tests in `tests/brokers/kafka` were successful like in **main** branch 😌

Lancetnik (approved):

Lancetnik (approved):

Lancetnik (approved):

PR #2226, opened 2025-05-20
Fix default rabbit timestamp

### LINKED ISSUE #2225: Bug: Timestamp is not set automatically when publishing a new RabbitMQ message
**Describe the bug**
The `timestamp` property of the RabbitMQ message is not set automatically, even though [the documentation](https://faststream.airt.ai/latest/rabbit/publishing/#message-parameters:~:text=message%20sending%20time%20(set%20automatically)) states that it should be.

**How to reproduce**
1. Start a local RabbitMQ instance:

```bash
docker run --rm --name some-rabbit -p 15672:15672 -p 5672:5672 rabbitmq:3-management
```
2. Run the following code snippet using `faststream-cli`:

```python
import datetime as dt
from typing import Any

from faststream.rabbit import RabbitBroker
from faststream.rabbit.annotations import RabbitMessage

broker = RabbitBroker()
app = FastStream(broker)


@broker.subscriber("first-handler")
async def first_handler(msg: RabbitMessage) -> dict[str, Any]:
    assert isinstance(msg.raw_message.timestamp, dt.datetime)
    assert dt.datetime.now(tz=dt.UTC) - msg.raw_message.timestamp < dt.timedelta(seconds=1)
    return {"hello": "world"}


@broker.subscriber("second-handler")
async def second_handler(msg: RabbitMessage) -> None:
    assert isinstance(msg.raw_message.timestamp, dt.datetime)
    assert dt.datetime.now(tz=dt.UTC) - msg.raw_message.timestamp < dt.timedelta(seconds=1)


@app.after_startup
async def test_publish():
    await broker.publish(
        "message",
        "first-handler",
        reply_to="second-handler",
    )
```

**Expected behavior**
The `timestamp` property should be set automatically when a message is published, and it should reflect the current datetime.

**Observed behavior**
The `timestamp` property is `None`.

**Environment**
Running FastStream 0.5.41 with CPython 3.12.8 on Darwin

dmder:
# Description

The docstring for `RabbitBroker` suggests that `timestamp: Message publish timestamp. Generated automatically if not present.`, however in fact it is not since `aio_pika.message.encode_timestamp` doesn't do anything with `None`s.

Fixing this annoyance here.

Fixes #2225 

## Type of change

## Checklist

### DIFF
--- a/tests/brokers/rabbit/test_publish.py
+++ b/tests/brokers/rabbit/test_publish.py
@@ -1,9 +1,9 @@
 import asyncio
 import datetime as dt
-from contextlib import AsyncExitStack
 from unittest.mock import Mock, patch
 
 import pytest
+from dirty_equals import IsNow
 
 from faststream import Context
 from faststream.rabbit import RabbitBroker, RabbitResponse, ReplyConfig
@@ -136,7 +136,6 @@ class TestPublish(BrokerPublishTestcase):
         event: asyncio.Event,
         mock: Mock,
     ):
-        real_now = dt.datetime.now(tz=dt.timezone.utc)
         pub_broker = self.get_broker(apply_types=True)
 
         @pub_broker.subscriber(queue)
@@ -144,12 +143,7 @@ class TestPublish(BrokerPublishTestcase):
             mock(body=msg.body, timestamp=msg.raw_message.timestamp)
             event.set()
 
-        async with AsyncExitStack() as stack:
-            br = await stack.enter_async_context(self.patch_broker(pub_broker))
-            mock_datetime = stack.enter_context(
-                patch("faststream.rabbit.broker.broker.datetime")
-            )
-            mock_datetime.datetime.now.return_value = real_now
+        async with self.patch_broker(pub_broker) as br:
             await br.start()
 
             await asyncio.wait(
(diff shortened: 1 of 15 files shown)


Lancetnik (changes requested):
I don't think, that we should impl this logic everywhere
Please, move it to `producer._publish` method – https://github.com/ag2ai/faststream/blob/main/faststream/rabbit/publisher/producer.py#L224

tests/brokers/rabbit/test_publish.py
```diff
@@ -126,3 +128,42 @@ async def handle():
...
+                ),
+                timeout=3,
+            )
+
+            assert event.is_set()
+
+        # microseconds are lost as timestamps are passed around
+        # as epochs with second resolution in Rabbit
+        mock.assert_called_once_with(
+            body=b"", timestamp=real_now.replace(microsecond=0)
```
Lancetnik, line 168: We can use dirty-equals here to avoid datetime call patching and minify test logic – https://dirty-equals.helpmanual.io/latest/types/datetime/#dirty_equals.IsNow

dmder: done

dmder:
> I don't think, that we should impl this logic everywhere Please, move it to `producer._publish` method – https://github.com/ag2ai/faststream/blob/main/faststream/rabbit/publisher/producer.py#L224

done

Lancetnik (approved):

PR #2251, opened 2025-06-01
docs: Simplify `JSON Basic Serialization` section

sobolevn:
It used to have 5 identical tabs:
Снимок экрана 2025-06-01 в 17 20 26

Only channel names were different, but it is not important.


Lancetnik:
I think, the example should has a broker-agnostic channel name then

Lancetnik:
Sorry, but it's not working this way. We should stay with code example includes to make them testable. We have a tests for all included code examples.

I can change the thing by myself – it requires docs_src and tests changes I suppose

PR #2287, opened 2025-06-11
feat: (#2061) Redis Binary Format

### LINKED ISSUE #2061: Bug: Custom Serialization/Deserialization logic in Redis doesn't work unless UTF-8 serializable
**Describe the bug**
I am working on a middleware to add msgpack serialization to all messages (using ormsgpack which natively handles Pydantic models).

The issue seems to be that Faststream JSON-serializes messages along with header here:

```python
# faststream/redis/parser.py
class RawMessage:
# line: 85
    @classmethod
    def encode(
        cls,
        *,
        message: Union[Sequence["SendableMessage"], "SendableMessage"],
        reply_to: Optional[str],
        headers: Optional["AnyDict"],
        correlation_id: str,
    ) -> bytes:
        msg = cls.build(
            message=message,
            reply_to=reply_to,
            headers=headers,
        )

        return dump_json(
            {
                "data": msg.data,
                "headers": msg.headers,
            }
        )
```

Technically msg.data is supposed to be able to be bytes but in practice it has to be utf-8 compatible or it raises an exception:
```bash
/.../.venv/lib/python3.12/site-packa │
│ ges/faststream/redis/publisher/producer.py:79 in publish                              │
│                                                                                       │
│    76 │   │   │   psub = self._connection.pubsub()                                    │
│    77 │   │   │   await psub.subscribe(reply_to)                                      │
│    78 │   │                                                                           │
│ ❱  79 │   │   msg = RawMessage.encode(                                                │
│    80 │   │   │   message=message,                                                    │
│    81 │   │   │   reply_to=reply_to,                                                  │
│    82 │   │   │   headers=headers,                                                    │
│                                                                                       │
│ ╭───────────────────────────────────── locals ──────────────────────────────────────╮ │
│ │        channel = None                                                             │ │
│ │ correlation_id = '2ac5886a-736d-4712-b878-448b8b041f43'                           │ │
│ │        headers = None                                                             │ │
│ │           list = 'job-queue'                                                      │ │
│ │         maxlen = None                                                             │ │
│ │        message = b'\x82\xa3url\xb1http://kelly.com/\xa4type\xa3far'               │ │
│ │           psub = None                                                             │ │
│ │  raise_timeout = False                                                            │ │
│ │       reply_to = ''                                                               │ │
│ │            rpc = False                                                            │ │
│ │    rpc_timeout = 30.0                                                             │ │
│ │           self = <faststream.redis.publisher.producer.RedisFastProducer object at │ │
│ │                  0x1137972f0>                                                     │ │
│ │         stream = None                                                             │ │
│ ╰───────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                       │
│ /.../.venv/lib/python3.12/site-packa │
│ ges/faststream/redis/parser.py:101 in encode                                          │
│                                                                                       │
│    98 │   │   │   correlation_id=correlation_id,                                      │
│    99 │   │   )                                                                       │
│   100 │   │                                                                           │
│ ❱ 101 │   │   return dump_json(                                                       │
│   102 │   │   │   {                                                                   │
│   103 │   │   │   │   "data": msg.data,                                               │
│   104 │   │   │   │   "headers": msg.headers,                                         │
│                                                                                       │
│ ╭────────────────────────────────── locals ───────────────────────────────────╮       │
│ │ correlation_id = '2ac5886a-736d-4712-b878-448b8b041f43'                     │       │
│ │        headers = None                                                       │       │
│ │        message = b'\x82\xa3url\xb1http://kelly.com/\xa4type\xa3far'         │       │
│ │            msg = <faststream.redis.parser.RawMessage object at 0x10e69c790> │       │
│ │       reply_to = ''                                                         │       │
│ ╰─────────────────────────────────────────────────────────────────────────────╯       │
│                                                                                       │
│ /.../.venv/lib/python3.12/site-packa │
│ ges/faststream/_compat.py:93 in dump_json                                             │
│                                                                                       │
│    90 │   │   return to_jsonable_python(model, **kwargs)                              │
│    91 │                                                                               │
│    92 │   def dump_json(data: Any) -> bytes:                                          │
│ ❱  93 │   │   return json_dumps(model_to_jsonable(data))                              │
│    94 │                                                                               │
│    95 │   def get_model_fields(model: Type[BaseModel]) -> Dict[str, Any]:             │
│    96 │   │   return model.model_fields                                               │
│                                                                                       │
│ ╭───────────────────────────────────── locals ─────────────────────────────────────╮  │
│ │ data = {                                                                         │  │
│ │        │   'data': b'\x82\xa3url\xb1http://kelly.com/\xa4type\xa3far',           │  │
│ │        │   'headers': {                                                          │  │
│ │        │   │   'correlation_id': '2ac5886a-736d-4712-b878-448b8b041f43'          │  │
│ │        │   }                                                                     │  │
│ │        }                                                                         │  │
│ ╰──────────────────────────────────────────────────────────────────────────────────╯  │
│                                                                                       │
│ /.../.venv/lib/python3.12/site-packa │
│ ges/faststream/_compat.py:90 in model_to_jsonable                                     │
│                                                                                       │
│    87 │   │   model: BaseModel,                                                       │
│    88 │   │   **kwargs: Any,                                                          │
│    89 │   ) -> Any:                                                                   │
│ ❱  90 │   │   return to_jsonable_python(model, **kwargs)                              │
│    91 │                                                                               │
│    92 │   def dump_json(data: Any) -> bytes:                                          │
│    93 │   │   return json_dumps(model_to_jsonable(data))                              │
│                                                                                       │
│ ╭───────────────────────────────────── locals ──────────────────────────────────────╮ │
│ │ kwargs = {}                                                                       │ │
│ │  model = {                                                                        │ │
│ │          │   'data': b'\x82\xa3url\xb1http://kelly.com/\xa4type\xa3far',          │ │
│ │          │   'headers': {                                                         │ │
│ │          │   │   'correlation_id': '2ac5886a-736d-4712-b878-448b8b041f43'         │ │
│ │          │   }                                                                    │ │
│ │          }                                                                        │ │
│ ╰───────────────────────────────────────────────────────────────────────────────────╯ │
╰───────────────────────────────────────────────────────────────────────────────────────╯
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x82 in position 0: invalid utf-8
```

Here's some test code:
```python

@app.after_startup
async def startup():
    faker = Faker("en_US")

    urls = [faker.url() for _ in range(10)]

    for url in tqdm.tqdm(urls):
        obj = Request(
            url=url,
            type=faker.word(),
        )
        await broker.publish(
            ormsgpack.packb(obj, option=ormsgpack.OPT_SERIALIZE_PYDANTIC),
            list="job-queue",
        )
```

The error is occurring on the producer side and the only workaround I've found so far is doing a runtime monkey-patch of the RawMessage.encode method so do the msgpack serialization at the final messaging encoding phase. This complicates things on the parsing side though as it breaks the normal message parsing (i.e. in order to get message headers, correlation_id, etc)

Any suggestions? Perhaps there needs to be additional Middleware hooks for handling the final message serialization and initial message deserialization so that serialization methods that utilize non utf-8 compatible binary are supported?

The only other option I can think of is base64 encoding the binary before message serialization which would kind of defeat the space-saving purpose of using a binary format.

Related: https://github.com/airtai/faststream/issues/1255

ilya-4real:
# Description

Fixes bug of serialization/deserialization of binary data. Logic of working with plain text and json remains the same so as not to break existing API. But when user provides binary data, 4-bytes unsigned int is added to indicate the length of headers needed by FastStream.
May be it's needed to add new type to ContentType enum to make this solution more verbose

Fixes #2061

## Type of change

## Checklist


ilya-4real:
Is it correct way to deprecate JSON format? Or should it use JSON by default and fallback to new format in case of parsing error?

faststream/redis/parser.py
```diff
@@ -90,31 +149,72 @@ def encode(
...
+                    value = reader.read_string()
+                    headers[key] = value
+
+                data = reader.read_bytes()
+            else:
+                # JSON message format
+                parsed_data = json_loads(data)
+                data = parsed_data["data"].decode()
+                headers = parsed_data["headers"]
+                warnings.warn(
```
Lancetnik: I don't think, that we should show deprecation message here. Users do not select the format they consume.

faststream/redis/parser.py
```diff
@@ -90,31 +149,72 @@ def encode(
         reply_to: Optional[str],
         headers: Optional["AnyDict"],
         correlation_id: str,
+        to_json: bool = False,
```
Lancetnik: This is not public API object, should we add json-encoder option here?
I think, we can use just a binary format

docs/docs/en/redis/message_format.md
```diff
@@ -0,0 +1,240 @@
...
+The message compiled by **FastStream** has the following structure:
+
+```txt
+# Format metadata
+[Identification header: 8 bytes]
+[Format version: 16 bit big-endian int]
+[Number of headers: 16 bit big-endian int]
+# headers
+[Header key length: 16 bit big-endian int]
+[Header key: UTF-8 string]
```
Lancetnik: Look like the format is incorrect. We should use [key:value, key:value] instead of [key, key, value, value].
Also, please use the same words everywhere: `16 bit big-endian int` or `16 big-endian bit int`

docs/docs/en/redis/message_format.md
```diff
@@ -0,0 +1,240 @@
+# Dealing with message encoded by FastStream
+
+To provide such great features as observability and many others **FastStream** needs to add extra data to your message, but suddenly **Redis** doesn't have any option to send it except the message itself. Since that, **FastStream** uses it's own binary format for messages that supports any type of data that you are going to use.
```
Lancetnik: We should note, that raw messages are OK as well, but they doesn't support headers-based features

tests/brokers/redis/test_parser.py
```diff
@@ -1,9 +1,37 @@
...
+from faststream.redis.parser import RawMessage
 from tests.brokers.base.parser import CustomParserTestcase
 
 
 @pytest.mark.redis
 class TestCustomParser(CustomParserTestcase):
     broker_class = RedisBroker
+
+
+@pytest.mark.redis
```
Lancetnik, line 78: This marker should be used only if test require connection to Redis

tests/brokers/redis/test_parser.py
```diff
@@ -1,9 +1,37 @@
...
+class TestRedisRawMessage:
```
Lancetnik: Current test can be just a function as I can see

Lancetnik (changes requested):

ilya-4real:
@Lancetnik please review

faststream/redis/broker/broker.py
```diff
@@ -202,6 +207,7 @@ def __init__(
         ] = (),
     ) -> None:
         self._producer = None
+        self.message_format = message_format
```
Lancetnik, line 218: Add deprecation warning in JSON case

faststream/redis/subscriber/factory.py
```diff
@@ -34,6 +35,7 @@ def create_subscriber(
     list: Union["ListSub", str, None],
     stream: Union["StreamSub", str, None],
     # Subscriber args
+    message_format: Type["MessageFormat"],
```
Lancetnik, line 40: Add deprecation warning in JSON case

Lancetnik (changes requested):
Please, add `message_schema` option to publisher registration as well

Lancetnik (approved):

PR #2295, opened 2025-06-14
Cli tests

borisalekseev:
# Description

Rewrite cli tests to subprocess running

### DIFF
--- a/tests/cli/conftest.py
+++ b/tests/cli/conftest.py
@@ -1,14 +1,25 @@
+import os
 import subprocess
 import threading
 import time
 from contextlib import contextmanager
-from pathlib import Path
 from textwrap import dedent
-from typing import ContextManager, Generator, List, Optional, Protocol
+from typing import (
+    TYPE_CHECKING,
+    ContextManager,
+    Dict,
+    Generator,
+    List,
+    Optional,
+    Protocol,
+)
 
 import pytest
 
 from faststream import FastStream
+
+if TYPE_CHECKING:
+    from pathlib import Path
 
 
 @pytest.fixture
@@ -37,17 +48,17 @@ def app(broker):
 class GenerateTemplateFactory(Protocol):
     def __call__(
         self, code: str, filename: str = "temp_app.py"
-    ) -> ContextManager[Path]: ...
+    ) -> ContextManager["Path"]: ...
 
 
 @pytest.fixture
 def generate_template(
-    tmp_path: Path,
+    tmp_path: "Path",
 ) -> GenerateTemplateFactory:
     @contextmanager
     def factory(
         code: str, filename: str = "temp_app.py"
-    ) -> Generator[Path, None, None]:
+    ) -> Generator["Path", None, None]:
         temp_dir = tmp_path / "faststream_templates"
         temp_dir.mkdir(exist_ok=True)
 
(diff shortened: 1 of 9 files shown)


tests/cli/test_run.py
```diff
@@ -1,210 +1,241 @@
...
+
+    @get
+    async def liveness_ping(scope):
+        return AsgiResponse(b"hello world", status_code=200)
 
+    app = AsgiFastStream(broker, asgi_routes=[
+        ("/liveness", liveness_ping),
+    ])
+    """
+    with generate_template(app_code) as app_path, faststream_cli(
```
Sehat1137, line 70: We can't use this syntax because python3.8 and python3.9 don't support it, instead of try this:

```
with ... as ...:
    with ... as ...:
```

tests/cli/test_run.py
```diff
@@ -1,210 +1,241 @@
...
+            "--factory",
+        ],
+        extra_env={
+            "PATH": f"{app_path.parent}:{os.environ['PATH']}",
+            "PYTHONPATH": str(app_path.parent),
+        },
+        wait_time=0.5,
+    ) as cli_thread:
+        pass
+    assert cli_thread.process.returncode != 0
```
Sehat1137: I'm not sure that we need this test.
@Lancetnik what do you think?

Sehat1137 (changes requested):

tests/cli/test_run.py
Lancetnik: I think the same

tests/cli/conftest.py
```diff
@@ -1,15 +1,20 @@
+from __future__ import annotations
```
Lancetnik: Please, do not use it. It is just a convetion – we don't using future anywahere, so we shouldn't use it here

tests/cli/test_run.py
```diff
@@ -1,210 +1,279 @@
...
+
+    @get
+    async def liveness_ping(scope):
+        return AsgiResponse(b"hello world", status_code=200)
+
+    app = AsgiFastStream(broker, asgi_routes=[
+        ("/liveness", liveness_ping),
+    ])
+    """
+    with contextlib.ExitStack() as exit_stack:
```
Lancetnik: I don't think, that is required

`with (generate_template(app_code) as app_path, faststream_cli(...))` should works fine

Another way – we can mark it test by python 3.10+ – only
It's okay cuz we are schedule to remove 3.8 and 3.9 in nearest 0.6 release

tests/cli/test_run.py
```diff
@@ -1,210 +1,279 @@
...
+                extra_env={
+                    "PATH": f"{app_path.parent}:{os.environ['PATH']}",
+                    "PYTHONPATH": str(app_path.parent),
+                },
+            )
         )
-        app.run.assert_awaited_once_with(
-            logging.INFO, {"host": "0.0.0.0", "port": "8000"}
+        response = exit_stack.enter_context(
+            urllib.request.urlopen(  # nosemgrep: python.lang.security.audit.insecure-transport.urllib.insecure-urlopen.insecure-urlopen
```
Lancetnik: I think, we can just add `tests/*` to `.semgrepignore` file

tests/cli/test_run.py
```diff
@@ -1,210 +1,279 @@
...
 def test_run_as_asgi_with_many_workers(
-    runner: CliRunner, workers: int, app: Application
+    generate_template,
+    faststream_cli,
+    workers: int,
 ):
-    asgi_multiprocess = "faststream.cli.supervisors.asgi_multiprocess.ASGIMultiprocess"
-    _import_obj_or_factory = "faststream.cli.utils.imports._import_obj_or_factory"
+    app_code = """
+    from faststream.asgi import AsgiFastStream, AsgiResponse, get
```
Lancetnik: Please, do no import useless objects and do not write useless for tests code to make code example as clear as possible

Lancetnik (changes requested):

Lancetnik (approved):

Lancetnik (approved):

PR #2352, opened 2025-07-13
fix: fixed error handling in asgi application lifespan

### LINKED ISSUE #2305: Bug: Faststream does not handle exception in ASGI startup lifespan correctly
**Describe the bug**
When running the ASGI application and the startup failed, the server will hang forever until you press CTRL+C. I think this behaviour is unintentionally and also is different to how FastAPI handles this case. In FastAPI an exception within startup will send a proper ASGI `{"type": "lifespan.startup.failed"}` event. In Faststream, this never happens as exceptions thrown in `_startup` method are not caught and properly handled which **always** sends a successfull startup event.

Here is the code:

https://github.com/ag2ai/faststream/blob/7c5dffc5c1f132af1cddb228979f39f91b2dfa10/faststream/asgi/app.py#L184-L215

The `started = False` variable lets me believe that this is a bug since `started` is not correctly set, since startup actually failed but `started` is `True`

**How to reproduce**
Run this script via `uv run script.py` or `uv run faststream script:app` or `uv run uvicorn script:app`.

```python
# /// script
# requires-python = ">=3.10"
# dependencies = [
#     "faststream",
# ]
# ///
from faststream.asgi import AsgiFastStream
from faststream.nats.broker import NatsBroker

broker = NatsBroker()
app = AsgiFastStream(broker)


@app.on_startup
async def setup():
    raise ValueError()


if __name__ == "__main__":
    import uvicorn

    uvicorn.run(app)

```

**Expected behavior**
Faststream should send correctly failed startup event. Uvicorn should exit in that case and not run for ever. 

Note: This would also happen in any other asgi compatible server than uvicorn.

**Observed behavior**
Faststream exists and uvicorn server runs for ever

**Environment**
All versions

**Additional context**
I think there is a simple fix:

```python
    @asynccontextmanager
    async def start_lifespan_context(self) -> AsyncIterator[None]:
        async with self.lifespan_context():
            try:
                await self._startup(self._log_level, self._run_extra_options)
                yield True # yield True only if startup is done
            except BaseException:
                self._log(self._log_level, "Fastream app encountered exception during startup")
                raise
            finally:
                await self._shutdown(self._log_level)

    async def lifespan(self, scope: "Scope", receive: "Receive", send: "Send") -> None:
        """Handle ASGI lifespan messa:ges to start and shutdown the app."""
        started = False
        await receive()  # handle `lifespan.startup` event

        try:
            async with self.start_lifespan_context() as started:
                await send({"type": "lifespan.startup.complete"})
                await receive()  # handle `lifespan.shutdown` event

        except BaseException:
            exc_text = traceback.format_exc()
            # started should only be True if contextmanager did not throw exception
            if started:
                await send({"type": "lifespan.shutdown.failed", "message": exc_text})
            else:
                await send({"type": "lifespan.startup.failed", "message": exc_text})
            raise

        else:
            await send({"type": "lifespan.shutdown.complete"})
```

This might not be 100% what you want, since in some cases you might want to wait for the startup to finish. However, when I run the script with the modified code from above, the uvicorn server exits correctly.

marcm-ml:
# Description

Fixes exception handling in ASGI application startup. The previous implementation had the side-effect that exception within startup (i.e. broker failure), would cause the underlying ASGI server to get stuck waiting for either the "lifespan.startup.complete" or "lifespan.startup.failed" event. In this state, the ASGI server would ignore all signals (like SIGINT) and becomes unresponsive.

Fixes #2305 

## Type of change

## Checklist


marcm-ml:
When running tests on my computer, I get unrelated errors. I assume these are flaky tests? However, the new tests run perfectly fine.

marcm-ml:
@Lancetnik feel free to review whenever you find the time. This should fix any of the issues discussed in the linked issue. Thanks

Lancetnik (changes requested):
Added tests doesn't proof related changes. All new tests are green with the old code as well

marcm-ml:
How did you test that? If i copy over the test cases into main, I get 4 failing tests as expected:
```python
FAILED tests/asgi/nats/test_asgi.py::TestNatsAsgi::test_lifespan_startup_failure - ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
FAILED tests/asgi/nats/test_asgi.py::TestNatsAsgi::test_lifespan_shutdown_failure - ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
FAILED tests/asgi/nats/test_asgi.py::TestNatsAsgi::test_on_startup_failure - ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
FAILED tests/asgi/nats/test_asgi.py::TestNatsAsgi::test_on_shutdown_failure - ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
```

marcm-ml:
Ah okay, i might understand the problem now.
The behaviour of the TestClient and running things via uvicorn is different.
When I run the example script from the issue with uvicorn, the uvicorn server lives forever with the old code and exists immediately with the new code. In both cases the faststream app indicates a shutdown.
If I replace the uvicorn server with the TestClient, I get an immediate exit with the old and new code.

```python
from time import sleep

from faststream.asgi import AsgiFastStream
from faststream.nats.broker import NatsBroker

broker = NatsBroker()
app = AsgiFastStream(broker)


@app.on_startup
async def setup():
    raise ValueError()


if __name__ == "__main__":
    from starlette.testclient import TestClient

    with TestClient(app, base_url="http://localhost:8888") as client:
        sleep(100)
```

This leaves me with the question, how I could test this properly. Perhaps, I could add a test case in the CLI test suite which uses uvicorn under the hood? Do you have other ideas? I think the current tests are useless as you mentioned

Lancetnik:
> This leaves me with the question, how I could test this properly. Perhaps, I could add a test case in the CLI test suite which uses uvicorn under the hood? Do you have other ideas? I think the current tests are useless as you mentioned

Thank you!
I hope, @Sehat1137 will help us here. He is a codeowner of ASGI CLI support.

marcm-ml:
I have added a better test that now fails with the old code and succeeds with the new code.
Funny enough, when using lifespan, the startup is correctly aborted.
```python
from contextlib import asynccontextmanager

from faststream.asgi import AsgiFastStream
from faststream.nats.broker import NatsBroker

broker = NatsBroker()


@asynccontextmanager
async def lifespan():
    raise ValueError("Lifespan error")
    yield


app = AsgiFastStream(broker, lifespan=lifespan)

if __name__ == "__main__":
    import uvicorn

    uvicorn.run(app)
```

marcm-ml:
@Sehat1137 do you know why the test_run test failed? When I am running this locally (and devcontainer), all tests pass

Edit:
also now that a broker failure will crash the ASGI startup, some tests fail since the broker cannot establish a proper connection. This was not a problem before as the broker startup was done after the lifespan startup event has been sent. More specifically, since the NatsBroker tries to reconnect 60 times each waiting for 2 seconds, the startup is super slow in this case. Before the ASGI app became immediately available but the broker did not.

Is this something that breaks the current behaviour of the ASGI app in your eyes @Lancetnik ?

tests/cli/test_run.py
```diff
@@ -350,3 +350,75 @@ def print_log_level():
...
+
+    assert cli_thread.process
+    stderr = cli_thread.process.stderr.read()
+    assert f'ValueError("Failure during {failure_type}")' in stderr
+    if failure_type in ("startup", "lifespan_start"):
+        # this comes from uvicorn and only is logged when "lifespan.startup.failed" event is send
+        assert "Application startup failed. Exiting." in stderr
+    elif failure_type in ("shutdown", "lifespan_shutdown"):
+        # this comes from uvicorn and only is logged when "lifespan.shutdown.failed" event is send
+        assert "Application shutdown failed. Exiting." in stderr
```
Sehat1137: Hi, I don't think this is a suitable test.
This module tests the `faststream run ...` command and checks basic application launch cases.

We need something like [this](https://github.com/ag2ai/faststream/blob/main/tests/cli/rabbit/test_app.py)

Sehat1137:
Please, rebase your branch

marcm-ml:
Hey, I have added the tests in tests/cli/rabbit/test_app.py. I had to call uvicorn directly to make sure I get access to the uvicron server instance to check if the appropiate lifespan hook has been called.

Additionally, I had to refactor how tests/cli/test_run.py was setup. Please take a look if you are happy with the result.

Sehat1137:
@marcm-ml 
Previously, this worked without using mocks.
If you had to edit existing tests, you most likely broke backward compatibility.
@Lancetnik what we can do with it?

marcm-ml:
> also now that a broker failure will crash the ASGI startup, some tests fail since the broker cannot establish a proper connection. This was not a problem before as the broker startup was done after the lifespan startup event has been sent. More specifically, since the NatsBroker tries to reconnect 60 times each waiting for 2 seconds, the startup is super slow in this case. Before the ASGI app became immediately available but the broker did not.

I tried to explain the reason previously here. I think it was not apparent that the broker setup in this specific test was _always_ unsuccessful since the broker startup was handled **after** asgi startup.

My original issue was created exactly due to this behaviour where a broker setup was not successful but the asgi server kept running.

Edit:
If you replace the NatsBroker in tests/cli/test_run.py with RabbitBroker, you will get failing tests due to connection issues with the current main branch.

Sehat1137:
@marcm-ml 
Yes, now I understand better what the problem is. Thank you, I will write a little later when I think about how to solve this situation

Sehat1137:
@marcm-ml 
I tried to run the simple application if nats is not available. The application freezes and cannot be stopped using ctrl+c.

Previously, such behavior was not observed. But this actual only for NatsBroker.

You may try to repeat this case
```python
from faststream.nats import NatsBroker  # Not ok
from faststream.rabbit import RabbitBroker  # ok
from faststream.redis import RedisBroker  # ok
from faststream.kafka import KafkaBroker  # ok


broker = NatsBroker()

app = FastStream(broker)
```

marcm-ml:
for me every broker is failing. Nats is appears stuck because it has a high reconnect attempt setting. if you set max_reconnect_attempts=1 than nats fails instantly for me. Redis is also stuck for me. Kafka and Rabbit exit instantly.

marcm-ml:
Maybe to clarify:
I get errors for all broker in main and in my code with your example.

The behaviour is expected that ctrl+c does not work in my code but works in main branch. We are now awaiting the startup and not sending it off as a _background_ task using anyio task group. During the startup lifespan, uvicorn traps any signal (like ctrl+c / SIGINT) until startup is done. since the startup is only done after the broker has been setup, we cannot abort the startup. It is unfortunate that nats default reconnect parameter is set as high (60*2s = 2mins), which means u need to wait 2 mins to exit. I argue that this was the reason the current implementation was setup in that way since it appears that the application is unresponsive. I would argue that this is a issue with nats implementation rather than the ASGI one. 

I think it is up to you guys to decide what kind of behaviour you would want to see here
@Lancetnik @Sehat1137 

Edit:

You can see the same behaviour without using fastream. Run via `uvicorn script:app` and try to press CTRL+C
```python
import asyncio
from contextlib import asynccontextmanager

from fastapi import FastAPI

@asynccontextmanager
async def lifespan(app):
    await asyncio.sleep(10) # this is equivalent to nats broker trying to connect
    yield

app = FastAPI(lifespan=lifespan)
```

Sehat1137:
@marcm-ml 
Maybe you can think about how to solve this problem without breaking the current behavior? If you don't want, you may create an issue and we try to solve this in near releases

marcm-ml:
I believe the current implementation is fundamentally broken and not how an ASGI application should be started. If you want to retain the current behaviour, you must make sure that the `self._startup` does not block. Then you need to figure out how to shutdown the ASGI server when an exception on self._startup was raised. For me this makes things more difficult since the broker setup should be part of the ASGI setup.
On a different note: Regarding semantic versioning, your are still in an early development phase (0.x.x) which should indicate to users that they need to expect breaking changes.

Lancetnik:
@marcm-ml sorry, I didn't investigate for the problem too close before

I can't agree, that broker startup error should fails ASGI runtime

```python
from faststream.asgi import AsgiFastStream, AsgiResponse, get

@get
async def index(scope):
    return AsgiResponse(b"Hello, World!", 200)

broker = NatsBroker()
app = AsgiFastStream(broker, asgi_routes=[("/test", index)])
```

The following code process `/test` endpoint correctly even if NatsBroker is not connected – so I can decide, that ASGI works fine and we shouldn't fail. Broker connection is in another application scope and should be checked by another mechanism (like probes)

But, lifespan errors should fail the app according ASGI specification. I suggest to split your Issue into two:
1. Fail ASGI startup if broker is not available – I think, we should reject it
2. Fail ASGI statup if lifespan error occured – this is the bug indeed and we should fix it

Lancetnik:
> On a different note: Regarding semantic versioning, your are still in an early development phase (0.x.x) which should indicate to users that they need to expect breaking changes.

It is not correct. We follow semantic versioning, so in our case it means that versions under 1.0.0 has a majority in 0.X release. Therefore we can't break compatibility in patches like `0.5.*` – all breaking changes should be made in next (or even later) major `0.6.0` for now.

Additionally, we have many users and a large community, so we cannot break compatibility without a deprecation phase. Any breaking changes must have the following lifespan:

1. New feature introduction
2. Old syntax must still be supported, deprecated, and mapped to the new feature. We must support both versions (new and old) for at least one major release
3. After that, we can remove the old syntax with a deprecation notice

Lancetnik:
The problem is, that we call `on_startup`, `broker.start` and `after_startup` as a single call here:

https://github.com/ag2ai/faststream/blob/main/faststream/_internal/application.py#L168-L180

And we move this call to asyncio task in ASGI – https://github.com/ag2ai/faststream/blob/main/faststream/asgi/app.py#L189

This reason, the following code doesn't fail ASGI application startup

```python
@app.on_startup
async def setup():
    raise ValueError()
```

To fix it we should separate `on_startup` from original `start` method and call it is a part of ASGI lifespan. But, `broker.start` and `after_startup` should stay in task

marcm-ml:
> > On a different note: Regarding semantic versioning, your are still in an early development phase (0.x.x) which should indicate to users that they need to expect breaking changes.
> 
> 
> 
> It is not correct. We follow semantic versioning, so in our case it means that versions under 1.0.0 has a majority in 0.X release. Therefore we can't break compatibility in patches like `0.5.*` – all breaking changes should be made in next (or even later) major `0.6.0` for now.
> 
> 
> 
> Additionally, we have many users and a large community, so we cannot break compatibility without a deprecation phase. Any breaking changes must have the following lifespan:
> 
> 
> 
> 1. New feature introduction
> 
> 2. Old syntax must still be supported, deprecated, and mapped to the new feature. We must support both versions (new and old) for at least one major release
> 
> 3. After that, we can remove the old syntax with a deprecation notice

Point 4. of the semantic versioning spec states that version 0.x is initial development and subject to any change.

https://semver.org/#spec-item-4

Lancetnik:
> Point 4. of the semantic versioning spec states that version 0.x is initial development and subject to any change.
> 
> [semver.org](https://semver.org)

Yeah, I had read the specification. Anyway, we conclude to use minor patches as major until 1.0 release. So, we introduce any minor features & patches as patch releases

marcm-ml:
I am confused that broker startup should not be part of the ASGI startup.

But if you do not want to change that behaviour, it is fine with me. 
You can close the PR if you want. Thanks for your time 👍

Lancetnik:
> I am confused that broker startup should not be part of the ASGI startup.
> 
> But if you do not want to change that behaviour, it is fine with me. You can close the PR if you want. Thanks for your time 

Thank you for your time and effort in investigating the issue. You have done a great job, and I apologize for not clarifying things sooner. I attempted to clarify my thoughts and concerns, but we can still continue to discuss them. This is just my perspective.

In any case, the original issue was regarding

```python
@app.on_startup
async def setup():
    raise ValueError()
```

which does not fail the application. The issue still exists and needs to be addressed, so I do not see any reason for closing the issue at this time. But, if you don't want to work on it, you can close the PR. Thank you and sorry one more time.

marcm-ml:
I can look into it in a few days again. Will let you know 👍

Sehat1137:
@marcm-ml Hi, any updates?

marcm-ml:
@Lancetnik @Sehat1137 this isssue is fixed now for both FastStream and Asgi Applications. I guess some of the recent changes fixed this issue 👍

Lancetnik:
> @Lancetnik @Sehat1137 this isssue is fixed now for both FastStream and Asgi Applications. I guess some of the recent changes fixed this issue 👍 

Great to know, thank you! Did you check it with 0.6 version?

marcm-ml:
Yes I tested main and 0.6.0rc0 👍

PR #2373, opened 2025-07-19
fix: Change f-string in logs

GredFIX:
# Description

My project includes loguru, which intercepts logs for their centralized processing. But in the place where I made changes, a lot is inserted without preliminary processing, which causes a KeyError in loguru. The error is not critical and does not crash the application, but purely visually I do not want to see an error message when starting the application

## Type of change

## Checklist

### DIFF
--- a/faststream/kafka/helpers/rebalance_listener.py
+++ b/faststream/kafka/helpers/rebalance_listener.py
@@ -68,7 +68,7 @@ class _LoggingListener(ConsumerRebalanceListener):  # type: ignore[misc]
         self.logger.log(
             logging.INFO,
             f"Consumer {self.consumer._coordinator.member_id} assigned to partitions: "
-            f"{assigned}",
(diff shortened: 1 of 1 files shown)


Lancetnik (approved):

Tishka17:
The root of the problem is that loguru is NOT compatible with current logging protocol. You could think differently, but it is not. You cannot use loguru logger directly here. 

In details. If you call `logger.log(level, XXX , extras)` then loguru treats XXX as a format string, not a message. Correct call for loguru is `logger.log(level, "{}", XXX, extras)`. Instead of patching library should create an adapter which transforms one call to another.

E.g.
```python
class MyLogger:
   def __init__(self, loguru_logger):
      self.logger = loguru_logger
  def log(self, level, msg, extras)
      self.logger.log(level, "{}", msg, extras)
````

Or just use logging.Logger here, as normally you have an integration between loguru and logging set up

PR #2387, opened 2025-07-28
Remove `on: workflow_dispatch: null`

sobolevn:
Is it really needed? We use just `on: released: [published]` in https://github.com/typeddjango/django-stubs/blob/master/.github/workflows/release.yml#L3-L5


Lancetnik:
Yes, I need it

PR #2416, opened 2025-08-06
CI: use pre-commit-ci-lite instead of manual commit step during linters

### LINKED ISSUE #2410: ci: pre-commit job checks
* [ ] We should migrate from https://github.com/pre-commit/action to https://pre-commit.ci/lite.html in https://github.com/ag2ai/faststream/blob/main/.github/workflows/pr_autoupdate.yaml.
* [ ] If the job fails, lock the PR to prevent accidental merges.
* [ ] Make sure the workflow is running successfully in the main branch.

kittywaresz:
# Description

Now `Run linters` workflow applies autofix commit using ready to go solution based on [pre-commit.ci](https://pre-commit.ci/)
Also now workflow triggered on `pull_request` event, it means more secured context of steps execution without access to base repository

Fixes #2410

## Type of change

## Checklist

### DIFF
--- a/.github/workflows/pr_autoupdate.yaml
+++ b/.github/workflows/pr_autoupdate.yaml
@@ -1,7 +1,7 @@
 name: Run linters
 
 on:
-  pull_request_target:
+  pull_request:
     types:
       - opened
       - synchronize
@@ -20,11 +20,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-        with:
-          token: ${{ secrets.AUTOMERGE_TOKEN }}
-          repository: ${{ github.event.pull_request.head.repo.full_name }}
-          ref: ${{ github.head_ref }}
-          persist-credentials: false
 
       - uses: astral-sh/setup-uv@e92bafb6253dcd438e0484186d7669ea7a8ca1cc # v6.4.3
         with:
@@ -56,10 +51,8 @@ jobs:
         with:
           extra_args: --hook-stage manual --all-files
 
-      - name: Commit
-        uses: stefanzweifel/git-auto-commit-action@778341af668090896ca464160c2def5d1d1a3eb0 # v6.0.1
-        with:
-          commit_message: "chore: apply autofixes"
-          commit_user_name: faststream-actions[bot]
-          commit_user_email: faststream-actions[bot]@users.noreply.github.com
-          commit_author: faststream-actions[bot] <faststream-actions[bot]@users.noreply.github.com>
+      # the step must have either the default name or contain the text 'pre-commit-ci-lite'.
+      # the application uses this to find the right workflow.
+      - name: Run pre-commit-ci-lite
+        uses: pre-commit-ci/lite-action@5d6cc0eb514c891a40562a58a8e71576c5c7fb43 # v1.1.0
+        if: always()
--- a/.github/workflows/pr_autoupdate.yaml
+++ b/.github/workflows/pr_autoupdate.yaml
@@ -45,6 +45,12 @@ jobs:
         with:
           just-version: 1.42.4
 
+      - name: Run pre-commit
+        uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
+        continue-on-error: true
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_autoupdate.yaml
```diff
@@ -1,7 +1,7 @@
 name: Run linters
 
 on:
-  pull_request_target:
+  pull_request:
```
kittywaresz, line 4: otherwise `pre-commit-ci/lite-action` wont run, it's [hardcoded here](https://github.com/pre-commit-ci/lite-action/blob/5d6cc0eb514c891a40562a58a8e71576c5c7fb43/main.mjs#L8)

.github/workflows/pr_autoupdate.yaml
```diff
@@ -20,11 +20,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-        with:
```
kittywaresz, line 23: I think we don't need it since we use `pull_request` event

.github/workflows/pr_autoupdate.yaml
```diff
@@ -56,10 +51,8 @@ jobs:
         with:
           extra_args: --hook-stage manual --all-files
 
-      - name: Commit
```
kittywaresz, line 59: @Lancetnik could you please explain, how it works? Or show example PRs with autofixes

I dont't understand, if pre-commit makes any changes, where these changes go? Where this commit will appear, directly in fork?

I am not so experienced in Github Actions, so I try to build strong understanding of current workflow to make sure anything will work as it was before

.github/workflows/pr_autoupdate.yaml
Lancetnik: Great choice!

.github/workflows/pr_autoupdate.yaml
Lancetnik: CI just commits applied changes directly to PR to not bother PR author about fixes could be solved automatically.

E.g. https://github.com/ag2ai/faststream/pull/2393

kittywaresz: Thank you, now I think it should work as expected

kittywaresz:
> If the job fails, lock the PR to prevent accidental merges.

also, could you please clarify what do you mean by "lock"?

> Make sure the workflow is running successfully in the main branch.

and if I [tested it](https://github.com/kittywaresz/faststream/pull/11/commits) on a main branch of my fork, does it counts? :thinking:

Lancetnik:
> also, could you please clarify what do you mean by "lock"?

CI should fail if we have any not-fixable jobs in pre-commit. Currently it passes any way

.github/workflows/pr_autoupdate.yaml
```diff
@@ -56,10 +51,8 @@ jobs:
         with:
           extra_args: --hook-stage manual --all-files
```
Lancetnik, line 52: Why you didn't remove old precommit run? Seems like we running it twice in CI

Lancetnik (dismissed):

.github/workflows/pr_autoupdate.yaml
kittywaresz: It seems weird, but I tested in my fork and `pre-commit-ci/lite-action` did nothing alone, it make commit only if it runs after `pre-commit/action`

Also please check an example in this action docs - https://github.com/pre-commit-ci/lite-action?tab=readme-ov-file#example-using-pre-commitaction, there are both actions presetn

kittywaresz: Let me show the difference:

- https://github.com/kittywaresz/faststream/actions/runs/16814067838 - checks in PR to branch with only pre-commit-ci-lite action
- https://github.com/kittywaresz/faststream/actions/runs/16814136250 - checks in PR to branch with both actions present

Lancetnik (approved):

Lancetnik:
@kittywaresz CI still passing even precommit fail – https://github.com/ag2ai/faststream/actions/runs/16863843703/job/47768003241?pr=2429

Can you please take a look?

kittywaresz:
@Lancetnik sure, I will take a look

Sorry for the late reply, it seems that notifications don't work for closed PRs

PR #2445, opened 2025-08-12
fix(asyncapi): promote nested pydantic  to components/schemas

### LINKED ISSUE #2443: bug: Nested Pydantic model are not included in AsyncAPI components/schemas (inplaced instead)
**Describe the bug**
With AsyncAPI 3.0, when registering multiple handlers on the same publishers with some of the return types having nested Pydantic models inside them, these nested models are not defined correctly in the AsyncAPI spec resulting in an invalid state (the model referenced does not exist)

**How to reproduce**
Include source code:

```python
class User(pydantic.BaseModel):
    class Email(pydantic.BaseModel):
        email: str

    email: Email
    name: str = ""
    id: int
    

broker = self.broker_class()

publisher = broker.publisher("test")

@publisher
def handle0(msg) -> User:
    ...


class Other(pydantic.BaseModel):
    id: int

@publisher
def handle1(msg) -> Other:
    ...

schema = self.get_spec(broker).to_jsonable()

payload = schema["components"]["schemas"]

assert "User" in payload
assert "Other" in payload

assert "Email" in payload # <-- Assert fails
```

**Expected behavior**
The AsyncAPI schema should be valid.

**Environment**
main branch / 0.6.0rc0

**Additional context**
I will try to find a workaround/bugfix for now

legau:
# Description

Promote nested Pydantic model definitions from message payloads (including oneOf cases) into components/schemas and replace inline schemas with $ref references. This resolves incorrect “inplace” embedding of nested models in generated AsyncAPI 3.0 documents.

Not sure if it is a good implementation, however tests are passing

Fixes #2443

## Type of change

## Checklist

### DIFF
--- a/tests/asyncapi/base/v3_0_0/publisher.py
+++ b/tests/asyncapi/base/v3_0_0/publisher.py
@@ -121,31 +121,3 @@ class PublisherTestcase(AsyncAPI300Factory):
         schema = self.get_spec(broker).to_jsonable()
 
         assert schema["channels"] == {}, schema["channels"]
-
-    def test_nested_models_in_one_of_should_be_in_schemas(self) -> None:
-        class Email(pydantic.BaseModel):
-            email: str
-
-        class User(pydantic.BaseModel):
-            email: Email
-            name: str = ""
-            id: int
-
-        broker = self.broker_class()
-
-        publisher = broker.publisher("test")
-
-        @publisher
-        def handle0(msg) -> User: ...
-
-        class Other(pydantic.BaseModel):
-            id: int
-
-        @publisher
-        def handle1(msg) -> Other: ...
-
-        schema = self.get_spec(broker).to_jsonable()
-
-        payload = schema["components"]["schemas"]
-
-        assert "Email" in payload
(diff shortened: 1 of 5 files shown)


tests/asyncapi/base/v3_0_0/publisher.py
```diff
@@ -121,3 +121,31 @@ async def handler(msg: str) -> None:
         schema = self.get_spec(broker).to_jsonable()
 
         assert schema["channels"] == {}, schema["channels"]
+
+    def test_nested_models_in_one_of_should_be_in_schemas(self) -> None:
```
Lancetnik: The current problem is not publisher-specific. As well it is not specific for 3.0.0 AsyncAPI version

It should exists for subscribers as well for (or with descriminator probably)

```python
@subscibrer
async def handler(body: Model1 | Model2): ... 
```

So, can you please add tests and fix for `3.0.0` and `2.6.0` versions both? And place them somewhere near `test_pydantic_model_mixed_regular`. Alse, please add a link to the original Issue as a test docstring

legau: This should be good

PR #2459, opened 2025-08-15
Fix: Handle non-UTF-8 headers in Kafka/Confluent message parsing

### LINKED ISSUE #2214: Feature: allow defining kafka headers as bytes

I'm running a kafka cluster with multiple consumers using shared interface for decoding and validating message body and headers. I've started experimenting with FastStream (0.5.40) for one of the message publishers and it's been generally a good experience, but being forced to pass all headers as strings is limiting.

There are some custom headers that each of my messages need to include, one of them being an `event_id`. Consumers try to decode it as `UUID(bytes=headers[HEADER_EVENT_ID])`. This means it needs to be encoded as `UUID.bytes`, which internally converts it to int, the runs `int.to_bytes(16)`.

It would be great if we could define headers as a dict of bytes instead. Then we can freely control how they are encoded.

**Feature code example**

```python
from uuid6 import uuid6
...
message_body = {...}
partition_key = "partition_key"

version = 0
event_id = uuid6()
message_type = "msg_type_1"

message_headers: dict[str, bytes] = {
  "version": version.to_bytes(),
  "event_id": event_id.bytes,
  "message_type": message_type.encode("utf-8"),
  "some_other_header": b"",
}

await kafka_publisher.publish(
  message=message_body,
  key=partition_key,
  headers=message_headers,
)
```

The only way I was able to work around this, was to add custom behaviour on the consumer side.

### LINKED ISSUE #2458: Bug:FastStream fails to decode non-UTF-8 Kafka message headers
**Describe the bug**
When processing Kafka messages in FastStream, a UTF-8 decoding error occurs while parsing message headers. The issue arises in faststream/confluent/parser.py when headers contain non-UTF-8 byte sequences.

**How to reproduce**

**1. Produce a Kafka Message with Non-UTF-8 Headers**
Send a message to a Kafka topic containing headers with non-UTF-8 byte sequences:
```python
import json
from confluent_kafka import Producer

if __name__ == "__main__":
    kafka_producer = Producer({"bootstrap.servers": "localhost:9092"})
    kafka_producer.produce(
        topic="test",
        value=json.dumps({"some_field": "some_data"}),
        headers={"trash_header": b"\xc3\x28"}  # Invalid UTF-8 sequence
    )
    kafka_producer.flush()
``` 
**2. Consume the Message Using FastStream**
Subscribe to the topic and attempt to process the message with non-UTF-8 headers:
```python
from faststream.confluent import KafkaBroker, KafkaMessage

broker = KafkaBroker()

@broker.subscriber("test", auto_offset_reset="earliest")
async def handle_message(message: KafkaMessage, logger: Logger) -> None:
    logger.info(f"Message headers: {message.headers}")

app = FastStream(broker)
```

**Expected behavior**
FastStream should either:
* Correctly handle message headers with different encodings, or
* Provide a way to specify the encoding for header decoding.

**Observed Behavior**
A UnicodeDecodeError is raised when trying to decode message headers as UTF-8:
```bash
 File ".../.venv/lib/python3.11/site-packages/faststream/confluent/parser.py", line 96, in _parse_msg_headers
    return {i: j if isinstance(j, str) else j.decode() for i, j in headers}
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.11/site-packages/faststream/confluent/parser.py", line 96, in <dictcomp>
    return {i: j if isinstance(j, str) else j.decode() for i, j in headers}
                                            ^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc3 in position 0: invalid continuation byte
```

**Environment**
```python
faststream >=0.5.48,<0.6.0
```

Bazarovinc:
This PR addresses the UnicodeDecodeError that occurred when processing Kafka messages containing headers with non-UTF-8 byte sequences

Fixes #2458, FIxes #2214

## Type of change

## Checklist


draincoder:
@Bazarovinc Please look at this https://github.com/ag2ai/faststream/issues/2214 issue as well and add comprehensive tests.

Bazarovinc:
@draincoder Here’s a preliminary implementation for handling Kafka message headers. Please review and let me know if this approach looks good or if any adjustments are needed.

faststream/confluent/parser.py
```diff
@@ -28,7 +27,15 @@ async def parse_message(
...
     ) -> KafkaMessage:
         """Parses a Kafka message."""
-        headers = _parse_msg_headers(message.headers() or ())
+        headers = dict(message.headers() or ())
+        decoded_headers = {
+            header: value.decode(errors="replace")
+            for header, value in (
+                headers.get("reply_to"),
+                headers.get("content-type"),
+                headers.get("correlation_id"),
```
Sehat1137: what does `headers.get` return? Can you show example or write type?

faststream/confluent/parser.py
```diff
@@ -69,10 +76,16 @@ async def parse_message_batch(
             body=body,
             headers=headers,
             batch_headers=batch_headers,
-            reply_to=headers.get("reply_to", ""),
-            content_type=headers.get("content-type"),
+            reply_to=headers.get("reply_to").decode()
+            if "content-type" in headers
```
Sehat1137: content-type -> reply_to

faststream/confluent/parser.py
```diff
@@ -69,10 +76,16 @@ async def parse_message_batch(
...
+            else None,
+            content_type=headers.get("content-type").decode()
+            if "content-type" in headers
+            else None,
             message_id=f"{first.offset()}-{last.offset()}-{first_timestamp}",
-            correlation_id=headers.get("correlation_id"),
+            correlation_id=headers.get("correlation_id").decode()
+            if "correlation_id" in headers
+            else None,
```
Sehat1137: What do u this about this way for better readable:
```python
reply_to=headers.get("reply_to", b"").decode() or None
content_type=headers.get("content-type", b"").decode() or None
correlation_id=headers.get("correlation_id", b"").decode() or None
```

Sehat1137 (changes requested):

faststream/confluent/parser.py
ozeranskii: I suggest doing it this way
```python
@staticmethod
def _decode_header(headers: dict[str, bytes | None], key: str) -> str | None:
    val = headers.get(key)
    return val.decode(errors="replace") if val else None
```
and then, where we need to get the value, we do this 
```python
reply_to=self._decode_header(headers, "reply_to")
content_type=self._decode_header(headers, "content-type")
correlation_id=self._decode_header(headers, "correlation_id")
```

This is already applicable in `parse_message` and `parse_message_batch` from what I've noticed.

faststream/confluent/parser.py
ozeranskii: A small note: comprehension iterates over values (`bytes`) rather than pairs `(key, value)`, so the expression `for header, value in (...)` will raise a `ValueError`.
It is worth iterating over `headers.items()` or decoding values using a separate helper.

```python
decoded_headers = {
    k: v.decode(errors="replace")
    for k, v in headers.items()
    if k in ("reply_to", "content-type", "correlation_id") and v
}
```

Or, even better and clearer, since iteration here seems excessive to me and makes the code difficult to understand. See https://github.com/ag2ai/faststream/pull/2459/files/d0ee2c6911627934e78d958536e428ebb2cb1ec6#r2427545938

```python
reply_to = self._decode_header(headers, "reply_to")
content_type = self._decode_header(headers, "content-type")
correlation_id = self._decode_header(headers, "correlation_id")
```

fstermann:
@Bazarovinc @Sehat1137 @draincoder Any updates on this PR? We are currently stumbling into the same original issue and needed to implement a workaround for it. Would be nice to have this included in a near future release, thanks for an update! :)

Issue #1798, opened 2024-09-16
Feature: Add DI for ASGI routes


Sehat1137:
**Is your feature request related to a problem? Please describe.**
My app may have different dependencies that I want check using k8s probes (database, s3, etc)

**Describe the solution you'd like**
I need possibility using DI for access to my dependency

**Feature code example**

```python
from faststream import FastStream, Depends
from faststream.rabbit import RabbitBroker

broker = RabbitBroker("amqp://guest:guest@localhost:5672/")
app = FastStream(broker)

def simple_dependency() -> int:
    return 1

async def readiness(scope, receive, send, d: int = Depends(simple_dependency)):
    assert d == 1
    return AsgiResponse(b"", status_code=200)

app = FastStream(broker).as_asgi(
    asgi_routes=[
        ("/readiness", readiness),
    ],
    asyncapi_path="/docs",
)
```

nightblure:
@Sehat1137 hi! 
this seems to work with the injection decorator:

```python3
from fast_depends import inject

@inject
async def readiness(scope, receive, send, d: int = Depends(simple_dependency)):
    assert d == 1
    return AsgiResponse(b"", status_code=200)
```

PR #2467, opened 2025-08-21
feat: asgi request

borisalekseev:
# Description

Fixes https://github.com/ag2ai/faststream/issues/1798

TODO:
- [x] `Depends` tests
- [x] remove hasattr
- [ ] check if possible move "Scope" to type checking import

Backward compatibility with old HTTP handlers that receive a scope argument is preserved.

Example usage:

```python3
import logging

from faststream.asgi import AsgiFastStream, get, AsgiResponse, post, Request, Header, Query

logger = logging.getLogger("faststream")


@get
async def access_request(request: Request):
    logger.info("Received request %s", request)
    # Received request <faststream.asgi.request.AsgiRequest object at 0x7cda02c674d0>
    return AsgiResponse(status_code=200)


@get
async def access_query(faststream: list[str] | None = Query(default=None)):
    logger.info("Faststream is %s", faststream)
    # Faststream is ['cool']
    return AsgiResponse()


# curl -X POST http://127.0.0.1:8000/access_header --header "X-auth-token: secret-token" -d '{"field": "val"}'
@post
async def access_header(request: Request, token: str = Header("X-auth-token")):
    if token != "secret-token":
        return AsgiResponse(status_code=401)
    data = await request.json()
    logger.info(data)  # {'field': 'val'}
    return AsgiResponse(status_code=200)


broker = NatsBroker()
app = AsgiFastStream(broker, asgi_routes=[
    ("/access_request", access_request),
    ("/access_query", access_query),
])
app.mount("/access_header", access_header)
```

## Type of change

## Checklist


faststream/asgi/app.py
```diff
@@ -123,7 +123,7 @@ def __init__(
 
         for path, app in self.routes:
             if isinstance(app, HttpHandler):
```
Lancetnik: double check

faststream/asgi/factories.py
```diff
@@ -1,22 +1,20 @@
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any, Union
 
+from faststream.asgi.handlers import get
```
Lancetnik: the same level imports should be related

faststream/asgi/handlers.py
```diff
@@ -33,12 +40,23 @@ async def __call__(self, scope: "Scope", receive: "Receive", send: "Send") -> No
...
+                        response = await self.func(scope)
+                else:
+                    response = await self.func(scope)
             except Exception:
+                logger.exception("Exception occurred while processing request")
                 response = AsgiResponse(body=b"Internal Server Error", status_code=500)
 
         await response(scope, receive, send)
 
+    def set_context(self, context_repo: ContextRepo) -> None:
```
Lancetnik: Please, take a look, how brokers consumes a config from an app. Here should be used the same mechanism

faststream/asgi/factories.py
Lancetnik, line 1: Please, revert this file changes

faststream/asgi/annotations.py
```diff
@@ -0,0 +1,9 @@
+from typing import Annotated
+
+from faststream._internal.context import Context
+
+from .request import AsgiRequest
+
+__all__ = ["Request"]
```
Lancetnik: Should be a tuple

faststream/asgi/handlers.py
```diff
@@ -19,26 +28,39 @@ def __init__(
...
 
         else:
             try:
-                response = await self.func(scope)
+                with self.fd_config.context.scope(
+                    "request", AsgiRequest(scope, receive, send)
+                ):
+                    response = await self.func(scope)
             except Exception:
+                logger.exception("Exception occurred while processing request")
```
Lancetnik: Should we use this logger or from the `context`? Also, I suggest to add an exception traceback by `logger.error(..., exc_info=exception)`

Lancetnik:
@Sehat1137 can you review, please?

Sehat1137:
I think we need support new features into AsyncAPI schema, because right now we don't see any information about required headers or query params

Sehat1137:
> I think we need support new features into AsyncAPI schema, because right now we don't see any information about required headers or query params

next pr

faststream/asgi/handlers.py
borisalekseev: Here I replaced the logger with the application logger and also added the logger forwarding to the context so it could be used in the handler.

faststream/asgi/app.py
```diff
@@ -155,7 +165,16 @@ def from_app(
...
+        asgi_route = (path, route)
+        self.routes.append(asgi_route)
+        self._register_route(asgi_route)
+
+    def _register_route(self, asgi_route: tuple[str, "ASGIApp"]) -> None:
+        path, route = asgi_route
+        if isinstance(route, HttpHandler):
+            self.schema.add_http_route(path, route)
+            route.update_fd_config(self.config)
+            route.set_logger(self.logger)
```
Sehat1137, line 175: I think we need use warnings, not just ignore

faststream/asgi/factories.py
```diff
@@ -134,7 +133,7 @@ def make_asyncapi_asgi(
         tags=tags,
         unique_id=unique_id,
     )
-    async def docs(scope: "Scope") -> AsgiResponse:
+    async def docs(scope: Scope) -> AsgiResponse:
         nonlocal cached_docs
```
Sehat1137, line 137: Let's change this and will use class property as the cache instead of nonlocal or context.

@Lancetnik Do you have any opinion about this?

faststream/asgi/app.py
Lancetnik: I don't think so. User could be able to register any ASGI endpoints like `prometheus.make_asgi_app()`. It is expected behavior, that `FastStream` can't generate AsyncAPI spec for such outer endpoints. So, warnings is an annoying thing here, not a useful note.

faststream/asgi/factories.py
Lancetnik: I am OK with current impl

faststream/asgi/app.py
```diff
@@ -103,12 +103,16 @@ def __init__(
         specification: Optional["SpecificationFactory"] = None,
         asyncapi_path: str | AsyncAPIRoute | None = None,
     ) -> None:
+        self.routes = list(asgi_routes)
+        context = context or ContextRepo()
+        context.set_global("app", self)
```
Lancetnik: It is already set in `super.__init__`, there is no need to double this logic here

borisalekseev: The problem is that I need access to context.app when registering the handler, which happens before the parent class installs the app in the context. In this case, can I change the behavior in `faststream._internal.application.Application`?

Lancetnik (approved):

PR #2470, opened 2025-08-25
Add configurations for subscriber and publisher

### LINKED ISSUE #2451: Feature:  broker-level topic configuration
Currently, users have to manually set up all incoming and outgoing topics in the `@router.subscriber()` and `@router.publisher()` decorators. It would be great if we could define topic configurations at the broker level.

If users want to specify topics from the settings, they have several options:

- Use a global settings object: `@router.subscriber(settings.TOPIC)`
- Import a function and set it up in the main:
    ```python
    from routes import handler
    
    def main():
        broker = Broker()  
        broker.subscriber(handler)  
  ```

- The closest option is using `Route`: https://faststream.ag2.ai/latest/getting-started/routers/#delay-handler-registration

All these alternatives require additional work from the user and are not as declarative as we'd like.
Therefore, I propose providing users with a more flexible configuration option. Something like this:

```python
@router.subscriber(Config("IN_TOPIC"))
async def handler(message: Message):
    ...

def main():
    broker = Broker(config={
        "IN_TOPIC": "test-topic",
    })
    broker.include_router(router)
```

This way, we can make any static routing definition truly lazy, and we don't need to rely on global state. This also allows for more flexible configuration options. The approach should be the same for both subscribers and publishers.

The configuration can be used as a value for any option, including nested ones. This is especially useful for complex configurations. For example

```python
@rabbit_broker.subscriber(
    queue=RabbitQueue(name="test", routing_key=Config("ROUTING_KEY"))
    exchange=Config("EXCHANGE_NAME")
)
async def handler(message: Message): ...
```

Flosckow:
# Description

Fixes #2451

## Type of change

## Checklist


Flosckow:
@Lancetnik, can you check it?
this simple example
```python
from faststream.rabbit import RabbitBroker, RabbitQueue
from faststream._internal.configs.settings import Settings

q = RabbitQueue("test")

settings = Settings(q1=q)
broker = RabbitBroker(settings=settings)

app = FastStream(broker)

@broker.subscriber(queue=Settings("q1"))
async def base_handler(body: str):
    print(body)


@app.on_startup
async def pub():
    await broker.connect()
    await broker.publish("hi", queue=Settings("q1"))

```

faststream/rabbit/publisher/usecase.py
```diff
@@ -70,6 +70,7 @@ def routing(
         queue: Union["RabbitQueue", str, None] = None,
         routing_key: str = "",
     ) -> str:
+        self.queue = self._outer_config.settings.resolve_from(self.queue)
```
Flosckow: here, if we call publisher.start() this is not necessary, since the queue is set correctly, however, if we do not call start, it turns out that this code is necessary, since routing is called earlier than start and an error occurs.

PR #2531, opened 2025-09-30
[ISS-2462] fix(redis): list subscriber consume/stop race

### LINKED ISSUE #2462: Bug: Redis ListSubscriber discards unprocessed messages on shutdown
**Describe the bug**
I've encountered an issue where the Redis [ListSubscriber](https://github.com/ag2ai/faststream/blob/0.5.48/faststream/redis/subscriber/usecase.py#L463-L510) discards messages without processing them during shutdown.

It appears this happens because the [_consume](https://github.com/ag2ai/faststream/blob/0.5.48/faststream/redis/subscriber/usecase.py#L186-L205) method is outside the lock that is checked by [stop](https://github.com/ag2ai/faststream/blob/0.5.48/faststream/broker/subscriber/usecase.py#L201-L209). Consequently, the [_get_msgs](https://github.com/ag2ai/faststream/blob/0.5.48/faststream/redis/subscriber/usecase.py#L191) can be [canceled](https://github.com/ag2ai/faststream/blob/0.5.48/faststream/redis/subscriber/usecase.py#L215) before it finishes fetching messages.

Additionally, another potential reason for messages being discarded is the check within the [consume](https://github.com/ag2ai/faststream/blob/0.5.48/faststream/broker/subscriber/usecase.py#L302-L303) method , which silently exits if self.running is False.

I have not checked whether other subscribers are affected by a similar issue.

**How to reproduce**
The following code reproduces the issue by patching Redis's lpop to simulate a delay in message retrieval.

```python
import asyncio
from unittest.mock import patch

from faststream.redis import RedisBroker
from faststream.redis.schemas import ListSub
from pydantic import BaseModel
from redis.asyncio import Redis


class Data(BaseModel):
    data: str


broker = RedisBroker(graceful_timeout=60)

input_subscriber = broker.subscriber(list=ListSub(list_name="input-list"))
input_publisher = broker.publisher(list=ListSub(list_name="input-list"))
output_publisher = broker.publisher(list=ListSub(list_name="output-list"))


@input_subscriber
@output_publisher
async def handle(msg: Data, logger: Logger) -> Data:
    logger.info(f"Received message: {msg.data}")
    return Data(data="bar")


app = FastStream(broker)


async def lpop(
    name: str,
    count: int | None = None,
):
    if count is not None:
        result = await broker._connection.execute_command("LPOP", name, count)
    else:
        result = await broker._connection.execute_command("LPOP", name)

    if result is None:
        return None

    await asyncio.sleep(20)

    return result

patcher = patch.object(Redis, "lpop", side_effect=lpop)
patcher.start()


@app.after_startup
async def after_startup():
    await input_publisher.publish(Data(data="foo"))


async def main():
    task = asyncio.create_task(app.run())
    await asyncio.sleep(10)
    app.exit()
    await task


if __name__ == "__main__":
    asyncio.run(main())
...
```

**Expected behavior**
Messages fetched from the Redis list should be processed unless a timeout occurs. If a message is discarded without being processed, an error log should be generated.

**Observed behavior**
The data from the Redis list is discarded without being processed and without any logs.

**Environment**
```python
Running FastStream 0.5.48 with CPython 3.12.11 on Linux
```

NelsonNotes:
# Description
Fixes redis ListSubscriber non-syncronized gap between message reading and handling which may lead to race condition in case of subscriber stopping.

Fixes #2462

## Type of change

Bug fix (a non-breaking change that resolves an issue)

## Checklist

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -42,7 +42,7 @@ services:
       - no-new-privileges:true
 
   faststream:
-    image: faststream:local
+    build: .
     volumes:
       - ./:/src
       - /src/.venv/
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -42,7 +42,7 @@ services:
       - no-new-privileges:true
 
   faststream:
-    build: .
+    image: faststream:local
     volumes:
       - ./:/src
       - /src/.venv/
--- a/justfile
+++ b/justfile
@@ -58,41 +58,41 @@ test-coverage-all +param="tests/":
 [doc("Build docs")]
 [group("docs")]
 docs-build:
-  cd docs && uv run --frozen python docs.py build
+  cd docs && VIRTUAL_ENV=.venv uv run --frozen python docs.py build
 
 [doc("Build API Reference")]
 [group("docs")]
 docs-build-api:
-  cd docs && uv run --frozen python docs.py build-api-docs
+  cd docs && VIRTUAL_ENV=.venv uv run --frozen python docs.py build-api-docs
 
 [doc("Update release notes")]
 [group("docs")]
 docs-update-release-notes:
-  cd docs && uv run --frozen python docs.py update-release-notes
+  cd docs && VIRTUAL_ENV=.venv uv run --frozen python docs.py update-release-notes
 
 [doc("Serve docs")]
 [group("docs")]
 docs-serve params="":
-  cd docs && uv run --frozen python docs.py live 8000 {{params}}
+  cd docs && VIRTUAL_ENV=.venv uv run --frozen python docs.py live 8000 {{params}}
 
 # Linter
 [doc("Ruff format")]
 [group("linter")]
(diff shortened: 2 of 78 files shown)


docker-compose.yaml
```diff
@@ -42,7 +42,7 @@ services:
       - no-new-privileges:true
 
   faststream:
-    build: .
+    image: faststream:local
```
Lancetnik: Please, revert

justfile
```diff
@@ -14,8 +14,8 @@ default:
 [doc("Init infra")]
 [group("infra")]
 init python="3.10":
-  docker build . --build-arg PYTHON_VERSION={{python}}
-  uv sync --group dev
+  docker build . --build-arg PYTHON_VERSION={{python}} --no-cache --tag faststream:local
```
Lancetnik: Is the change required?

PR #1256, opened 2024-02-23
feat (#1252): respect Redis StreamSub last_id with consumer group

Lancetnik:
# Description

Fixes #1254 

## Type of change

Please delete options that are not relevant.

- [ ] Documentation (typos, code examples, or any documentation updates)
- [ ] Bug fix (a non-breaking change that resolves an issue)
- [x] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a fix or feature that would disrupt existing functionality)
- [ ] This change requires a documentation update

## Checklist

- [x] My code adheres to the style guidelines of this project (`scripts/lint.sh` shows no errors)
- [x] I have conducted a self-review of my own code
- [ ] I have made the necessary changes to the documentation
- [x] My changes do not generate any new warnings
- [x] I have added tests to validate the effectiveness of my fix or the functionality of my new feature
- [x] Both new and existing unit tests pass successfully on my local environment by running `scripts/test-cov.sh`
- [x] I have ensured that static analysis tests are passing by running `scripts/static-anaylysis.sh`
- [ ] I have included code examples to illustrate the modifications

Issue #1254, opened 2024-02-23
Feature: support `last_id` for Redis stream consumer group


chrisgoddard:
Currently, you can set up a subscriber for a redis stream like this:

```python
@broker.subscriber(stream=StreamSub('stream-name', group='faststream', consumer='1'))
async def handle_msg(msg) -> None:
    logger.info(f"Received: {msg}")
```

However, if you try to past a `last_id` parameter:
```python
@broker.subscriber(stream=StreamSub('stream-name', group='faststream', consumer='1', last_id='0-0'))
async def handle_msg(msg) -> None:
    logger.info(f"Received: {msg}")
```

You get a warning:
`RuntimeWarning: `last_id` has no effect with consumer group`

It seems that in the redis/handler.py, it's hard-coding the '>' special id rather than passing through the last_id parameter (which it does when not using a consumer group)

```python
# redis/handler.py,
if stream.group and stream.consumer:
    read = client.xreadgroup(
        groupname=stream.group,
        consumername=stream.consumer,
        streams={stream.name: ">"},
        block=stream.polling_interval,
        noack=stream.no_ack,
    )

else:
    read = client.xread(
        {stream.name: self.last_id},
        block=stream.polling_interval,
    )
``

Is there a reason why we couldn't just pass self.last_id through to the dictionary and have last_id have the default value of ">"?

That way if you want to pull from a different offset while using a consumer group, you can.

Lancetnik:
@chrisgoddard seems like we broke Redis Streams behavior:

Consumer group start position can be setted only by (`xgroup_create` command)[https://redis.io/docs/data-types/streams/#consumer-groups]

But, last_id should be always `>`.

>If the ID is the special ID > then the command will return only new messages never delivered to other consumers so far, and as a side effect, will update the consumer group's last ID.

> If the ID is any other valid numerical ID, then the command will let us access our history of pending messages. That is, the set of messages that were delivered to this specified consumer (identified by the provided name), and never acknowledged so far with [XACK](https://redis.io/commands/xack).

Specific ID is suitable only for one-time requests, but not regular consuming. I should rollback this behavior

PR #2564, opened 2025-10-09
BUG: ensure that stream read from the subscriber.last_id not the stream_sub

JonathanSerafini:
# Description

When reading messages from the redis stream, the consumer always attempts to read from the `last_id` configured on the `stream_sub` instead of the `last_id` being tracked as an instance parameter. By switching over to the `last_id` that is saved on the subscriber, it is now possible to implement custom retry logic in middlewares or as background tasks via `add_task`.

I've labeled this as a bug since it appears to be the _only_ such occurence in this file.

## Type of change

## Checklist

### DIFF
--- a/faststream/redis/subscriber/usecases/stream_subscriber.py
+++ b/faststream/redis/subscriber/usecases/stream_subscriber.py
@@ -133,7 +133,7 @@ class _StreamHandlerMixin(LogicSubscriber):
                 return client.xreadgroup(
                     groupname=stream.group,
                     consumername=stream.consumer,
-                    streams={stream.name: stream.last_id},
+                    streams={stream.name: self.last_id},
                     count=stream.max_records,
                     block=stream.polling_interval,
                     noack=stream.no_ack,


Lancetnik:
@JonathanSerafini it doesn't look like a bug, let me check

JonathanSerafini:
hmm i'm curious, there wasn't a test case for this exception so i'm not sure of the use case.

it's effectively the same behaviour as before ( i.e. each call to xreadgroup continues to use the updated last_id )
the only functional change i can think of might be that if you `sub.start` `sub.stop` `sub.start` it would continue from the updated last_id rather than `stream_sub.last_id` ( i.e. from the last message id rather than something like `>` )

currently i'm trying to setup a background task per subscriber via `add_task` so that i can get the consumer to pull fro the PEL rather than from `>`. I know it's a bit hacky given how it interacts with `xreadgroup` polling but all i've got for now.

Lancetnik:
Well, it was made to read always with ">"

https://redis.io/docs/latest/commands/xreadgroup/#differences-between-xread-and-xreadgroup

> The ID to specify in the STREAMS option when using XREADGROUP can be one of the following two:

> The special > ID, which means that the consumer want to receive only messages that were never delivered to any other consumer. It just means, give me new messages.

> Any other ID, that is, 0 or any other valid ID or incomplete ID (just the millisecond time part), will have the effect of returning entries that are pending for the consumer sending the command with IDs greater than the one provided. So basically if the ID is not >, then the command will just let the client access its pending entries: messages delivered to it, but not yet acknowledged. Note that in this case, both BLOCK and NOACK are ignored.

So, the correct function should looks like

```python
def read(last_id: str):
    return client.xreadgroup(
        groupname=stream.group,
        consumername=stream.consumer,
        streams={stream.name: stream.last_id if stream.last_id == ">" else last_id},
        count=stream.max_records,
        noack=stream.no_ack,
    )
```

But, I don't like such check at every message. Can we make any workaround?

Lancetnik:
> it's effectively the same behaviour as before ( i.e. each call to xreadgroup continues to use the updated last_id )

It is not correct. `stream.last_id` is never changing. It is always `>` value for xreadgroup (for most cases)

@JonathanSerafini please, check original PR https://github.com/ag2ai/faststream/pull/1256 and issue https://github.com/ag2ai/faststream/issues/1254

Lancetnik:
Btw, @JonathanSerafini I’ve noticed you’ve been using FastStream for quite a while and have contributed a lot of valuable feedback and improvements — thank you for that! 🙌

We’d love to hear more about your experience: what works well for you and what could be improved. Your insights would really help us shape the future of FastStream.

Thanks again for your continued support!

JonathanSerafini:
hmm, my apologies there ... i must have made an incorrect assumption somewhere ...
upon closer look you are absolutely correct. :|

i'll close this ... and look to another approach

PR #2635, opened 2025-11-01
Feat/asyncapi add reply

### LINKED ISSUE #1586: feature: document RPC responses in AsyncAPI
### Discussed in https://github.com/airtai/faststream/discussions/1570

<div type='discussions-op-text'>

<sup>Originally posted by **gri38** July  1, 2024</sup>
Hello.

When we subscribe to a queue to answer a RPC call, it seems we cannot document the return type.

H§ere a small exemple:
```python
import logging

from faststream.rabbit import RabbitBroker
from pydantic import BaseModel, conint

broker = RabbitBroker()
app = FastStream(broker)


class In(BaseModel):
    i: str
    j: conint(gt=0)

class Out(BaseModel):
    k: str
    l: float


@broker.subscriber("RPC",
                   title="RPC",
                   description="This queue is used to for RPC",

                   )
async def rpc_handler(input: In) -> Out:
    logging.info(f"{input=}")
    return Out(k=input.i, l=input.j)
```

If I run with `faststream docs serve rpc:app` I cannot see Out in the doc (http://localhost:8000).

Is there a way to do so ?

When I use the RabbitRouter, I can see in the router @subscriber decorator: 
```python
# FastAPI args
        response_model: Annotated[
            Any,
            Doc(
                """
                The type to use for the response.

                It could be any valid Pydantic *field* type. So, it doesn't have to
                be a Pydantic model, it could be other things, like a `list`, `dict`,
                etc.

                It will be used for:

                * Documentation: the generated OpenAPI (and the UI at `/docs`) will
                    show it as the response (JSON Schema).
                * Serialization: you could return an arbitrary object and the
                    `response_model` would be used to serialize that object into the
                    corresponding JSON.
                * Filtering: the JSON sent to the client will only contain the data
                    (fields) defined in the `response_model`. If you returned an object
                    that contains an attribute `password` but the `response_model` does
                    not include that field, the JSON sent to the client would not have
                    that `password`.
                * Validation: whatever you return will be serialized with the
                    `response_model`, converting any data as necessary to generate the
                    corresponding JSON. But if the data in the object returned is not
                    valid, that would mean a violation of the contract with the client,
                    so it's an error from the API developer. So, FastAPI will raise an
                    error and return a 500 error code (Internal Server Error).

                Read more about it in the
                [FastAPI docs for Response Model](https://fastapi.tiangolo.com/tutorial/response-model/).
                """
            ),
```

But even with response_model: nothing in asyncAPI.

Thanks for any tips, even for "it's not possible" ;-)

</div>

stefan-vasilenko:
# Description

Added reply generation to AsyncApi scheme v3.0.0

Fixes #1586

## Type of change

## Checklist


stefan-vasilenko:
@Lancetnik Hi!
I’ve added a reply to the schema generation — could you please check the implementation?

I also have a question: after my changes, a lot of tests started failing because they assert the exact payload output, like this:
```python
        assert len(payload) == 1

        key, value = next(iter(payload.items()))

        assert key == "User"
        assert value == {
            "properties": IsDict({
                "id": {"title": "Id", "type": "integer"},
                "email": {"default": "", "title": "Email", "type": "string"},
            })
            | IsDict({
                "id": {"title": "Id", "type": "integer"},
                "name": {"default": "", "title": "Name", "type": "string"},
            }),
            "required": ["id"],
            "title": key,
            "type": "object",
        }
```
The failures are caused by the fact that I added a new payload to the output.

Do you think it makes sense to update all these tests, or is there a better way to handle this?

Lancetnik:
@stefan-vasilenko thank you! At first look everything looks fine. Let me check in details, what we can make with failing tests

stefan-vasilenko:
Sorry, for long inactive I was back to this feature and will complete it

Lancetnik:
@stefan-vasilenko sorry, it's my bad! Seems like the last thing we need to do - review. Can you just fix CI to let me investigate the feature?

stefan-vasilenko:
Of course, wait my commits!

stefan-vasilenko:
Hi! Can u please run tests, i thought i fix them

PR #2656, opened 2025-11-06
fix: Recreate group in case redis flush

### LINKED ISSUE #2611: Bug: When the redis server is killed, faststream can't recover
**Describe the bug**
Today, my redis server was killed.
Almost immediately, it was back alive and the data were flowing properly from the sender.

But faststream stayed stuck, complaining the group was not existing.

**How to reproduce**

- create a subscriber with a stream and consumer group on redis
- launch the application
- kill redis + empty it (if not done by the kill)
- faststream is sending the following message:

```python

2025-10-23T14:38:01.670669760Z 2025-10-23 16:38:01,670 ERROR    -                     |            - Message fetch error
2025-10-23T14:38:01.670708509Z Traceback (most recent call last):
2025-10-23T14:38:01.670717162Z   File "/venv/lib/python3.12/site-packages/faststream/redis/subscriber/usecases/basic.py", line 94, in _consume
2025-10-23T14:38:01.670722550Z     await self._get_msgs(*args)
2025-10-23T14:38:01.670727037Z   File "/venv/lib/python3.12/site-packages/faststream/redis/subscriber/usecases/stream_subscriber.py", line 289, in _get_msgs
2025-10-23T14:38:01.670732145Z     for stream_name, msgs in await read(self.last_id):
2025-10-23T14:38:01.670737974Z                              ^^^^^^^^^^^^^^^^^^^^^^^^
2025-10-23T14:38:01.670747609Z   File "/venv/lib/python3.12/site-packages/redis/asyncio/client.py", line 616, in execute_command
2025-10-23T14:38:01.670758445Z     return await conn.retry.call_with_retry(
2025-10-23T14:38:01.670787409Z            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-10-23T14:38:01.670793789Z   File "/venv/lib/python3.12/site-packages/redis/asyncio/retry.py", line 59, in call_with_retry
2025-10-23T14:38:01.670798466Z     return await do()
2025-10-23T14:38:01.670804846Z            ^^^^^^^^^^
2025-10-23T14:38:01.670808882Z   File "/venv/lib/python3.12/site-packages/redis/asyncio/client.py", line 590, in _send_command_parse_response
2025-10-23T14:38:01.670813128Z     return await self.parse_response(conn, command_name, **options)
2025-10-23T14:38:01.670816964Z            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-10-23T14:38:01.670820890Z   File "/venv/lib/python3.12/site-packages/redis/asyncio/client.py", line 637, in parse_response
2025-10-23T14:38:01.670825317Z     response = await connection.read_response()
2025-10-23T14:38:01.670829794Z                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-10-23T14:38:01.670834331Z   File "/venv/lib/python3.12/site-packages/redis/asyncio/connection.py", line 577, in read_response
2025-10-23T14:38:01.670838778Z     raise response from None
2025-10-23T14:38:01.670843765Z redis.exceptions.ResponseError: NOGROUP No such key 'global_input-stream' or consumer group 'call_activity_managers' in XREADGROUP with GROUP option
```

**Expected behavior**
I expect Faststream to detect the missing group and recreate it to recover the connection, in order to keep working as usual.
Throwing an error would be appreciated as well.

**Observed behavior**
FastStream application is dead

**Environment**
Faststream 0.6.1 on alpine docker image

**Additional context**
Redis being an in-memory DB, I think the reader need to make sure it can recover when redis is back.

pbonneaudiabolocom:
# Description

When using redis with group, if a flush is done by another parties or the stream is destroyed, faststream enter in an infinite loop, trying to get the next element but not finding the expected group.

Fixes #2611

## Type of change

## Checklist


pbonneaudiabolocom:
I added all the casting to pass the linter.
It might be another way as just the next read is returning a coroutine without raising any issue.. just tell me what you'd prefer me to do.

Also, I discivered that this part is not tested at all. (if I insert an exit in that part of the code the tests are running fine.

If you have some test example, I would imagine something such as: 

create stream with group
remove that group in redis(external command)
check that the group is recreated  + it works fine.

faststream/redis/subscriber/usecases/stream_subscriber.py
```diff
@@ -129,14 +132,42 @@ def read(
...
+                            groupname=stream.group,
+                            consumername=stream.consumer,
+                            streams={stream.name: stream.last_id},
+                            count=stream.max_records,
+                            block=stream.polling_interval,
+                            noack=stream.no_ack,
+                        )
+                    except ResponseError as e:
+                        if "NOGROUP" in str(e):
+                            await self._create_group()
```
Lancetnik: Should we change current `last_id` read position at stream recreation?

pbonneaudiabolocom: indeed we should. The group being reseted, next id should retart from scratch.

I added some tests as well to be sure it was working better

pbonneaudiabolocom:
My code is entering into conflict with the one from https://github.com/ag2ai/faststream/commit/9998c581648d0132456ea47a4398c6e382b276f0

I have no idea what the fix on autoclaim is, so it's very hard to understand how to mix it with my code and see what should happen.

May you help please ? @JonathanSerafini

faststream/_internal/cli/main.py
JonathanSerafini, line 2: not sure if this was intended, but your redis PR also includes this change and an unrelated package version change.

faststream/redis/subscriber/usecases/stream_subscriber.py
```diff
@@ -96,38 +96,145 @@ async def start(self) -> None:
...
+                                Offset,
+                                dict[bytes, bytes],
+                            ],
+                            ...,
+                        ],
+                    ],
+                    ...,
+                ],
+            ]:
+                async def read_from_group() -> tuple[
```
JonathanSerafini: i'm thinking at this point you may as well just make `def read` an async function and do without the nested `read_from_group` it may make things more readable ?

JonathanSerafini: this type annotation is likely the first conflict, you can use `ReadResponse` as a replacement

JonathanSerafini: the other conflict probably stems from the fact that there are now *2* stream group readers in this code now, one for `xreadgroup` and another for `xautoclaim` ... i would assuming that if you want to re-create the group for the former you may also want for the latter.

for the xautoclaim handling, i would just start over from `0-0` i.e the first position and it should be able to take it from there.

pbonneaudiabolocom:
Thanks for you feedback.

Thanks to it, I decided to crete a specific function that has to protect any provided method against group removal.

I updated the code accordingly.

Tests are passing.

I still have the collowing linter error, but I don't know how to make them disapeer. I'm totally lost in the return typing used to be honnest.
```
faststream/redis/subscriber/usecases/stream_subscriber.py:199: error: Incompatible types in assignment (expression has type "tuple[bytes, tuple[tuple[bytes, dict[bytes, bytes]], ...]]", variable has type "str")  [assignment]
faststream/redis/subscriber/usecases/stream_subscriber.py:202: error: Incompatible types in assignment (expression has type "tuple[bytes, tuple[tuple[bytes, dict[bytes, bytes]], ...]]", variable has type "bytes")  [assignment]
faststream/redis/subscriber/usecases/stream_subscriber.py:204: error: Non-overlapping equality check (left operand type: "tuple[bytes, tuple[tuple[bytes, dict[bytes, bytes]], ...]]", right operand type: "Literal[b'0-0']")  [comparison-overlap]
faststream/redis/subscriber/usecases/stream_subscriber.py:208: error: Incompatible return value type (got "tuple[tuple[bytes, tuple[bytes, tuple[tuple[bytes, dict[bytes, bytes]], ...]]]]", expected "tuple[tuple[bytes, tuple[tuple[bytes, dict[bytes, bytes]], ...]], ...]")  [return-value]
faststream/redis/subscriber/usecases/stream_subscriber.py:282: error: Incompatible types in assignment (expression has type "tuple[bytes, tuple[tuple[bytes, dict[bytes, bytes]], ...]]", variable has type "bytes")  [assignment]
faststream/redis/subscriber/usecases/stream_subscriber.py:286: error: Too many values to unpack (1 expected, 2 provided)  [misc]
faststream/redis/subscriber/usecases/stream_subscriber.py:347: error: Incompatible types in assignment (expression has type "tuple[bytes, tuple[tuple[bytes, dict[bytes, bytes]], ...]]", variable has type "bytes")  [assignment]
faststream/redis/subscriber/usecases/stream_subscriber.py:351: error: Too many values to unpack (1 expected, 2 provided)  [misc]
```

Any idea on how to correct that ?

pbonneaudiabolocom:
@JonathanSerafini Would you think it' possible to merge it ?

Last time it was working fine, but the more the system progress, the more I have to redo the job...

JonathanSerafini:
@pbonneaudiabolocom I'm not the right one to ask, i'm not a maintainer or affiliated with this project.

That being said re:typing, you have to account for the changes you're making. You've created `_xautoclaim_call` and stated it returns `Any`. Then wrap it in a decorator `_protect_read_from_group_removal` that states it returns `ReadResponse`. This is not actually true, xautoclaim's output is different. 

To account for different input functions you might want to look at how to type a 'parametrized decorator'.

pbonneaudiabolocom:
Hi,

I updated my code to fit the different updates.

I'm struggling with typing with mypy validation.

Honnestly, I have no idea on how to resolve the warning given.

I'm far from being at that level of expertise in typing. help would be welcome.

powersemmi:
@pbonneaudiabolocom  
Hi!

I believe the current implementation deviates from the framework's standards. While FastStream supports creating queues on startup (primarily for development convenience), it shouldn't be responsible for managing production infrastructure at runtime. Recreating the stream on the fly effectively hides the underlying infrastructure issue. 

The only correct approach here is to fail fast and raise an error indicating that the Redis stream no longer exists. This ensures the service stops predictably, triggering monitoring alerts so we can investigate the incident properly.

Also, it would be great if you could revert uv.lock

pbonneaudiabolocom:
Hi,

For me, whenever a system creates something, it is accountable for it.

Then there is 2 solutions: 
 - either to repair when something happen (that's the goal of my PR)
 - fail quickly and provides a way to trigger an action on user level.
 
 I'm not against the 2nd approach, but the question then would be, how can we ensure in case a stream is dead that the application will be able to recover somehow.
 In the simple case when there is one stream, application will fail, start again, init the group back and it will be ok.
 But in case there is multiple publishers / subscribers, failing the full app might have great consequences on other flow ongoing.
 
 It would be appreciable to raise an exception and to allow the user to capture it somehow and to call the group creation back if it's what is desired.
 That comes with 2 questions on my side: 
  - can we capture / handle exception raised by Faststream somewhere ?
  - can we call the group creation / subscriber restart ?
  
  What I would like to avoid is a group creation done in a separate function, ending up with 2 ways to create group, one in faststream and the othr one in the backup system... more likely to create intrical issues on production.
  
  I could adapt the PR to fit that other approach if you want.

  Best regards,
  Pierre

pbonneaudiabolocom:
Hello,

We are still wondering how that could be taken care of.

Today, in case of redis flush, any faststream application using stream + group will crash until faststream is restarted.

Best regards,
Pierre

Lancetnik:
Hi @pbonneaudiabolocom

Thanks for the persistence on this and for the detailed reasoning — both the problem and the trade-offs you've laid out are real, and I appreciate you engaging with the discussion rather than just pushing the patch through.

I want to expand on @powersemmi's point, because I don't think we framed the "why" clearly enough.
FastStream's stance on infrastructure management is intentional and consistent across brokers: the application is a consumer of infrastructure, not its owner. The startup-time queue/group/stream creation we do today is explicitly a developer-experience convenience for local environments, and we recommend disabling it in production. The analogy we usually reach for is databases — an application doesn't create or repair its own tables at runtime; that responsibility belongs to migrations, IaC, or the platform team. Streams and consumer groups sit in the same category. Silently recreating a group that disappeared means the application is now papering over an incident (a FLUSHALL, a misconfigured retention, an operator mistake) that someone almost certainly needs to know about.
So the correct behavior on NOGROUP really is to surface the error loudly, not to self-heal. "Crash and get restarted by the orchestrator with alerts firing" is a feature in this model, not a regression.
That said, your follow-up questions are the right ones, and I don't want to leave you without a path forward:

> Can we capture/handle exceptions raised by FastStream somewhere?
> Can we call the group creation / subscriber restart?

Both yes, and I think this is where the PR could actually land in a way that fits the framework:

Make the failure observable and typed. Today the loop swallows NOGROUP and spins. The minimal correct fix is to let that ResponseError (or a wrapped, more specific exception like StreamGroupNotFoundError) propagate out of the subscriber instead of being retried indefinitely. That alone solves the "infinite loop until restart" symptom from #2611 — the process fails fast, the orchestrator restarts it, the startup-time group creation re-runs, and you're back online. No multi-publisher app dies silently.

Give users a hook, not automatic recovery. If a user explicitly wants the self-healing behavior for their environment, exposing it as opt-in (an error handler, a lifespan callback, or a documented pattern using FastStream's existing exception middleware) is much more in line with the framework's philosophy than baking it into the default read path. The user owns the policy; FastStream just gives them the seam.

If you'd be open to reshaping the PR along those lines — propagating a clear, catchable exception instead of recreating the group inline — I think there's a much shorter path to merge here, and it would actually close #2611 more robustly than the current approach (which only handles the Redis-flush case, not e.g. the stream being deleted while groups exist elsewhere).

Happy to discuss the exception type and where exactly it should be raised if that direction works for you.

powersemmi:
I fix this here https://github.com/ag2ai/faststream/pull/2855

pbonneaudiabolocom:
thank you both for your time.

Our team suggest to add a warning in case someone trust Faststream to create the group, reminding him that's it's not production ready / advised)

I'll follow up in the other PR :)

Issue #799, opened 2023-10-04
Feature: PostgreSQL as a Queue


Vad1mo:
**Is your feature request related to a problem? Please describe.**

I am aware that PostgreSQL is now a Message Queue per se; however, it can be utilized as such. Many developers realize that it suffices for the majority of tasks, and it comes with zero added costs if a PostgreSQL db is present.

**Describe the solution you'd like**
Clearly and concisely describe the desired outcome or solution.

**Feature code example**
Nothing in particular 

**Describe alternatives you've considered**
You can do exactly this with MySQL and SQL server too because they both support SKIP LOCKED.

**Additional context**

- https://adriano.fyi/posts/2023-09-24-choose-postgres-queue-technology/
- https://www.prequel.co/blog/sql-maxis-why-we-ditched-rabbitmq-and-replaced-it-with-a-postgres-queue
- https://dagster.io/blog/skip-kafka-use-postgres-message-queue

acaloiaro:
@Vad1mo :wave: I came across this by chance and see you've referenced my write up on the topic of PG as a queue. 

If you find Go easy enough to read, you might benefit from neoq (https://github.com/acaloiaro/neoq) as a reference implementation for the ideas lain out in my post.

Cheers!

Vad1mo:
> @Vad1mo :wave: I came across this by chance and see you've referenced my write up on the topic of PG as a queue. 
> 
> 
> 
> If you find Go easy enough to read, you might benefit from reading the neoq source (https://github.com/acaloiaro/neoq) as a reference implementation for the ideas lain out in my post. The whole PG implementation is contained within: https://github.com/acaloiaro/neoq/blob/main/backends/postgres/postgres_backend.go
> 
> 
> 
> Cheers!

Found it already 👀 https://github.com/acaloiaro/neoq/issues/70

Lancetnik:
I don't like an idea to poll DB for updates
I am thinking about Listen/Notify PG feature or even CDC (replication journal) consuming

Vad1mo:
Yes, I was referring to listen/notify. No mention about polling. 
Regarding CDC, here I am not sure how or if it is low level, hence if it can be easily setup within the offerings of cloud providers.

Lancetnik:
CDC available from any pg driver: asyncpg/psycopg/etc
It is just a part of protocol
So, it can be a useful feature as I can image

mixartemev:
[...]

PR #2704, opened 2025-12-02
feat: add sqla broker

Arseniy-Popov:
# Description

Work-in-progress for a relational database based broker. Preliminary state of the docs is available at [arseniy-popov.github.io/faststream/sqla/](https://arseniy-popov.github.io/faststream/sqla/).  

[#799](https://github.com/ag2ai/faststream/issues/799)

## Motivation

The primary benefit of a message queue built on top of a relational database is the ability to insert messages transactionally, atomically with other database operations, thus enabling the transactional outbox pattern. Also, the relational database is usually the most readiliy available, already provisioned piece of infrastructure for a given service. While implementing all patterns / semantics of full-blown message queue / streaming platform (e.g. kafka-like partitions-enabled horizontal scaling with local ordering, etc.) would be problematic, given a proper understanding of the trade-offs involved, the relational database based queue would be an appropriate tool for many low-to-medium throughput, latency-tolerant uses, including as parts of a larger messaging flow that involves a "proper" queue (e.g. as an outbox between a service and a queue). 

## Design

The key components are in [message.py](https://github.com/Arseniy-Popov/faststream/blob/feat/sqla-broker/faststream/sqla/message.py), [usecase.py](https://github.com/Arseniy-Popov/faststream/blob/feat/sqla-broker/faststream/sqla/subscriber/usecase.py), [client.py](https://github.com/Arseniy-Popov/faststream/blob/feat/sqla-broker/faststream/sqla/client.py).

#### Flow:  

On start, the subscriber spawns four types of concurrent loops:

1. Fetch loop: Periodically fetches PENDING or RETRYABLE messages from the database, simultaneously updating them in the database: marking as PROCESSING, setting acquired_at to now, and incrementing deliveries_count. Only messages with next_attempt_at <= now are fetched, ordered by next_attempt_at. The fetched messages are placed into an internal queue. The fetch limit is the minimum of fetch_batch_size and the free buffer capacity (fetch_batch_size * overfetch_factor minus currently queued messages). If the last fetch was "full" (returned as many messages as the limit), the next fetch happens after min_fetch_interval; otherwise after max_fetch_interval.

2. Worker loops (max_workers instances): Each worker takes a message from the internal queue and first checks if max_deliveries has been exceeded; if so, the message is Reject'ed without processing. Otherwise, processing proceeds. Depending on the processing result, AckPolicy, and manual Ack/Nack/Reject, the message is Ack'ed, Nack'ed, or Reject'ed. For Nack'ed messages, the retry_strategy is consulted to determine if and when the message might be retried. If allowed to be retried, the message is marked as RETRYABLE, otherwise as FAILED. Ack'ed messages are marked as COMPLETED and Reject'ed messages are marked as FAILED. The message is then buffered for flushing.

3. Flush loop: Periodically flushes the buffered message state changes to the database. COMPLETED and FAILED messages are moved from the primary table to the archive table. The state of RETRYABLE messages is updated in the primary table.

4. Release stuck loop: Periodically releases messages that have been stuck in PROCESSING state for longer than release_stuck_timeout since acquired_at. These messages are marked back as PENDING.

On stop, all loops are gracefully stopped. Messages that have been acquired but are not yet being processed are drained from the internal queue and marked back as PENDING. The subscriber waits for all tasks to complete within graceful_shutdown_timeout, then performs a final flush.

#### Notes:

This design allows for work sharing between processes/nodes because "select for update skip locked" is utilized.

This design adheres to the "at least once" processing guarantee because flushing changes to the database happens only after a processing attempt. A flush might not happen due to e.g. a crash. This might lead to messages being processed more times than allowed by the retry_strategy, and to the database state being inconsistent with the true number of attempts.

This design opts for separate short-lived transactions instead of a single one. This first one fetches messages with SELECT FOR UPDATE SKIP LOCKED and sets their state to PROCESSING, which prevents them from being fetched by other processes/nodes. The other two transactions flush message state updates to the database.

Setting max_deliveries to a non-None value provides protection from the poison message problem (messages that crash the worker without the ability to catch the exception due to e.g. OOM terminations) because deliveries_count is incremented and max_deliveries is checked prior to a processing attempt. However, this comes at the expense of violating the "at most once" processing guarantee, especially for messages that are being processed at the same time with the poison message (a crash would leave them with incremented deliveries_count despite possibly not having been processed).

Why not use LISTEN/NOTIFY? It is specific to Postgres, while it is preferable to start with functionality universal to any database. When using multiple nodes/processes, distributing messages among them would still require "SELECT FOR UPDATE SKIP LOCKED", because the notification will be delivered to all nodes/processes. A notification may also fail to arrive, especially if a node restarts. That is, polling is needed in any case. And once polling is in place, listen/notify can be integrated to “wake up” the polling loop earlier than as per the interval-based schedule.

## Type of change

## Checklist


Lancetnik:
CLAUDE REVIEW

<h1 style="margin-top: 0px; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Code Review: PR #2704 —<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">feat: add sqla broker</code></h1><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong>Author:</strong> Arseniy-Popov | <strong>+6,033 / -89</strong> across 68 files</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">This PR adds a SQLAlchemy-based message broker that implements a transactional outbox pattern using relational databases as message queues. The design is well thought-out (polling with <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">SELECT FOR UPDATE SKIP LOCKED</code>, batched flushing, worker pools), and the PR description is excellent. However, there are several issues that should be addressed before merging.</p><hr style="font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><h2 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Critical Bugs</h2><h3 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">1.<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">headers=None</code><span> </span>crashes the parser</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/parser.py#L23" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80); text-decoration: rgb(230, 180, 80);">parser.py:23</a> — <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">message.headers.get("content-type")</code> will raise <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">AttributeError</code> when <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">headers</code> is <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">None</code>, which is allowed since the column is <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">nullable=True</code> in <a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/client.py#L50" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80); text-decoration: rgb(230, 180, 80);">client.py:50</a> and <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">enqueue()</code> accepts <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">headers=None</code>.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong>Fix:</strong> Use <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">(message.headers or {}).get(...)</code>.</p><h3 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">2. Validation function is defined but never called</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/subscriber/factory.py#L56-L104" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80); text-decoration: rgb(230, 180, 80);">subscriber/factory.py:56-104</a> — <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_validate_input_for_misconfiguration()</code> contains important guardrails (checking <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">max_deliveries &gt; 0</code>, warning about incompatible ack_policy/retry_strategy combos) but is dead code — never called from <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">create_subscriber</code>.</p><h3 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">3.<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">NoRetryStrategy.get_next_attempt_at</code><span> </span>raises<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">ValueError</code><span> </span>unexpectedly</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/retry.py#L121-L133" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80); text-decoration: rgb(230, 180, 80);">retry.py:121-133</a> — When <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">max_attempts &gt; 1</code>, calling <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">get_next_attempt_at</code> for <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">attempts_count &lt; max_attempts</code> raises <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">ValueError</code> instead of returning a datetime. The <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_nack</code> flow in <a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/message.py#L89-L98" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80); text-decoration: rgb(230, 180, 80);">message.py:89-98</a> will crash the worker.</p><h3 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">4.<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">SmallInteger</code><span> </span>overflow for retry counters</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/client.py#L59-L60" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80); text-decoration: rgb(230, 180, 80);">client.py:59-60</a> — <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">attempts_count</code> and <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">deliveries_count</code> use <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">SmallInteger</code> (16-bit), but unbounded retry strategies (<code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">max_attempts=None, max_total_delay_seconds=None</code>) are supported and tested. Should use <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">Integer</code>.</p><hr style="font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><h2 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Race Conditions</h2><h3 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">5.<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_result_buffer</code><span> </span>accessed without consistent locking</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/subscriber/usecase.py#L184" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80); text-decoration: rgb(230, 180, 80);">subscriber/usecase.py:184</a> — <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_buffer_results()</code> appends to <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_result_buffer</code> <strong>without</strong> acquiring <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_result_buffer_lock</code>, while <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_flush_results()</code> at line 223 does acquire it. Since worker coroutines and the flush loop run concurrently, a worker appending during <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">list() + clear()</code> in flush could lose a message.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong>Fix:</strong> Always acquire the lock in <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_buffer_results</code>, or use an <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">asyncio.Queue</code>.</p><h3 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">6. Worker/requeue race during shutdown</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/subscriber/usecase.py#L96-L103" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80); text-decoration: rgb(230, 180, 80);">subscriber/usecase.py:96-103</a> — <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_stop_event.set()</code> is called, then <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_requeue_pending_consume_queue()</code> drains the queue. But a worker could <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">get()</code> a message after the drain but before it sees the stop event, process it, and buffer results <em>after</em> the final flush. That message update would be lost.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong>Fix:</strong> After <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_finalize_workers</code>, check <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_result_buffer</code> and flush once more.</p><hr style="font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><h2 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Design Issues</h2><h3 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">7. Broker unconditionally disposes user-provided engine</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/broker/broker.py#L130-L138" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80); text-decoration: rgb(230, 180, 80);">broker/broker.py:130-138</a> — <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">engine.dispose(close=True)</code> is called on stop, but the engine is user-provided. If shared across components, this kills all connections. Consider adding an <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">owns_engine</code> flag or not disposing at all.</p><h3 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">8. Timezone stripping without UTC conversion</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Throughout the codebase, <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">datetime.now(timezone.utc).replace(tzinfo=None)</code> is used. But in <a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/response.py#L25-L37" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80); text-decoration: rgb(230, 180, 80);">response.py:25-37</a>, user-provided <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">next_attempt_at</code> with a non-UTC timezone gets the tzinfo stripped without conversion, silently storing the wrong time.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong>Fix:</strong> <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">next_attempt_at.astimezone(timezone.utc).replace(tzinfo=None)</code>.</p><h3 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">9. Non-atomic archive with no conflict handling</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/client.py#L241-L268" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80); text-decoration: rgb(230, 180, 80);">client.py:241-268</a> — If a transient failure causes a retry of <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">archive()</code>, the INSERT into the archive table will fail with a duplicate key error, leaving the message stuck in the primary table forever. Use <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">INSERT ... ON CONFLICT DO NOTHING</code>.</p><hr style="font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><h2 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Truthiness Bugs (use<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">is not None</code>)</h2><ul style="padding-inline-start: 2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li><a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/message.py#L132" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80);">message.py:132</a>:<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">if max_deliveries and ...</code><span> </span>—<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">max_deliveries=0</code><span> </span>silently allows infinite deliveries</li><li><a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/retry.py#L37" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80);">retry.py:37</a>:<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">if self.max_attempts and ...</code><span> </span>—<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">max_attempts=0</code><span> </span>treated as no limit</li><li><a href="vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/retry.py#L83" target="_blank" rel="noopener noreferrer" style="color: rgb(230, 180, 80);">retry.py:83</a>:<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">if self.max_delay_seconds:</code><span> </span>—<span> </span><code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">0.0</code><span> </span>treated as no cap</li></ul><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">All of these should use <code style="font-family: monospace; color: rgb(179, 177, 173); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">if X is not None</code> instead of truthiness.</p><hr style="font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><h2 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Lower Priority</h2>
# | Issue | Location
-- | -- | --
10 | SQL injection in test fixture (f"CREATE DATABASE {worker_id}") | conftest.py:91
11 | SqlaSubscriberSpecification returns "STUB" — no AsyncAPI spec support | subscriber/specification.py
12 | Most tests never call broker.stop(), leaking background tasks | test_consume.py
13 | fetch() with empty queues silently returns nothing instead of erroring | client.py:188
14 | _types_compatible doesn't verify two Enum columns represent the same enum type | client.py:500

<hr style="font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><h2 style="color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Overall Assessment</h2><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(179, 177, 173); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(10, 14, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The architecture is solid — the polling design, worker pool, batched flushing, and retry strategies are well-designed for a DB-backed queue. The PR description and documentation are thorough. However, <strong>items 1-6 should be fixed before merge</strong> as they represent crashes in normal usage paths and data loss scenarios during shutdown. Items 7-9 are important design issues worth discussing. The truthiness bugs are easy wins.</p>Code Review: PR #2704 — feat: add sqla broker
Author: Arseniy-Popov | +6,033 / -89 across 68 files

This PR adds a SQLAlchemy-based message broker that implements a transactional outbox pattern using relational databases as message queues. The design is well thought-out (polling with SELECT FOR UPDATE SKIP LOCKED, batched flushing, worker pools), and the PR description is excellent. However, there are several issues that should be addressed before merging.

Critical Bugs
1. headers=None crashes the parser
[parser.py:23](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/parser.py#L23) — message.headers.get("content-type") will raise AttributeError when headers is None, which is allowed since the column is nullable=True in [client.py:50](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/client.py#L50) and enqueue() accepts headers=None.

Fix: Use (message.headers or {}).get(...).

2. Validation function is defined but never called
[subscriber/factory.py:56-104](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/subscriber/factory.py#L56-L104) — _validate_input_for_misconfiguration() contains important guardrails (checking max_deliveries > 0, warning about incompatible ack_policy/retry_strategy combos) but is dead code — never called from create_subscriber.

3. NoRetryStrategy.get_next_attempt_at raises ValueError unexpectedly
[retry.py:121-133](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/retry.py#L121-L133) — When max_attempts > 1, calling get_next_attempt_at for attempts_count < max_attempts raises ValueError instead of returning a datetime. The _nack flow in [message.py:89-98](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/message.py#L89-L98) will crash the worker.

4. SmallInteger overflow for retry counters
[client.py:59-60](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/client.py#L59-L60) — attempts_count and deliveries_count use SmallInteger (16-bit), but unbounded retry strategies (max_attempts=None, max_total_delay_seconds=None) are supported and tested. Should use Integer.

Race Conditions
5. _result_buffer accessed without consistent locking
[subscriber/usecase.py:184](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/subscriber/usecase.py#L184) — _buffer_results() appends to _result_buffer without acquiring _result_buffer_lock, while _flush_results() at line 223 does acquire it. Since worker coroutines and the flush loop run concurrently, a worker appending during list() + clear() in flush could lose a message.

Fix: Always acquire the lock in _buffer_results, or use an asyncio.Queue.

6. Worker/requeue race during shutdown
[subscriber/usecase.py:96-103](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/subscriber/usecase.py#L96-L103) — _stop_event.set() is called, then _requeue_pending_consume_queue() drains the queue. But a worker could get() a message after the drain but before it sees the stop event, process it, and buffer results after the final flush. That message update would be lost.

Fix: After _finalize_workers, check _result_buffer and flush once more.

Design Issues
7. Broker unconditionally disposes user-provided engine
[broker/broker.py:130-138](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/broker/broker.py#L130-L138) — engine.dispose(close=True) is called on stop, but the engine is user-provided. If shared across components, this kills all connections. Consider adding an owns_engine flag or not disposing at all.

8. Timezone stripping without UTC conversion
Throughout the codebase, datetime.now(timezone.utc).replace(tzinfo=None) is used. But in [response.py:25-37](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/response.py#L25-L37), user-provided next_attempt_at with a non-UTC timezone gets the tzinfo stripped without conversion, silently storing the wrong time.

Fix: next_attempt_at.astimezone(timezone.utc).replace(tzinfo=None).

9. Non-atomic archive with no conflict handling
[client.py:241-268](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/client.py#L241-L268) — If a transient failure causes a retry of archive(), the INSERT into the archive table will fail with a duplicate key error, leaving the message stuck in the primary table forever. Use INSERT ... ON CONFLICT DO NOTHING.

Truthiness Bugs (use is not None)
[message.py:132](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/message.py#L132): if max_deliveries and ... — max_deliveries=0 silently allows infinite deliveries
[retry.py:37](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/retry.py#L37): if self.max_attempts and ... — max_attempts=0 treated as no limit
[retry.py:83](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/retry.py#L83): if self.max_delay_seconds: — 0.0 treated as no cap
All of these should use if X is not None instead of truthiness.

Lower Priority
#	Issue	Location
10	SQL injection in test fixture (f"CREATE DATABASE {worker_id}")	[conftest.py:91](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/tests/brokers/sqla/conftest.py#L91)
11	SqlaSubscriberSpecification returns "STUB" — no AsyncAPI spec support	[subscriber/specification.py](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/subscriber/specification.py)
12	Most tests never call broker.stop(), leaking background tasks	[test_consume.py](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/tests/brokers/sqla/test_consume.py)
13	fetch() with empty queues silently returns nothing instead of erroring	[client.py:188](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/client.py#L188)
14	_types_compatible doesn't verify two Enum columns represent the same enum type	[client.py:500](vscode-webview://0m3f5p3c1nsq9s6qold1bivlkqllbn7r3th4rlsot0576t4lil4a/faststream/sqla/client.py#L500)
Overall Assessment
The architecture is solid — the polling design, worker pool, batched flushing, and retry strategies are well-designed for a DB-backed queue. The PR description and documentation are thorough. However, items 1-6 should be fixed before merge as they represent crashes in normal usage paths and data loss scenarios during shutdown. Items 7-9 are important design issues worth discussing. The truthiness bugs are easy wins.

Arseniy-Popov:
> 6. Worker/requeue race during shutdown

This is wrong: there is a final flush after workers are stopped: [faststream/sqla/subscriber/usecase.py#L112](https://github.com/Arseniy-Popov/faststream/blob/feat/sqla-broker/faststream/sqla/subscriber/usecase.py#L112).

Arseniy-Popov:
> 3. NoRetryStrategy.get_next_attempt_at raises ValueError unexpectedly  

`NoRetryStrategy` getting `attempts_count` that is less than 1 is erroneous behaviour.

Arseniy-Popov:
> 8. Timezone stripping without UTC conversion  

Fixed in https://github.com/ag2ai/faststream/pull/2704/changes/15d8f537892bee99b5409f01f96a9c1daadf1688.

Sehat1137:
I think an `sqla` broker is a poor solution for this use case, and I believe it would be better to move in the direction of a dedicated PostgreSQL broker instead.

My main concern is that a generic SQLAlchemy-based broker forces the design around the lowest common denominator across databases. In practice, that means polling becomes the core mechanism, because the implementation has to stay compatible with backends that do not provide broker-oriented primitives.

That tradeoff does not look attractive to me here.

Polling is a universal approach, but it comes with clear costs:
- higher latency
- constant database load even under low traffic
- worse efficiency compared to event-driven mechanisms
- architecture shaped by generic compatibility rather than by the strongest backend capabilities

Because of that, I do not think a universal SQL-backed broker is the right feature to prioritize in this case. If we want a database-backed broker in FastStream, I think we should optimize for the strongest and most relevant real-world case: PostgreSQL.

PostgreSQL has database-specific functionality such as `LISTEN/NOTIFY`, and that should influence the design directly instead of being ignored in favor of a polling model chosen for genericity. A PostgreSQL-first broker would also not necessarily need SQLAlchemy at all, which makes me question whether `sqla` is the right abstraction layer for this feature.

From a product perspective, PostgreSQL alone likely covers the large majority of meaningful demand here. Supporting a generic SQLA broker means paying the complexity and performance cost of portability up front, while the main value seems to come from one backend anyway.

Arseniy-Popov:
> I think an `sqla` broker is a poor solution for this use case, and I believe it would be better to move in the direction of a dedicated PostgreSQL broker instead.
> 
> My main concern is that a generic SQLAlchemy-based broker forces the design around the lowest common denominator across databases. In practice, that means polling becomes the core mechanism, because the implementation has to stay compatible with backends that do not provide broker-oriented primitives.
> 
> That tradeoff does not look attractive to me here.
> 
> Polling is a universal approach, but it comes with clear costs:
> 
> * higher latency
> * constant database load even under low traffic
> * worse efficiency compared to event-driven mechanisms
> * architecture shaped by generic compatibility rather than by the strongest backend capabilities
> 
> Because of that, I do not think a universal SQL-backed broker is the right feature to prioritize in this case. If we want a database-backed broker in FastStream, I think we should optimize for the strongest and most relevant real-world case: PostgreSQL.
> 
> PostgreSQL has database-specific functionality such as `LISTEN/NOTIFY`, and that should influence the design directly instead of being ignored in favor of a polling model chosen for genericity. A PostgreSQL-first broker would also not necessarily need SQLAlchemy at all, which makes me question whether `sqla` is the right abstraction layer for this feature.
> 
> From a product perspective, PostgreSQL alone likely covers the large majority of meaningful demand here. Supporting a generic SQLA broker means paying the complexity and performance cost of portability up front, while the main value seems to come from one backend anyway.

Listen/notify is not a replacement for polling even in Postgres:
- no persistence: a notification may fail to arrive during consumer restarts/deploys or network blips
- no acknowledgments: a notification is considered delivered on.. delivery, not on processing
- no work sharing: notifications aren't distributed between consumers
- no backpressure: notifications are pushed to a consumer without the ability to throttle the flow
- no delays: notifications aren't a match for delayed deliveries.

That said, listen/notify can be integrated into the current polling-based solution to reduce the frequency of idle polling and to "wake up" polling when messages appear in the queue.

P.S. SQLite is already supported.

P.P.S. SQLAlchemy is indeed not needed in the consumer. This is a potential future optimization. In the producer, the ability to send within a transaction opened by the "client" to the database somewhere in the application is needed. Currently only SQLAlchemy is supported, but this list can be extended. The name should probably be changed from SqlaBroker to SqlBroker, since there is no strict coupling to SQLAlchemy.

Arseniy-Popov:
> 1. headers=None crashes the parser  

This is wrong: headers dict is always added on send in [faststream/sqla/publisher/producer.py#L77](https://github.com/Arseniy-Popov/faststream/blob/feat/sqla-broker/faststream/sqla/publisher/producer.py#L77).

> 2. Validation function is defined but never called  

Fixed in https://github.com/ag2ai/faststream/pull/2704/changes/3683ccb27f8b55c1ca3e9573d1527d9152ad99ed.

> 3. NoRetryStrategy.get_next_attempt_at raises ValueError unexpectedly  

`NoRetryStrategy` getting `attempts_count` that is less than 1 is erroneous behaviour.

> 4. SmallInteger overflow for retry counters  

Fixed in https://github.com/ag2ai/faststream/pull/2704/changes/77cec0036b0259e58a7654acc817e7874b10790b.

> 5. _result_buffer accessed without consistent locking  

This one is wrong: there is no `await` between `list` and `clear`. To keep this clear moved this to a sync method: https://github.com/ag2ai/faststream/pull/2704/changes/b1f392e8b1409a1b1c9774069718f3777e653d9d.

> 6. Worker/requeue race during shutdown

This is wrong: there is a final flush after workers are stopped: [faststream/sqla/subscriber/usecase.py#L112](https://github.com/Arseniy-Popov/faststream/blob/feat/sqla-broker/faststream/sqla/subscriber/usecase.py#L112).  

> 8. Timezone stripping without UTC conversion  

Fixed in https://github.com/ag2ai/faststream/pull/2704/changes/15d8f537892bee99b5409f01f96a9c1daadf1688.

> 9. Non-atomic archive with no conflict handling  

Fixed in https://github.com/ag2ai/faststream/pull/2704/changes/2c06080bc45e7906f4b3360ffda17620503a0c80.

> Truthiness Bugs (use is not None)  

Fixed in https://github.com/ag2ai/faststream/pull/2704/changes/be78542d682a6961a323f1c2733442692c204564.

> Most tests never call broker.stop()  

Fixed in https://github.com/ag2ai/faststream/pull/2704/changes/58417ee6ce08c9ea960ce42b926fa435cf18035e.

> _types_compatible doesn't verify two Enum columns represent the same enum type  

Fixed in https://github.com/ag2ai/faststream/pull/2704/changes/0f87113118138357cd72d236879ccd680514215b.

.github/workflows/pr_tests.yaml
```diff
@@ -456,6 +456,61 @@ jobs:
...
+    services:
+      postgres:
+        image: postgres:latest # zizmor: ignore[unpinned-images]
+        env:
+          POSTGRES_DB: broker
+          POSTGRES_USER: broker
+          POSTGRES_PASSWORD: brokerpass # pragma: allowlist secret
+        ports:
+          - 5432:5432
+      mysql:
```
Lancetnik, line 536: Do we really need 2 databases to test implementation?

docs/docs_src/sqla/publish.py
```diff
@@ -0,0 +1,30 @@
...
+publisher_sqla = broker.publisher()
+
+@app.after_startup
+async def publish_examples():
+    await publisher_sqla.publish("Hello, SQLA!", queue="my_queue")
+
+    await publisher_sqla.publish(
+        "Process me later",
+        queue="my_queue",
+        next_attempt_at=datetime.now(timezone.utc) + timedelta(minutes=5),
```
Lancetnik, line 21: Do we really need this logic? If so, can we revisit naming - I believe, we have smth close in NATS

docs/docs_src/sqla/subscribe.py
```diff
@@ -0,0 +1,32 @@
+from sqlalchemy.ext.asyncio import create_async_engine
+
+from faststream import FastStream
+from faststream.middlewares.acknowledgement.config import AckPolicy
```
Lancetnik:
```suggestion
from faststream import AckPolicy
```

docs/docs_src/sqla/tables.py
```diff
@@ -0,0 +1,73 @@
...
+
+metadata = MetaData()
+
+message = Table(
+    "message",
+    metadata,
+    Column("id", BigInteger, primary_key=True),
+    Column("queue", String(255), nullable=False, index=True),
+    Column("headers", JSON, nullable=True),
+    Column("payload", LargeBinary, nullable=False),
```
Lancetnik, line 25: Can we make it configurable at broker-level? `LargeBinary` is too large (64kb) for common cases. We can use Binary, LargeBinary or even Text (with json / bas64 encoded payloads) to store information

faststream/_internal/endpoint/subscriber/usecase.py
Lancetnik, line 1: Do we need this file changes?

faststream/_internal/broker/broker.py
Lancetnik, line 1: Seems unrelated to current PR

tests/brokers/base/connection.py
Lancetnik, line 1: Seems unrelated

tests/brokers/sqla/conftest.py
```diff
@@ -0,0 +1,205 @@
...
+        case "sqlite":
+            timestamp_type = DateTime
+            json_type = JSON
+            pk_type = BigInteger().with_variant(Integer, "sqlite")
+        case _:
+            raise ValueError
+
+    metadata = MetaData()
+
+    message = Table(  # noqa: F841
```
Lancetnik, line 128: Is it copy? Can we just import original tables?

tests/brokers/sqla/conftest.py
```diff
@@ -0,0 +1,205 @@
...
+        await conn.run_sync(metadata.drop_all, checkfirst=True)
+        await conn.run_sync(metadata.create_all)
+
+
+@pytest_asyncio.fixture
+async def settings(engine: AsyncEngine, recreate_tables: None) -> Settings:
+    return Settings(engine=engine)
+
+
+def as_datetime(value: datetime | str) -> datetime:
```
Lancetnik: Do not place helper functions in `conftest.py`. Any file should no import conftest explicitly

tests/brokers/sqla/test_ack_policy.py
```diff
@@ -0,0 +1,309 @@
...
+
+        await broker.publish({"message": "hello1"}, queue="default1")
+        await broker.start()
+
+        await asyncio.sleep(0.5)
+
+        async with engine.begin() as conn:
+            result = await conn.execute(text("SELECT * FROM message;"))
+        result = result.mappings().one()
+        assert result["queue"] == "default1"
```
Lancetnik, line 64: Rewrite such test assertion blocks with dirty-equals

tests/brokers/sqla/test_ack_policy.py
```diff
@@ -0,0 +1,309 @@
...
+        )
+
+    @pytest.mark.asyncio()
+    async def test_consume_manual_no_manual_ack(
+        self, engine: AsyncEngine, recreate_tables: None, event: asyncio.Event
+    ) -> None:
+        """Error was logged because the message was neither manually nor automatically acknowledged.
+        The message was Reject'ed.
+        """
+        logger = MagicMock()
```
Lancetnik, line 273: use `mock` fixtures

ruff.toml
```diff
@@ -80,6 +80,7 @@ ignore = [
     "D105",     # missing docstring in magic methods
     "D106",     # missing docstring in public nested class
     "D107",     # missing docstring in __init__
+    "D205",     # 1 blank line required between summary line and description
```
Lancetnik, line 83: Remove

pyproject.toml
Lancetnik, line 1: Can we make `SqlaBroker` driver-independent to let users choice it by theirself?

docker-compose.yaml
```diff
@@ -41,6 +41,31 @@ services:
...
+    environment:
+      POSTGRES_DB: broker
+      POSTGRES_USER: broker
+      POSTGRES_PASSWORD: brokerpass # pragma: allowlist secret
+    ports:
+      - "5432:5432"
+    security_opt:
+      - no-new-privileges:true
+
+  mysql:
```
Lancetnik, line 55: Do we need 2 databases to run tests?

tests/brokers/sqla/test_schema_validation.py
Lancetnik, line 1: I don't see benefits of such tests. They are too fragile

faststream/sqla/broker/logging.py
```diff
@@ -0,0 +1,50 @@
...
+    def get_logger(self, *, context: "ContextRepo") -> "LoggerProto":
+        message_id_ln = 10
+
+        # TODO: generate unique logger names to not share between brokers
+        if not (lg := self._get_logger_ref()):
+            lg = get_broker_logger(
+                name="sqla",
+                default_context={},
+                message_id_ln=message_id_ln,
+                fmt="".join((
```
Lancetnik, line 34: I don't see information about listening queue

faststream/sqla/publisher/producer.py
```diff
@@ -0,0 +1,88 @@
...
+    def disconnect(self) -> None: ...
+
+    @abstractmethod
+    async def publish(self, cmd: "SqlaPublishCommand") -> None: ...
+
+    async def request(self, cmd: "SqlaPublishCommand") -> None:
+        msg = "SqlaBroker doesn't support synchronous requests."
+        raise FeatureNotSupportedException(msg)
+
+    async def publish_batch(self, cmd: "SqlaPublishCommand") -> None:
```
Lancetnik, line 38: Can we support batching?

Lancetnik (changes requested):
Also, I didn't find `TestBroker` and Opentelemetry integrations

.github/workflows/pr_tests.yaml
Arseniy-Popov: Yes, the queries are quite different.

docs/docs_src/sqla/publish.py
Arseniy-Popov: "Schedule" is nats-specific wording that is not as immediately clear outside that context as "next_attempt_at".

tests/brokers/sqla/conftest.py
Arseniy-Popov: These use specific types: https://github.com/Arseniy-Popov/faststream/blob/feat/sqla-broker/tests/brokers/sqla/conftest.py#L115-L129

lesnik512:
Hi) here is my solution for outbox pattern https://github.com/modern-python/faststream-outbox

Will try in production soon

Arseniy-Popov:
@lesnik512 It's not so much "yours" as it is my code that you ~~stole~~copied, so your presence is unwelcome in this discussion.

lesnik512:
@Arseniy-Popov 
You're right that the design of this package comes from your PR #2704 — the FastStream broker/registrator/subscriber wiring, the file layout, the SELECT … FOR UPDATE SKIP  LOCKED fetch-CTE, the retry-strategy hierarchy, and the in-transaction publish contract all originate from your work. I should have credited that from day one and I'm sorry I didn't. I've just added an Acknowledgements section to the README pointing at #2704: https://github.com/modern-python/faststream-outbox/blob/main/README.md#acknowledgements

To be precise about what's in the repo: none of it is a verbatim copy — every file was rewritten. It's a Postgres-only reimplementation that drops the state column and archive table in favor of a lease-token model, collapses your four loops down to two, adds LISTEN/NOTIFY wake-ups, and adds timer mechanics (activate_in / activate_at / timer_id / cancel_timer) that aren't in your PR. But the design lineage is unambiguously yours and the README now says so.

Arseniy-Popov:
@lesnik512 I appreciate that. Glad this could be of inspiration. Maybe we'll compare notes later on.

docs/docs_src/sqla/tables.py
Arseniy-Popov: Column type can be any of `(LargeBinary, BLOB, BINARY, VARBINARY)`.

docs/docs_src/sqla/subscribe.py
Arseniy-Popov: 👍

tests/brokers/sqla/test_schema_validation.py
Arseniy-Popov: I don't see what's so fragile about them.

docker-compose.yaml
Arseniy-Popov: Yes, the queries are quite different.

faststream/sqla/publisher/producer.py
Arseniy-Popov: Added that.

tests/brokers/sqla/conftest.py
Arseniy-Popov: Fixed.

Lancetnik:
Hey @Arseniy-Popov, thank you so much for the effort you've put into this — the design is thoughtful, the PR description is excellent, and the implementation is solid.

After careful consideration, we've decided not to merge this into the core repository. Our roadmap for 1.0.0 is to keep a focused set of brokers in core (RMQ, Kafka, NATS, Redis, MQTT, SQS, ZMQ), and support for additional brokers — including SQL-based ones — will be community-driven. Our team is fairly small, and we simply don't have the capacity to maintain every broker ourselves.

The good news: we're planning to continue this work as a community-maintained package at https://github.com/faststream-community/faststream-sqlbroker, and we'd love your work to serve as the foundation there. We'll be happy to point to it from the main docs once it's ready.

Thank you again for the contribution!

PR #2705, opened 2025-12-02
feat: merge `Broker(context=...)` and `FastStream(context=...)`

### LINKED ISSUE #2693: feat: merge `Broker(context=...)` and `FastStream(context=...)` at broker-level
I recently migrated my project to faststream 0.6. I was very interested in how I could add my dependencies to the context. Prior to version 0.6, I did something like this:

```python
from faststream.annotations import ContextRepo
from faststream.utils.context import context

broker = KafkaBroker()


@broker.subscriber("my_topic", group_id="my_group")
async def handle(
    context: ContextRepo,
):
    print("dependency: ", context.get("dependency"))  # 42


async def lifespan(*args, **kwargs):
    context.set_global("dependency", 42)

    await broker.start()
    try:
        yield
    finally:
        await broker.stop()
```

I launched the broker as part of my application in lifespan without using the FastStream class.

For version 0.6, I saw examples where it was suggested to pass the context to FastStream, but that solution did not suit me. I discovered that the broker also accepts context, and that solves my problem:

```python
broker = KafkaBroker(context=ContextRepo({"dependency": 42}))

...

async def lifespan(*args, **kwargs):
    await broker.start()
    try:
        yield
    finally:
        await broker.stop()
```

But I also discovered that if I create a FastStream instance, its context will be used, even though I didn't use it to start the broker.

```python
from fastapi import FastAPI
from faststream import ContextRepo, FastStream

broker = KafkaBroker(context=ContextRepo({"broker_dependency": 2}))
app = FastStream(broker, context=ContextRepo({"application_dependency": 1}))


@broker.subscriber("my_topic", group_id="my_group")
async def handle(
    context: ContextRepo,
):
    print("broker_dependency: ", context.get("broker_dependency"))            # None
    print("application_dependency: ", context.get("application_dependency"))  # 1


async def lifespan(*args, **kwargs):
    await broker.start()
    try:
        yield
    finally:
        await broker.stop()


asgi = FastAPI(lifespan=lifespan)
```

I'm not sure that's normal behavior. It would make much more sense if only the broker's dependency were available.

---

Running FastStream 0.6.3 with CPython 3.12.4 on Linux

nectarindev:
# Description

Fixes #2693

## Changes

## Checklist

### DIFF
--- a/faststream/_internal/context/repository.py
+++ b/faststream/_internal/context/repository.py
@@ -17,7 +17,7 @@ class ContextRepo:
             _global_context : a dictionary representing the global context
             _scope_context : a dictionary representing the scope context
         """
-        self._global_context: dict[str, Any] = {"context": self} | (initial or {})
+        self._global_context: dict[str, Any] = (initial or {}) | {"context": self}
         self._scope_context: dict[str, ContextVar[Any]] = {}
 
     @property
@@ -164,5 +164,21 @@ class ContextRepo:
         return v
 
     def clear(self) -> None:
+        """Reset global and scope contexts.
+
+        Returns:
+            None
+        """
         self._global_context = {"context": self}
         self._scope_context.clear()
+
+    def merge_global(self, other: "ContextRepo") -> None:
+        """Merge the global context of another repository into the current one.
+
+        Args:
+            other: Another ContextRepo instance.
+
+        Returns:
+            None
+        """
+        self._global_context |= other._global_context | {"context": self}
--- a/faststream/_internal/di/config.py
+++ b/faststream/_internal/di/config.py
@@ -50,6 +50,7 @@ class FastDependsConfig:
 
     def __or__(self, value: "FastDependsConfig", /) -> "FastDependsConfig":
         use_fd = False if not value.use_fastdepends else self.use_fastdepends
+        self.context.merge_global(value.context)
 
         return FastDependsConfig(
             use_fastdepends=use_fd,
--- a/docs/docs/en/getting-started/context.md
+++ b/docs/docs/en/getting-started/context.md
@@ -58,6 +58,10 @@ You can store your own objects in the `Context`.
 
 ### Global
 
+Global context fields can be set at both the application-level and the broker-level.
+
+#### Application-level
+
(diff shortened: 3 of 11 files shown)


nectarindev:
The current implementation merges the broker context into the application context, so a single context will be shared between multiple brokers.
I tried to make the context isolated between brokers, but the mutable design of the library makes this very difficult. If I merge the application context into the broker context, subsequent changes to the application context, such as a modification in app.on_startup, will not be visible in the brokers context.

If the current version is accepted, I would further supplement the documentation.

faststream/_internal/context/repository.py
```diff
@@ -164,5 +164,21 @@ def resolve(self, argument: str) -> Any:
...
+    def merge_global(self, other: "ContextRepo") -> None:
+        """Merge the global context of another repository into the current one.
+
+        Args:
+            other: Another ContextRepo instance.
+
+        Returns:
+            None
+        """
+        self._global_context |= other._global_context | {"context": self}
```
borisalekseev, line 184: I think the result of merging should be a new context object and may be implemented using the `__or__` method. What do you think, @Lancetnik?

faststream/_internal/di/config.py
borisalekseev, line 59: This way, here will be `context=self.context|value.context`

faststream/_internal/context/repository.py
nectarindev: We cannot create a new instance because once the broker is binded to the application, subsequent changes in the application context, for example, at startup, will not be visible in the broker.

borisalekseev: It's not clear to me. Can you please write a test case when the behavior is present?

nectarindev:
```python
broker = KafkaBroker(context=ContextRepo({"broker_dependency": 1}))
app = FastStream(broker, context=ContextRepo({"application_dependency": 2})

@app.on_startup
async def setup(context: ContextRepo):
    # will not be visible in broker if broker holds new ContextRepo instance
    context.set_global("setup_dependency", 3)
```

IvanKirpichnikov:
#2693

PR #2718, opened 2025-12-17
Feat/add options support

### LINKED ISSUE #2658: Feature: Add OPTIONS support for GET/POST endpoints
Currently the framework handles `GET` and `POST` routes, but does not automatically expose `OPTIONS` handlers.  
Adding `OPTIONS` support would improve compatibility with browsers (CORS preflight), API introspection tools and frontend clients that expect `OPTIONS` responses.  
It also reduces boilerplate: developers wouldn’t need to manually define an `OPTIONS` route for every endpoint.

I would be happy to help with PR, I already have a minimal code that solves the problem if we do this. Please let me know if there are any preferences or limitations that I should be aware of in this case.

Shepard2154:
## Summary

This PR adds automatic OPTIONS request support for ASGI routes created with `@get` and `@post` decorators. This enables CORS preflight requests from browsers without requiring manual OPTIONS route definitions.

**Changes:**
1. **Automatic OPTIONS request support** - Added automatic OPTIONS handling in `HttpHandler` that returns `204 No Content` with basic CORS headers. CORS headers are also added to main responses (GET, POST) when `Origin` is present in the request. This provides minimal CORS support: returns the `Origin` from the request, allows route methods, and the `Content-Type` header.

2. **Tests** - Added parameterized test `test_options_request` to verify OPTIONS handling for both GET and POST routes.

3. **Documentation** - Added a note about automatic OPTIONS support in ASGI documentation.

**Result:**
Browsers can now perform CORS preflight requests without manually defining OPTIONS routes. This provides basic CORS support for simple use cases.

**Code examples:**
OPTIONS requests are automatically handled for all routes created with `@get` or `@post` decorators:
```python
from faststream.asgi import AsgiFastStream, AsgiResponse, get
from faststream.asgi.types import Scope

broker = NatsBroker()

@get
async def liveness_ping(scope: Scope) -> AsgiResponse:
    return AsgiResponse(b"", status_code=200)

app = AsgiFastStream(
    broker,
    asgi_routes=[("/health", liveness_ping)],
)
```
You can test OPTIONS requests using curl:
```bash
curl -X OPTIONS http://localhost:8000/health \
  -H "Origin: http://example.com" -v
```

Resolves #2658

## Type of change

## Checklist

### DIFF
--- a/faststream/asgi/handlers.py
+++ b/faststream/asgi/handlers.py
@@ -42,25 +42,25 @@ class HttpHandler:
 
     def _get_cors_headers(self, scope: "Scope") -> dict[str, str]:
         headers: dict[str, str] = {}
-        
+
         origin = None
         for key, value in scope.get("headers", []):
             if key.decode("latin-1").lower() == "origin":
                 origin = value.decode("latin-1")
                 break
-        
+
         if origin:
             headers["access-control-allow-origin"] = origin
             headers["access-control-allow-methods"] = ", ".join(self.methods) or "OPTIONS"
             headers["access-control-allow-headers"] = "Content-Type"
-        
+
         return headers
 
     def _merge_cors_headers(self, response: AsgiResponse, scope: "Scope") -> AsgiResponse:
         cors_headers = self._get_cors_headers(scope)
         if not cors_headers:
             return response
-        
+
         existing_keys = set()
         existing_headers = {}
         if response.raw_headers:
@@ -68,12 +68,10 @@ class HttpHandler:
                 key = key_bytes.decode("latin-1").lower()
                 existing_keys.add(key)
                 existing_headers[key] = value_bytes.decode("latin-1")
-        
-        merged_cors = {
-            k: v for k, v in cors_headers.items() if k not in existing_keys
-        }
+
+        merged_cors = {k: v for k, v in cors_headers.items() if k not in existing_keys}
         merged_headers = {**merged_cors, **existing_headers}
-        
+
         return AsgiResponse(
             body=response.body,
             status_code=response.status_code,
@@ -123,10 +121,10 @@ class HttpHandler:
                     response = AsgiResponse(
                         body=b"Internal Server Error", status_code=500
                     )
-        
+
         if isinstance(response, AsgiResponse):
             response = self._merge_cors_headers(response, scope)
-        
+
         await response(scope, receive, send)
 
     def update_fd_config(self, config: FastDependsConfig) -> None:
--- a/tests/asgi/testcase.py
+++ b/tests/asgi/testcase.py
@@ -241,9 +241,14 @@ class AsgiTestcase:
 
         async with self.get_test_broker(broker):
             with TestClient(app) as client:
-                response = client.options("/test", headers={"Origin": "http://example.com"})
+                response = client.options(
+                    "/test", headers={"Origin": "http://example.com"}
+                )
                 assert response.status_code == 204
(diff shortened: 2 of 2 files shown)


borisalekseev:
@Shepard2154 hello!

Thanks for the PR. I have a few concerns before we merge this as-is:

1. No actual CORS implementation
Returning responses to OPTIONS requests alone doesn’t implement CORS. It described in [mdn](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS) and implemented in [starlette](https://github.com/Kludex/starlette/blob/main/starlette/middleware/cors.py), for example.

2. Not opt-in / always enabled
As proposed, the behavior seems to be applied automatically for everyone. That’s a potentially breaking or at least surprising behavioral change, especially for users who want full control over routing/HTTP semantics, or who don’t want OPTIONS automatically answered at all. This feels like something that should be configurable / explicitly enabled.

**I think that ASGI middleware support might be a better foundation.**
Instead of baking OPTIONS (and potentially partial CORS behavior) into the core app, it might be better to support plugging in ASGI middleware. Then users can add `starlette.middleware.cors.CORSMiddleware` (or any other ASGI middleware) themselves, which is a standard, flexible solution and avoids forcing one behavior on all users.

If you’re open to it, I’d strongly prefer moving in the direction of “ASGI middleware support + documented CORS middleware example” rather than implementing OPTIONS handling universally in core.

Shepard2154:
> @Shepard2154 hello!
> 
> Thanks for the PR. I have a few concerns before we merge this as-is:
> 
> 1. No actual CORS implementation
>    Returning responses to OPTIONS requests alone doesn’t implement CORS. It described in [mdn](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS) and implemented in [starlette](https://github.com/Kludex/starlette/blob/main/starlette/middleware/cors.py), for example.
> 2. Not opt-in / always enabled
>    As proposed, the behavior seems to be applied automatically for everyone. That’s a potentially breaking or at least surprising behavioral change, especially for users who want full control over routing/HTTP semantics, or who don’t want OPTIONS automatically answered at all. This feels like something that should be configurable / explicitly enabled.
> 
> **I think that ASGI middleware support might be a better foundation.** Instead of baking OPTIONS (and potentially partial CORS behavior) into the core app, it might be better to support plugging in ASGI middleware. Then users can add `starlette.middleware.cors.CORSMiddleware` (or any other ASGI middleware) themselves, which is a standard, flexible solution and avoids forcing one behavior on all users.
> 
> If you’re open to it, I’d strongly prefer moving in the direction of “ASGI middleware support + documented CORS middleware example” rather than implementing OPTIONS handling universally in core.

@borisalekseev hello! In general, I agree with your comment. If I understood you correctly, then we need this option: 
```python
from faststream.asgi import AsgiFastStream, get, AsgiResponse
from starlette.middleware.cors import CORSMiddleware

broker = NatsBroker()

@get
async def health(scope) -> AsgiResponse:
    return AsgiResponse(b"", status_code=200)

app = AsgiFastStream(broker, asgi_routes=[("/health", health)])
asgi_app = CORSMiddleware(
    app,
    allow_origins=["*"],
    allow_methods=["*"],
    allow_headers=["*"],
)
```
By the way, this already works without additional modifications, then can I just specify this example in the documentation? But there is a caveat: after applying any middleware, the application will no longer be launched using the faststream command. Are we satisfied with this?

Lancetnik:
Sorry, this functional is not required for Try It Out feature anymore. So, I close the PR

Issue #2785, opened 2026-02-27
Bug: Expired is missing from rabbit testing header properties


marcm-ml:
Hello,

the expiration property is not respected when using the rabbit test broker and setting the "expiration" flag in the testbroker.publish method.

I think its simply missing here.

https://github.com/ag2ai/faststream/blob/8148f2de21dfc5e922421bdf1e1f5d5b86b36501/faststream/rabbit/testing.py#L173-L185

Up until this point in the code, the expiration property is propagated but not passed to the aiormq.abc.DeliveredMessage header property

Lancetnik:
@marcm-ml hi! Thank you for the report!
I think, the problem is a little bit dipper than just flag missing. We can set the headed, but it has no sense for testbroker. Test broker emulates message delivery by direct subscriber function execution. So, is is a synchronous call, not a real publishing.

What behavior you expect in this case?

marcm-ml:
Hey, thanks for the reply.

I have a test suite and just encountered this when writing the following test. I have already implemented the expected behavior by mocking the faststream build_message method.

```python
import asyncio
from datetime import datetime, timedelta
from typing import TYPE_CHECKING, Optional, Union
from unittest.mock import AsyncMock

import pytest
from aio_pika.message import encode_expiration
from faststream.rabbit import RabbitBroker
from faststream.rabbit.subscriber.usecase import RabbitSubscriber
from faststream.rabbit.testing import TestRabbitBroker
from pytest_mock import MockerFixture

if TYPE_CHECKING:
    from aio_pika.abc import DateType, HeadersType
    from fast_depends.library.serializer import SerializerProto
    from faststream.rabbit.schemas import RabbitExchange, RabbitQueue
    from faststream.rabbit.subscriber import RabbitSubscriber
    from faststream.rabbit.types import AioPikaSendableMessage


@pytest.mark.asyncio
@pytest.mark.parametrize(
    ("expiration", "expected"),
    [
        (None, None),
        (1, 1),
        (1.0, 1.0),
        (1e1, 1e1),
        (datetime(2025, 10, 1), encode_expiration(datetime(2025, 10, 1))),
        (timedelta(2), encode_expiration(timedelta(2))),
    ],
)
async def test_base_handler_expiration_value(
    mocker: MockerFixture,
    broker: RabbitBroker,
    secret: Any,
    consumer: RabbitSubscriber,
    sink_mock: AsyncMock,
    expiration,
    expected,
):
[...]

PR #2787, opened 2026-02-28
fix: propagate expiration property in rabbit test broker

marcm-ml:
# Description

Fixes https://github.com/ag2ai/faststream/issues/2785

## Type of change

## Checklist

### DIFF
--- a/tests/brokers/rabbit/test_test_client.py
+++ b/tests/brokers/rabbit/test_test_client.py
@@ -4,6 +4,7 @@ from typing import Any
 from unittest.mock import MagicMock
 
 import pytest
+from freezegun import freeze_time
 
 from faststream import BaseMiddleware
 from faststream.exceptions import SubscriberNotFound
@@ -19,6 +20,8 @@ from tests.brokers.base.testclient import BrokerTestclientTestcase
 
 from .basic import RabbitMemoryTestcaseConfig
 from .test_publish import TestPublishWithExchange as PublishWithExchangeCase
+
+_frozen_time = dt.datetime(2026, 2, 10, 12, 0, 0, tzinfo=dt.timezone.utc)
 
 
 @pytest.mark.rabbit()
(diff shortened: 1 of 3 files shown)


tests/brokers/rabbit/test_test_client.py
```diff
@@ -187,6 +189,40 @@ async def test_broker_with_real_patches_publishers_and_subscribers(
...
+    @pytest.mark.parametrize(
+        ("expiration", "expected"),
+        (
+            pytest.param(None, None, id="none"),
+            pytest.param(1, 1, id="int"),
+            pytest.param(1.5, 1.5, id="float"),
+            pytest.param(dt.timedelta(seconds=1.1), 1.1, id="timedelta"),
+            pytest.param(dt.datetime.now(tz=dt.timezone.utc), 10, id="datetime"),
+        ),
+    )
```
marcm-ml, line 205: two things:

1. I didn't see any test already testing test broker publish kwargs, so i created one but was not sure if i should test more than just "expiration"
2. I was not sure how far i need to go here in testing different types. for example, providing the datetime parameter turned out to be flaky as described in the comment below.

let me know what you think

faststream/rabbit/testing.py
```diff
@@ -178,6 +178,7 @@ def build_message(
                     content_encoding=msg.content_encoding,
                     priority=msg.priority,
                     correlation_id=msg.correlation_id,
+                    expiration=encode_expiration(msg.expiration),
```
marcm-ml, line 181: this is necessary, because the aio_pika.Message construct is not encoding expiration. It is most likely encoded when serializing the message to bytes. I wanted to capture this behavior.

tests/brokers/rabbit/test_test_client.py
```diff
@@ -187,6 +189,40 @@ async def test_broker_with_real_patches_publishers_and_subscribers(
...
+
+        @broker.subscriber(*args, **kwargs)
+        async def m(msg: RabbitMessage) -> None:
+            mock(msg)
+
+        async with self.patch_broker(broker) as br:
+            await br.start()
+            await br.publish("hello", queue, expiration=expiration)
+            msg = mock.call_args[0][0]
+            if isinstance(expiration, dt.datetime):
```
Lancetnik: I suggest to introduce https://github.com/spulec/freezegun to prevent such flaky tests

tests/brokers/rabbit/test_test_client.py
Lancetnik: I am fine with your current implementation, thank you

tests/brokers/rabbit/test_test_client.py
marcm-ml: great suggestion. i have added it hopefully to the correct dep-group

PR #2788, opened 2026-03-01
feat: use MRO-based exception handler resolution

### LINKED ISSUE #2400: Bug: incorrect error type traversal in error handlers
**Describe the bug**

Current logic [here](https://github.com/ag2ai/faststream/blob/main/faststream/middlewares/exception.py#L166) seems to be counterintuitive - take the current exception and check if it's a subclass of any exception type of the registered handlers. This leads to a potential problem that the order of handler registration really matters, but it shouldn't. 

**How to reproduce**

1. Say I have `class ExcA(Exception)` and `class ExcB(ExcA)` and I have `@exception_middleware.add_handler(ExcA)` defined _before_ `@exception_middleware.add_handler(ExcB)`
2. Then the handler for `ExcA` (say handler_A) will swallow all exceptions of ExcB type simply because this handler_A is registered first.

**How to fix**

Use starlette's [approach](https://github.com/encode/starlette/blob/master/starlette/_exception_handler.py#L16):
1. Convert `self._publish_handlers` and `self._handlers` to dicts (possibly a breaking change?)
2. Iterate over `type(exc).__mro__` and check if this type is in `self._publish_handlers` or `self._handlers` respectively
3. This guarantees that exception handling middleware will call the closest handler based on the current exception's mro rather than the order in which handlers were registered.

1. ...

**Expected behavior**
Explain what you expected to happen clearly and concisely.

**Observed behavior**
Describe what is actually happening clearly and concisely.

**Screenshots**
If applicable, attach screenshots to help illustrate the problem.

**Environment**
Include the output of the `faststream -v` command to display your current project and system environment.

**Additional context**
Provide any other relevant context or information about the problem here.

Br1an67:
# Description

Use `type(exc).__mro__` traversal to find the most specific exception handler instead of iterating handlers in registration order. This follows starlette's approach and ensures child exception handlers are always preferred over parent handlers regardless of registration order.

Fixes #2400

## Type of change

## Checklist

### Changes

**`faststream/middlewares/exception.py`**:
- Changed `_handlers` and `_publish_handlers` storage from `list[tuple[type, handler]]` to `dict[type, handler]`
- Updated `consume_scope` and `after_processed` to iterate `type(exc).__mro__` for handler lookup instead of iterating the handler list in registration order
- Updated `add_handler` to use dict assignment instead of list append

**`tests/brokers/base/middlewares.py`**:
- Updated `test_exception_middleware_init_handler_same` and `test_exception_middleware_init_publish_handler_same` for dict-based storage
- Added `test_exception_middleware_mro_resolution` verifying that a child exception handler is selected over a parent handler when the parent is registered first

### DIFF
--- a/faststream/confluent/broker/registrator.py
+++ b/faststream/confluent/broker/registrator.py
@@ -1,4 +1,4 @@
-from collections.abc import Iterable, Sequence
+from collections.abc import Callable, Iterable, Sequence
 from typing import (
     TYPE_CHECKING,
     Annotated,
@@ -78,6 +78,10 @@ class KafkaRegistrator(
             "read_uncommitted",
             "read_committed",
         ] = "read_uncommitted",
+        # rebalance callbacks
+        on_assign: Callable[..., None] | None = None,
+        on_revoke: Callable[..., None] | None = None,
+        on_lost: Callable[..., None] | None = None,
         batch: Literal[False] = False,
         max_records: int | None = None,
         # broker args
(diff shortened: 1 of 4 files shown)


Lancetnik:
Thank you for the PR!

Lancetnik (approved):

PR #2813, opened 2026-03-30
fix: include pattern subscribers in AsyncAPI specification

### LINKED ISSUE #2804: Bug: topic patterns are not correctly documented
**Describe the bug**
When using a pattern for a kafka topic the generated asyncapi spec (and docs) does not include that handler

**How to reproduce**
With something along the lines of:

```python

broker = KafkaBroker(...)

router = KafkaRouter()

@router.subscriber(pattern="some.wildcard.topic.*")
def handle_event(event: Event): ...

broker.include_router(router)

app = AsgiFastStream(
        broker,
        asyncapi_path="/docs",
    )
```

**Expected behavior**
The subscriber is included in the asyncapi spec and docs. As far as I can see the asyncapi spec actually allows for this even when using Path variables.

**Observed behavior**
The asyncapi docs don't contain that subscriber.

**Screenshots**
-

**Environment**
Running FastStream 0.6.7 with CPython 3.14.0 on Linux

**Additional context**

aazmv:
# Description

Subscribers created with `pattern=` (e.g. `@broker.subscriber(pattern="events.*")`)
were not showing up in the generated AsyncAPI spec or docs page.

The root cause is that `KafkaSubscriberSpecificationConfig` didn't have a `pattern`
field, so it was never passed to the specification layer. As a result,
`KafkaSubscriberSpecification.topics` returned an empty list for pattern-only
subscribers and `get_schema()` produced no channels.

The fix adds `pattern` to the specification config and includes it in the topic
list used for schema generation.

Fixes #2804

## Type of change

## Checklist

### DIFF
--- a/tests/asyncapi/kafka/v2_6_0/test_naming.py
+++ b/tests/asyncapi/kafka/v2_6_0/test_naming.py
@@ -61,5 +61,5 @@ class TestNaming(NamingTestCase):
         schema = self.get_spec(broker).to_jsonable()
 
         assert len(schema["channels"]) == 1
-        channel = list(schema["channels"].values())[0]
+        channel = next(iter(schema["channels"].values()))
         assert channel["bindings"]["kafka"]["topic"] == "events.*"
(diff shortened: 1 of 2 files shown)


aazmv:
> Please, add test to pin the behavior

Got it, I’ll add a test

aazmv:
> Please, add test to pin the behavior

Done. Added a schema test for pattern subscribers

Lancetnik:
Снимок экрана 2026-03-31 в 11 09 42

Please, apply ruff changes

aazmv:
> Снимок экрана 2026-03-31 в 11 09 42
> Please, apply ruff changes

Done

PR #2827, opened 2026-04-10
feat: support broker-level ack_policy with per-subscriber override

### LINKED ISSUE #2826: Feature: Support broker-level ack_policy default with per-subscriber override
## Problem

Currently `ack_policy` can only be set per-subscriber:

```python
@broker.subscriber("orders.topic", ack_policy=AckPolicy.NACK_ON_ERROR)
async def handle_order(msg: OrderMessage) -> None:
    ...
```

There is no way to set a default `ack_policy` at the broker (or router) level. In practice, most services want the same policy across all subscribers — typically `NACK_ON_ERROR` for at-least-once delivery with DLQ. This forces every `@broker.subscriber()` call to repeat the same `ack_policy=` argument, which is error-prone: forgetting it on a single subscriber silently falls back to `ACK_FIRST` (at-most-once), which can cause silent message loss.

## Proposed solution

Add an `ack_policy` parameter to `KafkaBroker` (and equivalently to `RabbitBroker`, `NatsBroker`, etc.) that sets the default for all subscribers registered on that broker. Individual subscribers can still override it.

```python
broker = KafkaBroker("localhost:9092", ack_policy=AckPolicy.NACK_ON_ERROR)

# Inherits NACK_ON_ERROR from broker
@broker.subscriber("orders.topic")
async def handle_order(msg: OrderMessage) -> None:
    ...

# Overrides to ACK for this specific subscriber
@broker.subscriber("notifications.topic", ack_policy=AckPolicy.ACK)
async def handle_notification(msg: NotificationMessage) -> None:
    ...
```

The resolution order would be: **subscriber-level > broker-level > built-in default (`ACK_FIRST`)**.

## Why this matters

- **Safety**: `ACK_FIRST` as a silent default is dangerous for services that use DLQ or need at-least-once delivery. A broker-level default lets teams enforce their delivery guarantee in one place.
- **DRY**: Services with 10+ subscribers shouldn't need to repeat `ack_policy=AckPolicy.NACK_ON_ERROR` on every one.
- **Consistency with other broker-level settings**: `decoder`, `middlewares`, `security`, and `logger` are all broker-level defaults that subscribers inherit. `ack_policy` is the notable exception.

ce1ebrimbor:
# Description
Support broker-level ack_policy default with per-subscriber override

Fixes #2826

## Type of change

## Checklist

### DIFF
--- a/docs/docs/en/getting-started/acknowledgement.md
+++ b/docs/docs/en/getting-started/acknowledgement.md
@@ -49,6 +49,32 @@ Each `AckPolicy` variant includes behavior examples for both successful processi
 
 ---
 
+## Broker-Level Default
+
+By default, `ack_policy` is set per subscriber. If most of your subscribers share the same policy, you can set a broker-level default to avoid repetition:
+
+```python linenums="1" hl_lines="4 8 13"
+from faststream import FastStream, AckPolicy
+from faststream.nats import NatsBroker
+
+broker = NatsBroker(ack_policy=AckPolicy.NACK_ON_ERROR)
+app = FastStream(broker)
+
+# Inherits NACK_ON_ERROR from broker
+@broker.subscriber("test")
+async def process_order(msg: str) -> None:
+    ...
+
+# Overrides to MANUAL for this subscriber
+@broker.subscriber("events", ack_policy=AckPolicy.MANUAL)
+async def handle_event(msg: str) -> None:
+    await msg.ack()
+```
+
+The resolution order is: **subscriber-level > broker-level > built-in default**.
+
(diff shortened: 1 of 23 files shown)


Lancetnik (changes requested):
Hi @ce1ebrimbor ! Thank you for the PR! It looks good, but I have just a single comment about tests.
In FastStream we are trying to follow e2e tests everywhere it possible - they are much sustainable because doesn't rely on impl details and pin exact user API they should pin.

So, can you please rewrite your tests to cover real cases?

Like the follow

```python
def test_broker_ack_policy():
     broker = Broker(ack_policy=AckPolicy.MANUAL)
     subscriber = broker.subscriber("")
     assert subscriber.config.ack_policy is AckPolicy.MANUAL

def test_sub_policy_overrides_broker(): ...

def test_router_ack_policy(): ...

def test_broker_and_router_ack_policy(): ...
```

I imagine, that correctly priority should looks like: Subscriber > Router > Broker
And tests should represent it

Lancetnik (approved):

Lancetnik (approved):

PR #2828, opened 2026-04-13
fix: propagate outer context to nested StreamRouter on include

### LINKED ISSUE #2657: Bug: Empty messages published in KafkaBroker
**Describe the bug**
empty messages published by KafkaBroker

**How to reproduce**

```python
import typing

import faststream
from faststream.kafka import KafkaRouter


ROUTER: typing.Final = KafkaRouter()


@ROUTER.subscriber(
    "topic.in",
    group_id="some_group",
    ...
    ack_policy=faststream.AckPolicy.ACK,
)
@ROUTER.publisher(
    "topic.out",
    schema=OutPydanticModel,
)
async def operator_status_consumer(
    message: InPydanticModel,
) -> DataLakeOperatorStatusEvent:
    return OutPydanticModel.model_construct(...)

```

1. After updating from 0.5.48 to 0.6.2 in topic "topic.out" we started to receive empty messages: null key and body, header `content-type` is also empty.
2. The number of such messages is low: 6 out of thousands in couple of days
3. There is no errors in sentry and logs

lesnik512:
Closes #2657

# Description

When StreamRouter.include_router(inner: StreamRouter) was called, the inner router's subscribers still held a reference to the inner router's ContextRepo in their _call_decorators closure. process_message stored the native message in the outer ContextRepo, so FastAPI's Context('message') dependency (used by KafkaMessage, StreamMessage, etc.) looked in the wrong ContextRepo and returned EmptyPlaceholder.

Fix by recursively propagating the outer router's ContextRepo to all included StreamRouters at include_router time via _propagate_context.

Fixes #2657

## Checklist

  - My code adheres to the style guidelines of this project (just lint shows no errors)                                                                                         
  - I have conducted a self-review of my own code                                                                                                                             
  - My changes do not generate any new warnings                                                                                                                                 
  - I have added tests to validate the effectiveness of my fix or the functionality of my new feature                                                                         
  - Both new and existing unit tests pass successfully on my local environment by running just test-coverage                                                                    
  - I have ensured that static analysis tests are passing by running just static-analysis

### DIFF
--- a/faststream/_internal/fastapi/router.py
+++ b/faststream/_internal/fastapi/router.py
@@ -37,7 +37,7 @@ from faststream._internal.types import (
     P_HandlerParams,
     T_HandlerReturn,
 )
-from faststream._internal.utils.functions import fake_context, to_async
+from faststream._internal.utils.functions import to_async
 from faststream.middlewares import BaseMiddleware
 from faststream.specification.asyncapi.site import get_asyncapi_html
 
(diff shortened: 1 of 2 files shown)


Lancetnik:
Hi @lesnik512, thanks a lot for the PR and for digging into the root cause — the analysis of how the inner router's ContextRepo gets out of sync is spot on.

That said, I don't think we should merge this as a fix, because the underlying scenario — including one StreamRouter (the FastAPI integration variant) into another via include_router — isn't actually a supported pattern. The docs on this are here: [Multiple FastStream Routers in a FastAPI application](https://faststream.ag2.ai/latest/getting-started/integrations/fastapi/#multiple-routers). The expected pattern is:

Use the FastAPI StreamRouter only at the top level, included into the FastAPI app.
For nesting / grouping subscribers, use the regular FastStream Router (e.g. KafkaRouter, etc.) and include those into the StreamRouter.

I do agree the current behavior is non-obvious, though — silently returning EmptyPlaceholder is a bad UX. I think the right fix here is to detect this case in StreamRouter.include_router and emit a clear warning (or even raise) pointing users to the docs, so they know to switch to a regular FastStream Router instead. Would you be interested in reshaping the PR in that direction? I'd be happy to review.
Thanks again for the contribution and sorry for not being able to take it as-is!

lesnik512:
@Lancetnik Hi) please check if it is ok

Lancetnik:
@lesnik512 yeah, smth like that. But I can't make a choice between warning / error. Do you have a position here?

lesnik512:
@Lancetnik 
Points for warning
- non-breaking change if someone already uses nested stream router

Points for exception
- not everyone pays attention to warnings
- code is simpler
- prevents misusage completely

I'd go for exception

Lancetnik (approved):

Lancetnik:
@lesnik512 your luck - 0.7.0 stable release is a good point to make such breaking changes😆 Another case I'll prefer warning

Issue #2837, opened 2026-04-14
Feature: unified codec configuration


ce1ebrimbor:
## Problem

Encoding and decoding use two different patterns in FastStream. Decoding is a function passed to the broker constructor (`decoder=`), called inside the handler DI chain with no middleware hook. Encoding is done through middleware via `publish_scope`. There is no `decode_scope` equivalent.

```python
broker = KafkaBroker(
    "localhost:9092",
    decoder=my_avro_decoder,                   # function, wired into DI
    middlewares=[MyAvroEncoderMiddleware(...)],  # middleware class, wired into publish_scope
)
```

This means a developer working with a binary format like Confluent Avro or Protobuf has to implement the same serialization concern in two different places, with two different lifecycles, often sharing state awkwardly between them.

The workaround today is to decode in `on_consume` middleware and set a passthrough `decoder=` to prevent the default JSON decoder from re-processing the already-decoded body. It works, but requires understanding FastStream internals.

## Proposal

Add a `codec=` parameter to the broker that handles both directions through a single object:

```python
broker = KafkaBroker(
    "localhost:9092",
    codec=ConfluentAvroCodec(schema_registry_url="http://localhost:8081"),
)
```

### Codec protocol

```python
@runtime_checkable
class Codec(Protocol):
    async def decode(self, msg: StreamMessage[Any]) -> DecodedMessage: ...
    async def encode(self, cmd: PublishCommand) -> PublishCommand: ...
```

Both methods are async to support codecs that fetch schemas or perform IO during serialization.

### Internal wiring

When `codec=` is provided, FastStream:

- Wires `codec.decode` as the decoder function, slotting into the existing `_get_parser_and_decoder` chain
- Auto-generates a middleware that calls `codec.encode` in `publish_scope`

No changes to the DI chain, middleware lifecycle, or parser composition. The codec is syntactic sugar over existing extension points.

### Deprecation of `decoder=`

The `codec=` parameter replaces `decoder=`. Once `codec` is available, `decoder=` should be deprecated with a warning pointing users to `codec=` instead. The `decoder=` parameter would remain functional for a deprecation cycle but be removed in a future major version.

If both `codec=` and `decoder=` are provided, raise `ValueError` — the two are mutually exclusive to avoid ambiguous decode paths.

## Example

```python
codec = ConfluentAvroCodec(
    schema_registry_url="http://localhost:8081",
    topics_versions={"orders.events": "latest"},
)
[...]

PR #2839, opened 2026-04-20
refactor: parser codec protocols

ce1ebrimbor:
# Parser & Codec Interface Unification

First part of #2837 

## Motivation

Every broker in FastStream has its own parser class (`AioKafkaParser`, `AsyncConfluentParser`, `AioPikaParser`, `NatsParser`, `SimpleParser`, etc.), but none of them share a formal interface. They all implement `parse_message` and `decode_message` by convention — duck-typed, with no Protocol or ABC enforcing the contract. Batch parsing is worse: Kafka overrides `parse_message` with an incompatible signature, Confluent calls its methods `parse_message_batch`/`decode_message_batch`, and NATS uses `parse_batch`/`decode_batch`.

Path extraction (`get_path`) is scattered across four brokers with four different signatures — Kafka takes a `str`, NATS takes a `str` but returns `None` on miss, Redis takes the full message mapping, and Rabbit does it inline. This inconsistency blocks defining a common parser interface.

This PR eliminates these inconsistencies to lay the groundwork for user-defined custom parser and encoder classes in future PRs.

### 1. `match_path()` utility function

**File**: `faststream/_internal/utils/path.py`

A new function that consolidates the duplicated path-extraction logic that existed across four broker parsers. Every broker's `get_path` method was doing the same thing: `pattern.match(subject).groupdict()`. Now they all call `match_path()` instead.

#### How it works

`match_path` is the runtime complement to the existing `compile_path` function. While `compile_path` runs at subscriber setup time to convert a pattern string like `"logs.{level}"` into a compiled regex with named capture groups, `match_path` runs at message time to apply that regex against the actual routing key/topic/subject of an incoming message.

```python
def match_path(pattern: Pattern[str] | None, subject: str) -> dict[str, Any]:
```

**Parameters**:
- `pattern` — A compiled regex produced by `compile_path` (or `compile_nats_wildcard` for NATS). Can be `None` when the subscriber's topic has no path variables (e.g., a plain topic like `"orders"` rather than `"orders.{region}"`).
- `subject` — The routing key from the incoming message. This is `message.topic` for Kafka, `message.subject` for NATS, `message.routing_key` for RabbitMQ, or `message["channel"]` for Redis.

**Return value**: Always returns `dict[str, Any]`. On a successful match, returns the named groups from the regex (e.g., `{"level": "error"}` for subject `"logs.error"` against pattern `"logs.{level}"`). On no match or when `pattern` is `None`, returns an empty dict `{}`.

**Why it never returns `None`**: The previous NATS implementation returned `None` on miss, while Kafka and Redis returned `{}`. Standardizing on `{}` means callers never need `path or {}` guards — the result goes directly into `StreamMessage.path`.

**Redis-specific note**: Redis pub/sub is unique because only pattern-subscribed messages carry a `"pattern"` field in the message dict. The guard for this (`message.get("pattern")`) stays at the Redis call site — `match_path` itself remains broker-agnostic.

### 2. Protocol definitions

**File**: `faststream/_internal/parser.py` (new)

Four `typing.Protocol` interfaces that formalize the parser/decoder contract:

| Protocol | Method | Purpose |
|---|---|---|
| `ParserProto[MsgType]` | `parse_message(message: MsgType) → StreamMessage[MsgType]` | Convert raw broker message to StreamMessage |
| `DecoderProto` | `decode_message(msg: StreamMessage[Any]) → DecodedMessage` | Deserialize message body to Python object |
| `BatchParserProto[MsgType]` | `parse_batch(messages: Sequence[MsgType]) → StreamMessage[Sequence[MsgType]]` | Convert batch of raw messages to StreamMessage |
| `BatchDecoderProto[MsgType]` | `decode_batch(msg: StreamMessage[Sequence[MsgType]]) → list[DecodedMessage]` | Deserialize batch message bodies |

These follow the existing `ProducerProto` pattern — `typing.Protocol` with `@abstractmethod`, generic over message type where applicable. They are not `@runtime_checkable`; they exist for static type checking.

Parser and decoder are separate protocols because they're already consumed as independent callables at the subscriber wiring level (`config.parser = self.parser.parse_message`, `config.decoder = self.parser.decode_message`). This separation reflects the existing architecture.

## What Did NOT Change

- `ParserComposition` in `faststream/_internal/endpoint/utils.py` — the runtime wiring that wraps custom user parsers with defaults. Untouched.
- `CustomCallable` type alias — the public API for `broker.subscriber(parser=...)`. Untouched.
- Shared `decode_message()` utility in `faststream/message/utils.py` — the content-type based JSON/text decoder. Untouched.
- MQTT parser — no path support, no batch methods. Untouched.
- `_setup(consumer)` on Kafka/Confluent parsers — lifecycle wiring. Untouched.

## Future Direction

This PR is the foundation for enabling **user-defined custom parser and encoder classes**. Currently, users can only pass callables via `broker.subscriber(parser=...)`. The next PRs will:

1. **Accept `ParserProto`/`DecoderProto` instances** as custom parsers — not just callables. `ParserComposition` will be updated to detect whether the custom parser is a callable or a Protocol-conformant object and wire it accordingly.

2. **Standardize the encoder side** — `encode_message` currently only exists as a static method on `AioPikaParser`. Future work will define an `EncoderProto` and give each broker a formal encoding interface, enabling custom serialization strategies (e.g., Protobuf, Avro, MessagePack) via class-based encoders.

3. **Unify batch parsing further** — with consistent `parse_batch`/`decode_batch` naming across all brokers, a single `BatchParserProto`-conformant class can be swapped in regardless of the broker.

The key insight is that parsers and decoders are already consumed as independent callables at the wiring level. The Protocols formalize what's already happening, making it possible to type-check custom implementations and provide clear extension points without changing the runtime architecture.

### DIFF
--- a/faststream/_internal/utils/path.py
+++ b/faststream/_internal/utils/path.py
@@ -51,9 +51,8 @@ def compile_path(
     return regex, original_path
 
 
-def match_path(pattern: "Pattern[str] | None", subject: str) -> "dict[str, Any]":
+def match_path(pattern: Pattern[str] | None, subject: str) -> dict[str, Any]:
     """Match subject against pattern and return named groups, or {} if no match."""
-    # Redis callers must check message.get("pattern") guard before calling this
     if pattern is not None and (match := pattern.match(subject)):
         return match.groupdict()
     return {}
--- a/faststream/_internal/utils/path.py
+++ b/faststream/_internal/utils/path.py
@@ -1,6 +1,7 @@
 import re
 from collections.abc import Callable
 from re import Pattern
+from typing import Any
 
 from faststream.exceptions import SetupError
 
@@ -48,3 +49,11 @@ def compile_path(
 
     original_path += path[idx:]
     return regex, original_path
+
+
+def match_path(pattern: "Pattern[str] | None", subject: str) -> "dict[str, Any]":
+    """Match subject against pattern and return named groups, or {} if no match."""
+    # Redis callers must check message.get("pattern") guard before calling this
+    if pattern is not None and (match := pattern.match(subject)):
+        return match.groupdict()
+    return {}
--- a/faststream/_internal/parser.py
+++ b/faststream/_internal/parser.py
@@ -1,0 +1,49 @@
+from abc import abstractmethod
+from collections.abc import Sequence
+from typing import TYPE_CHECKING, Any, Protocol, TypeVar
+
+if TYPE_CHECKING:
+    from faststream._internal.basic_types import DecodedMessage
+    from faststream.message import StreamMessage
+
+MsgType = TypeVar("MsgType")
+
+
+class ParserProto(Protocol[MsgType]):
+    """Protocol for parsing raw messages into StreamMessage."""
+
+    @abstractmethod
+    async def parse_message(self, message: MsgType) -> "StreamMessage[MsgType]":
+        """Parse a raw message into a StreamMessage."""
+        ...
+
+
+class DecoderProto(Protocol):
+    """Protocol for decoding StreamMessage into DecodedMessage."""
+
+    @abstractmethod
+    async def decode_message(self, msg: "StreamMessage[Any]") -> "DecodedMessage":
+        """Decode a StreamMessage into a DecodedMessage."""
+        ...
+
+
+class BatchParserProto(Protocol[MsgType]):
+    """Protocol for parsing a batch of raw messages into StreamMessage."""
+
+    @abstractmethod
+    async def parse_batch(
+        self, messages: Sequence[MsgType]
+    ) -> "StreamMessage[Sequence[MsgType]]":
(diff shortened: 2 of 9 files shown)


faststream/_internal/utils/path.py
```diff
@@ -48,3 +49,10 @@ def compile_path(
 
     original_path += path[idx:]
     return regex, original_path
+
+
+def match_path(pattern: Pattern[str] | None, subject: str) -> dict[str, Any]:
```
ce1ebrimbor, line 54: Wondering if this should be a Mixin instead.

Lancetnik (approved):

faststream/_internal/utils/path.py
Lancetnik: LGTM, we can change it later if it is required

PR #2844, opened 2026-04-26
Feat: Ability to configure topic params in confluent create_topics

### LINKED ISSUE #1827: Ability to configure topic params in confluent create_topics
**Current state:** every topic FastStream creates is created with `num_partitions=1, replication_factor=1`, hardcoded:

https://github.com/ag2ai/faststream/blob/2e36197300841f4cf1fc66537655ae2d0bb1bf10/faststream/confluent/helpers/admin.py#L52-L55

The only knob users have is the broker-wide `allow_auto_create_topics` flag — all topics of a broker are created, or none of them are. There is no way to configure a single topic, nor to opt a single topic out of creation.

**Request:** introduce a `Topic` schema object accepted by `subscriber()` / `publisher()` alongside plain strings, carrying per-topic settings:

```python
from faststream.confluent import KafkaBroker, Topic

broker = KafkaBroker()

@broker.subscriber(
    Topic("topic-name", num_partitions=3),
    Topic("topic-name2", num_partitions=1, replication_factor=2),
    Topic("externally-managed", declare=False),
    "topic-without-settings",
)
```

**Scope**

1. **`num_partitions` / `replication_factor` per topic** — the original ask, from https://github.com/airtai/faststream/discussions/1821. Today both are pinned to `1`, which makes FastStream-created topics unusable in any real deployment.

2. **`declare: bool = True`** — opt a single topic out of creation while auto-creation stays the default for everything else. This subsumes #2679: the global default is **not** changing, because a broker-wide switch is the wrong granularity — a service typically owns some of its topics and consumes others that are provisioned by a different team or by IaC.

**Proposed semantics of `declare=False`:** skip the `create_topics` call for that topic and nothing else — do not probe the cluster for existence, do not fail if the topic is missing. This matches what `allow_auto_create_topics=False` does today (a warning, then let the consumer proceed) and matches NATS' `JStream(declare=False)`. Note this deliberately differs from RabbitMQ's `RabbitQueue(declare=False)`, which maps to AMQP `passive=True` and *does* raise when the queue is absent — Kafka has no cheap equivalent of a passive declare.

**Interaction with the broker-level flag:** `allow_auto_create_topics=False` on the broker keeps winning over everything — it stays the "create nothing at all" switch. `declare` only narrows creation further when the broker-level flag is on.

**Consistency across brokers:** `declare` is the established name for this in FastStream — [`RabbitQueue(declare=...)`](https://github.com/ag2ai/faststream/blob/2e36197300841f4cf1fc66537655ae2d0bb1bf10/faststream/rabbit/schemas/queue.py#L108), `JStream(declare=...)`, `KvWatch(declare=...)`, `ObjWatch(declare=...)`. `Topic` should use the same name rather than inventing a Kafka-specific one.

**AioKafka is out of scope.** FastStream never creates topics for `faststream.kafka` — there is no `AdminClient.create_topics` call on that path, and `aiokafka` (0.13.0) does not support an `allow_auto_create_topics` consumer option at all:

```python
$ grep -rn "auto_create\|auto\.create" .venv/lib/python3.11/site-packages/aiokafka/
$ grep -rn "auto_create" faststream/kafka/
```

Both return nothing. Topic creation on that path is entirely the Kafka server's `auto.create.topics.enable`, which FastStream cannot influence. `Topic` may still be accepted there later for symmetry, but `declare` would be a no-op, so it should not block this issue.

**Implementation notes**

- `create_subscriber()` / `create_publisher()` signatures must accept `str | Topic` — `*topics: str` is what currently trips mypy on the branch in progress.
- Normalise `str` → `Topic(name)` at registration time, as RabbitMQ does with `str` → `RabbitQueue`.
- `AsyncConfluentConsumer.topics_to_create` should filter on `declare` — https://github.com/ag2ai/faststream/blob/2e36197300841f4cf1fc66537655ae2d0bb1bf10/faststream/confluent/helpers/client.py#L292-L294
- `Topic` needs `__hash__` / `__eq__` consistent with each other, since topics end up as dict keys (see #2796 for the RabbitMQ precedent).

Related: #2679 (closed in favour of this), #1486, #1658, #2451.

Vijay-K-2003:
# Description
In context of Confluent create_topics function:
Instead of using default values for num_partitions and replication_factor as 1. We want to give an ability to make it configurable, like 

```python
@broker.subscriber(
    Topic("topic-name", num_partitions=3),
    Topic("topic-name2", num_partitions=1),
    "topic-without-settings"
)
```

So to do this, I created a Topic Schema which can be initialized with parameters like name, num_partitions and replication_factor (I am not sure what other parameters needs to be configurable.) . Then, we need the subscriber to allow parameter as "Topic" so I changed the topics: Sequence[str] to Sequence[str, "Topic"] in the subscriber/config.py .
The LogicSubscriber property topics() handles the str input ,, so i added an if statement to handle the Topic input to function as

```@property
    def topics(self) -> list[Any]:
        res: list[Any] = []
        for t in self._topics:
            if isinstance(t, Topic):
                res.append(t.add_prefix(self._outer_config.prefix))
            else:
                res.append(f"{self._outer_config.prefix}{t}")
        return res
```
then similarly for AdminService and AsyncConfluentConsumer create_topics function.

Then the Registrator also requires to handle input of type string and Topic so changed all topics: str to topics: Union[str, "Topic"].

Then added a new test that creates new topics and checks if the configuration is correct.

The just static-analysis gave me an error that says that create_subscriber cannot have Topic as arg type.
```python
just static-analysis                      
just _static mypy 
uv run --frozen mypy
faststream/confluent/broker/registrator.py:405: error: Argument 1 to "create_subscriber" has incompatible type "*tuple[str | Topic, ...]"; expected "str"  [arg-type]
Found 1 error in 1 file (checked 502 source files)
error: Recipe `_static` failed on line 115 with exit code 1
error: Recipe `mypy` failed on line 120 with exit code 1
```

And the test-coverage is 16 percent total.

Fixes #1827 

## Type of change

## Checklist

### DIFF
--- a/faststream/confluent/__init__.py
+++ b/faststream/confluent/__init__.py
@@ -4,7 +4,7 @@ try:
     from .annotations import KafkaMessage
     from .broker import KafkaBroker, KafkaPublisher, KafkaRoute, KafkaRouter
     from .response import KafkaPublishCommand, KafkaPublishMessage, KafkaResponse
-    from .schemas import TopicPartition
+    from .schemas import Topic, TopicPartition
     from .testing import TestKafkaBroker
 
 except ImportError as e:
@@ -26,5 +26,6 @@ __all__ = (
     "KafkaRouter",
     "TestApp",
     "TestKafkaBroker",
+    "Topic",
     "TopicPartition",
 )
--- a/faststream/confluent/broker/registrator.py
+++ b/faststream/confluent/broker/registrator.py
@@ -31,7 +31,7 @@ if TYPE_CHECKING:
         BatchPublisher,
         DefaultPublisher,
     )
-    from faststream.confluent.schemas import TopicPartition
+    from faststream.confluent.schemas import Topic, TopicPartition
     from faststream.confluent.subscriber.usecase import (
         BatchSubscriber,
         ConcurrentDefaultSubscriber,
@@ -47,7 +47,7 @@ class KafkaRegistrator(
     @overload  # type: ignore[override]
     def subscriber(
         self,
-        *topics: str,
+        *topics: Union[str, "Topic"],
         partitions: Sequence["TopicPartition"] = (),
         polling_interval: float = 0.1,
         group_id: str | None = None,
@@ -90,7 +90,7 @@ class KafkaRegistrator(
     @overload
     def subscriber(
         self,
-        *topics: str,
+        *topics: Union[str, "Topic"],
         partitions: Sequence["TopicPartition"] = (),
         polling_interval: float = 0.1,
         group_id: str | None = None,
@@ -133,7 +133,7 @@ class KafkaRegistrator(
     @overload
     def subscriber(
         self,
-        *topics: str,
+        *topics: Union[str, "Topic"],
         partitions: Sequence["TopicPartition"] = (),
         polling_interval: float = 0.1,
         group_id: str | None = None,
@@ -176,7 +176,7 @@ class KafkaRegistrator(
     @overload
     def subscriber(
         self,
-        *topics: str,
+        *topics: Union[str, "Topic"],
         partitions: Sequence["TopicPartition"] = (),
         polling_interval: float = 0.1,
         group_id: str | None = None,
@@ -223,7 +223,7 @@ class KafkaRegistrator(
     @override
     def subscriber(
         self,
-        *topics: str,
+        *topics: Union[str, "Topic"],
         partitions: Sequence["TopicPartition"] = (),
         polling_interval: float = 0.1,
         group_id: str | None = None,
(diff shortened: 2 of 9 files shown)


Lancetnik:
@Vijay-K-2003 welcome and thank you for the PR! It looks good, but I have just a little comment: can avoid passing `str | Topic` deep inside the framework? Such unions is really hard to maintain in the future. I suggest to map any `str | Topic` to `Topic` type at publisher & subscriber factory level?

I mean `create_publisher` and `create_subscriber` functions should consume `str | Topic` and pass `Topic` object to all other inner structure. Just to make inner logic linear without additional runtime conditions.

Also, do we create `@broker.publisher("topic")` publisher-declared topics? If so, we should support the same behavior here

PR #2859, opened 2026-05-07
Fix logger not passed to Confluent Producer and AdminClient

### LINKED ISSUE #2691: Bug: Logger not properly passed to Confluent Kafka Producer and AdminClient
**Describe the bug**

There are two related issues with logger configuration in the Confluent Kafka components:

1. Late Logger Setup in AsyncConfluentProducer

      In faststream/confluent/helpers/client.py at line 46, the Producer object is created before the _setup() method is called for loger_state. This results in the Producer receiving a NoSetLoggerObject instead of the intended logger passed during initialization.

2. Missing Logger in AdminClient
     The AdminClient in AdminService does not accept a logger parameter, even though it should. Currently, only the configuration is passed, leaving the AdminClient without proper logging.

**How to reproduce**
Include source code:

```python
import logging

import uvicorn
from fastapi import FastAPI
from faststream.confluent.fastapi import KafkaRouter

logger = logging.getLogger("faststream")

router = KafkaRouter(
    bootstrap_servers="kafka:9092",
    enable_idempotence=True,
    allow_auto_create_topics=False,
    schema_url="/asyncapi",
    include_in_schema=True,
    logger=logger,
)

app = FastAPI()
app.include_router(router)

uvicorn.run(app, host="0.0.0.0", port=8000)
```

**Expected behavior**

1. The Producer in AsyncConfluentProducer should use the logger passed to KafkaRouter.
2. The AdminClient should inherit the same logger as other components.

**Observed behavior**

1. The Producer uses NoSetLoggerObject instead of the provided logger.
2. The AdminClient lacks logger configuration, leading to potential silent failures or inadequate logging.

**Environment**
Running FastStream 0.6.3 with CPython 3.13.1 on Linux

mara-werils:
Fixes #2691

## Problem

1. **Producer receives placeholder logger**: `_setup_logger()` was called *after* `_connect()` in `BrokerUsecase.connect()`, so when `AsyncConfluentProducer` was created during `_connect()`, `logger_state` was not yet initialized — the Producer received a `NoSetLoggerObject` instead of the real logger.

2. **AdminClient has no logger**: `AdminService.connect()` did not accept or pass a `logger` parameter to the `AdminClient` constructor.

## Fix

- **Swap call order** in `BrokerUsecase.connect()`: call `_setup_logger()` before `_connect()` so the logger state is initialized when clients are created.
- **Pass logger to AdminClient**: add an optional `logger` parameter to `AdminService.connect()` and wire it from `KafkaBrokerConfig.connect()`.

## Changes

- `faststream/_internal/broker/broker.py` — call `_setup_logger()` before `_connect()`
- `faststream/confluent/helpers/admin.py` — accept and pass `logger` to `AdminClient`
- `faststream/confluent/configs/broker.py` — pass logger when connecting admin service

### DIFF
--- a/faststream/_internal/broker/broker.py
+++ b/faststream/_internal/broker/broker.py
@@ -108,8 +108,8 @@ class BrokerUsecase(
     async def connect(self) -> ConnectionType:
         """Connect to a remote server."""
         if self._connection is None:
-            self._connection = await self._connect()
             self._setup_logger()
+            self._connection = await self._connect()
 
         return self._connection
 
--- a/faststream/_internal/broker/broker.py
+++ b/faststream/_internal/broker/broker.py
@@ -108,8 +108,8 @@ class BrokerUsecase(
     async def connect(self) -> ConnectionType:
         """Connect to a remote server."""
         if self._connection is None:
-            self._setup_logger()
             self._connection = await self._connect()
+            self._setup_logger()
 
         return self._connection
 
--- a/faststream/confluent/helpers/admin.py
+++ b/faststream/confluent/helpers/admin.py
@@ -1,5 +1,3 @@
-import logging
-
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
@@ -9,6 +7,8 @@ from confluent_kafka.admin import (  # type: ignore[attr-defined]
 )
 
 if TYPE_CHECKING:
+    from faststream._internal.logger import LoggerState
+
     from .config import ConfluentFastConfig
 
 
@@ -25,12 +25,16 @@ class AdminService:
     async def connect(
         self,
         config: "ConfluentFastConfig",
-        logger: logging.Logger | None = None,
+        logger: "LoggerState | None" = None,
     ) -> None:
         if self.admin_client is None:
+            from .client import _LazyLoggerProxy
+
             admin_config = config.admin_config
             if logger is not None:
-                self.admin_client = AdminClient(admin_config, logger=logger)
+                self.admin_client = AdminClient(
+                    admin_config, logger=_LazyLoggerProxy(logger),
+                )
             else:
                 self.admin_client = AdminClient(admin_config)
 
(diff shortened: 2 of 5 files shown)


faststream/_internal/broker/broker.py
```diff
@@ -108,8 +108,8 @@ def _setup_logger(self) -> None:
     async def connect(self) -> ConnectionType:
         """Connect to a remote server."""
         if self._connection is None:
-            self._connection = await self._connect()
             self._setup_logger()
+            self._connection = await self._connect()
```
Lancetnik, line 111: We can't change connection order here

mara-werils: Thanks for pointing this out! I've reworked the approach, instead of changing the connection order, I introduced a `_LazyLoggerProxy` that wraps `LoggerState` and is passed to `Producer`, `Consumer`, and `AdminClient` at construction time. It silently drops log records until `_setup_logger()` initializes the real logger, then forwards everything through. No changes to `broker.py` anymore.

faststream/confluent/helpers/admin.py
```diff
@@ -20,9 +22,21 @@ class AdminService:
...
 
-    async def connect(self, config: "ConfluentFastConfig") -> None:
+    async def connect(
+        self,
+        config: "ConfluentFastConfig",
+        logger: "LoggerState | None" = None,
+    ) -> None:
         if self.admin_client is None:
-            self.admin_client = AdminClient(config.admin_config)
+            from .client import _LazyLoggerProxy
```
Lancetnik: Please, avoid function-level imports

Lancetnik (changes requested):
Great decision! Just a single comment about code
And I need any test to verify, that LoggerProxy correctly uses real one after bootstrapping

mara-werils:
@Lancetnik Addressed both comments: moved the import to module level and added tests for `_LazyLoggerProxy` (verifies it drops records before setup and forwards after bootstrapping). Ready for re-review!

Lancetnik (approved):
Thank you!

PR #2870, opened 2026-05-15
feat: expose client_rack option for Kafka broker consumer

### LINKED ISSUE #2868: Bug: aiokafka broker doesn't expose `client_rack` option
**Describe the bug**
Can't forward `client_rack` to `aiokafka` consumer through FastStream Kafka API.

**How to reproduce**
Include source code:

```python

# Attempt to configure at broker level

broker = KafkaBroker(
    "localhost:9092",
    client_rack="us-east-1a",  # not supported
)
app = FastStream(broker)
```

And/Or steps to reproduce the behaviour:

1. Create a KafkaBroker.
2. Try to pass client_rack to the broker constructor.
3. Observe that the parameter is not accepted.
4. There is no alternative way to configure it via subscriber(...) either.

**Expected behavior**
KafkaBroker should allow passing client_rack (or generic consumer kwargs) and forward it to the underlying aiokafka.AIOKafkaConsumer.

**Observed behavior**
client_rack is not supported in KafkaBroker constructor, nor in subscriber configuration, so it cannot be passed to aiokafka.

**Screenshots**
N/A.

**Environment**
Running FastStream 0.6.7 with CPython 3.14.5 on Linux

**Additional context**
aiokafka.AIOKafkaConsumer supports client_rack since 0.14.0 version

algojogacor:
## Summary

Expose `client_rack` option in `KafkaBroker` so users can configure rack-aware replica selection (KIP-392) when creating Kafka consumers.

## Root Cause

The `client_rack` parameter, supported by `aiokafka.AIOKafkaConsumer` since v0.14.0, was not exposed through FastStream's `KafkaBroker`. The parameter was missing from:
- `ConsumerConnectionParams` TypedDict (filtered out by `filter_by_dict`)
- `KafkaBroker.__init__` signature
- `KafkaInitKwargs` TypedDict

Any parameter not listed in `ConsumerConnectionParams` is silently dropped during the `filter_by_dict` call in the broker constructor, preventing it from reaching `aiokafka.AIOKafkaConsumer`.

## Solution

Added `client_rack: str | None = None` to:
1. `ConsumerConnectionParams` in `faststream/kafka/schemas/params.py` — so it passes through `filter_by_dict`
2. `KafkaBroker.__init__` — new explicit parameter
3. `KafkaInitKwargs` TypedDict — for type completeness
4. `connection_params` dict — wired through so it reaches the consumer builder

Also added `group_instance_id` to `KafkaInitKwargs` fields, as it was documented in the docstring but missing from the actual TypedDict definition.

## Testing

Syntax verified with `ast.parse()`. The change follows the existing pattern used by other consumer parameters (e.g., `group_instance_id`, `client_id`). Full test suite requires a running Kafka instance; the change is purely additive (default `None` preserves existing behavior).

Closes #2868

### DIFF
--- a/faststream/kafka/broker/broker.py
+++ b/faststream/kafka/broker/broker.py
@@ -67,6 +67,9 @@ if TYPE_CHECKING:
         """Kafka broker initialization keyword arguments.
 
         Attributes:
+            group_instance_id: Name of the group instance ID used for static membership (KIP-345).
+            client_rack: Rack identifier for this consumer (KIP-392). Used to fetch
+                from the closest replicas when rack-aware replica selection is enabled.
             request_timeout_ms: Client request timeout in milliseconds.
             retry_backoff_ms: Milliseconds to backoff when retrying on errors.
             metadata_max_age_ms: The period of time in milliseconds after
@@ -157,6 +160,8 @@ if TYPE_CHECKING:
         sasl_oauth_token_provider: AbstractTokenProvider | None
         loop: asyncio.AbstractEventLoop | None
         client_id: str | None
+        group_instance_id: str | None
+        client_rack: str | None
         # publisher args
         acks: Literal[0, 1, -1, "all"] | object
         key_serializer: Callable[[Any], bytes] | None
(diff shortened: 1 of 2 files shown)


Lancetnik:
Sorry, #2871 already merged

PR #2871, opened 2026-05-16
Expose client_rack option on the Kafka broker

### LINKED ISSUE #2868: Bug: aiokafka broker doesn't expose `client_rack` option

00yhj22-debug:
# Description

`KafkaBroker` did not accept the `client_rack` consumer option, so there was no way to forward it to the underlying `aiokafka.AIOKafkaConsumer` — either at the broker level or via `subscriber(...)`.

This PR adds a `client_rack` argument to `KafkaBroker` (and `KafkaBrokerConfig`) and forwards it to the consumer builder.

**Design note** — `client_rack` is a *consumer-only* option. `AIOKafkaProducer` does not accept it, and in `KafkaBrokerConfig.connect()` the broker's `connection_kwargs` are passed straight to `AIOKafkaProducer(**connection_kwargs)`. So `client_rack` is deliberately **kept out of `connection_params`** and injected into the consumer options separately, and **only when a value is provided**. When `client_rack` is left as the default (`None`) nothing is passed to the consumer, so existing behavior is completely unchanged.

Fixes #2868

## Note on the aiokafka version

`client_rack` exists on `AIOKafkaConsumer` since **aiokafka 0.14.0**, but the `kafka` extra currently pins `aiokafka>=0.9,<0.14`. This PR does **not** change that pin, because:

- The new option is dormant unless the user explicitly sets it, so the change is safe on the currently pinned aiokafka range.
- Bumping the aiokafka upper bound is a separate concern that deserves its own review.

Happy to bump the pin here as well if you'd prefer `client_rack` to be usable out of the box — just let me know which you'd like.

## Type of change

## Checklist

- [x] My code adheres to the style guidelines of this project (`ruff check` and `ruff format` report no errors)
- [x] I have made the necessary changes to the documentation (the new argument is documented in the `KafkaBroker` docstring and `KafkaInitKwargs`)
- [x] I have added tests to validate the effectiveness of my fix or the functionality of my new feature (`tests/brokers/kafka/test_client_rack.py`)

The new tests pass, and the kafka unit-test suite (`pytest tests/brokers/kafka -m "not connected and not slow"`) passes locally — 133 passed. I could not run the full `just test-coverage` / connected suites locally, so I am relying on CI for those.

### DIFF
--- a/faststream/kafka/broker/broker.py
+++ b/faststream/kafka/broker/broker.py
@@ -85,6 +85,9 @@ if TYPE_CHECKING:
                 server-side log entries that correspond to this client. Also
                 submitted to :class:`~.consumer.group_coordinator.GroupCoordinator`
                 for logging with respect to consumer group administration.
+            client_rack: A rack identifier for this client. Used by the
+                broker for rack-aware fetching, letting a consumer read from
+                the closest replica. Requires aiokafka 0.14.0 or newer.
             acks: One of ``0``, ``1``, ``all``. The number of acknowledgments
                 the producer requires the leader to have received before considering a
                 request complete. This controls the durability of records that are
@@ -157,6 +160,8 @@ if TYPE_CHECKING:
         sasl_oauth_token_provider: AbstractTokenProvider | None
         loop: asyncio.AbstractEventLoop | None
         client_id: str | None
+        # consumer args
+        client_rack: str | None
         # publisher args
         acks: Literal[0, 1, -1, "all"] | object
         key_serializer: Callable[[Any], bytes] | None
@@ -197,6 +202,8 @@ class KafkaBroker(
         sasl_oauth_token_provider: Optional["AbstractTokenProvider"] = None,
         loop: Optional["asyncio.AbstractEventLoop"] = None,
         client_id: str | None = SERVICE_NAME,
+        # consumer args
+        client_rack: str | None = None,
         # publisher args
         acks: Literal[0, 1, -1, "all"] | object = _missing,
         key_serializer: Callable[[Any], bytes] | None = None,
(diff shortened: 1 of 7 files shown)


Lancetnik:
@00yhj22-debug hi! Thank you for the PR!
I think we should support `broker.subscriber(client_rack=...)` options as well. It is commona pattern in FS: broker option using as a default, endpoint variable overrides it

00yhj22-debug:
Thanks for the review, @Lancetnik!

I've added subscriber-level support in 608c6df: `client_rack` can now be passed to `broker.subscriber(...)` and `KafkaRoute`, and the subscriber value overrides the broker-level default.

It is forwarded through the subscriber `connection_args` only when explicitly set, so an unset subscriber value leaves the broker-level `client_rack` in place, while a set value takes precedence when the consumer is built. Tests for the broker/subscriber/override cases are in `tests/brokers/kafka/test_client_rack.py`.

I scoped it to `broker.subscriber()` and `KafkaRoute` for now (mirroring `group_instance_id`). Happy to extend it to the FastAPI `KafkaRouter` as well if you would like — just let me know.

Lancetnik:
@00yhj22-debug 
> Happy to extend it to the FastAPI KafkaRouter as well if you would like — just let me know.
Great catch, we need to provide same experience all cases

00yhj22-debug:
Done in c482020 — `client_rack` is now supported on the FastAPI `KafkaRouter` as well, forwarded to the underlying subscriber so it behaves the same as `broker.subscriber()` and `KafkaRoute`. Added a test for it too.

One thing I noticed while wiring this up: the FastAPI router's `subscriber()` accepts `group_instance_id` in its signature but does not forward it in the `super().subscriber()` call, so it currently has no effect there. I kept that out of this PR to stay focused on `client_rack`, but I am happy to send a small follow-up fix for it if that would be useful.

Lancetnik (approved):

PR #2874, opened 2026-05-17
fix(cli): surface YAML and AsyncAPI errors from `docs serve`

SarthakB11:
# Description

Refs #2709 (diagnostic portion).

Per @Sehat1137's recommendation on #2858 ("a smaller PR limited to that behavior, with deterministic parsing and regression tests"), this is the diagnostic-only redo: no new CLI options, no parser plugin, default `yaml.safe_load` preserved. The pluggable-parser proposal from #2709 is intentionally deferred per maintainer feedback.

`faststream docs serve <asyncapi.yaml>` previously collapsed three different failure modes into one opaque `SCHEMA_NOT_SUPPORTED` message:

- YAML parse errors were not surfaced. `yaml.safe_load` ran outside any `try`, so the user either got an uncaught traceback or, more often, a downstream pydantic failure on the garbage that partial parsing produced.
- AsyncAPI validation errors were dropped by `contextlib.suppress(ValidationError)`, so there was no indication of which schema version was tried or why it failed.
- YAML scalar gotchas like `protocolVersion: 3.2` (parsed as a `float`, rejected by AsyncAPI's `string` type) gave no hint that the value just needed quoting. This is the exact repro from #2709.

## Fix

Edits only `_parse_and_serve` in `faststream/_internal/cli/docs.py`:

1. Wrap `yaml.safe_load` in `try / except yaml.YAMLError`, echo the parser exception, exit 1.
2. Replace `suppress(ValidationError)` with an explicit collector that records `(version_label, ValidationError)`. When neither v3.0 nor v2.6 validates, print `SCHEMA_NOT_SUPPORTED` followed by both per-version error blocks on stderr.
3. For each block, if any sub-error has `type == "string_type"` and an `int | float | bool` input, append a one-line hint pointing at the location and suggesting YAML quoting (e.g. `'3.2'`).

Default parser is unchanged. `_parse_and_serve`'s public signature is untouched. `dto.py` is untouched. No new modules, no parser registry, no `--yaml-parser` flag.

## Tests

Two regression tests in `tests/cli/test_asyncapi_docs.py`:

- `test_serve_asyncapi_yaml_unquoted_scalar_reports_hint` mutates the existing `yaml_asyncapi_doc` fixture to `protocolVersion: 3.2`, asserts exit code 1 and the v3.0 / v2.6 error enumeration plus the quoting hint.
- `test_serve_asyncapi_reports_yaml_parse_error` feeds malformed YAML (`foo: [unclosed`) and asserts the parser exception is surfaced rather than swallowed.

Both follow the existing `faststream_cli + generate_template + wait_for_stderr` pattern in the same file.

## Type of change

## Checklist


Sehat1137:
Sorry for the long wait on this PR, and thank you for your patience.

The overall direction looks good, and the improved validation diagnostics are useful. Before we merge it, could you please make the following adjustments:

1. Show the YAML-specific quoting hint only for `.yaml` and `.yml` inputs. Currently, a JSON document containing a numeric value may also receive a misleading YAML hint.
2. Add a regression test confirming that JSON inputs do not produce the YAML-specific hint.
3. Emit the quoting hint only once, rather than potentially once per AsyncAPI version.
4. Please add a test covering the exact reproduction from #2709, including `protocolVersion: ! "3.2"`.

The YAML parsing error handling and surfaced Pydantic validation errors are otherwise a useful improvement. Thanks for working on this.

SarthakB11:
All four are addressed: the hint now fires only for .yaml/.yml and prints once, a JSON test confirms it stays absent, and a new test covers the `protocolVersion: ! "3.2"` case from #2709.

Sehat1137 (approved):

Sehat1137:
Can you resolve problem with tests?

SarthakB11:
Looked into the 3 failures: test_worker_id_extra_option.py, test_run_asgi.py::test_many_workers, test_logs.py::test_run_as_asgi_mp_with_log_level. All three are in the multi-worker CLI subprocess tests, a different subsystem from this PR's diff (faststream/_internal/cli/docs.py and its own test file only). All 6 tests in that area pass locally for me on this branch. Rebased onto latest main and pushed to get a clean CI run, that should confirm it either way.

Sehat1137:
@SarthakB1 problem with the tests is relevant, try to rebase ur branch

Sehat1137 (approved):

PR #2878, opened 2026-05-19
fix: TestRedisBroker delivers freshly-published messages to subscribers with min_idle_time

### LINKED ISSUE #2848: Bug: TestRedisBroker delivers freshly-published messages to subscribers with min_idle_time
**Describe the bug**

In `TestRedisBroker`, when a stream has two stacked subscribers — one without `min_idle_time` (which in production uses XREADGROUP) and one with `min_idle_time` (which in production uses XAUTOCLAIM) — every published message is delivered to both. In production these subscribers handle disjoint message sets via different Redis commands, but the test broker treats them identically and the handler runs twice per publish.

**How to reproduce**

```python
import asyncio
from faststream.redis import RedisBroker, StreamSub, TestRedisBroker

broker = RedisBroker()
calls = 0

@broker.subscriber(stream=StreamSub('events', group='g', consumer='c'))
@broker.subscriber(stream=StreamSub('events', group='g', consumer='c', min_idle_time=60_000))
async def handle(msg) -> None:
    global calls
    calls += 1

async def main() -> None:
    async with TestRedisBroker(broker):
        await broker.publish('hello', stream='events')
    print(f'calls = {calls}')

asyncio.run(main())
```

**Expected behavior**

`calls == 1` — only the XREADGROUP-style subscriber should receive the freshly-published message. The XAUTOCLAIM-style subscriber would, in a real Redis, only see the message after it had been pending and idle long enough to be reclaimed.

**Observed behavior**

`calls == 2` — both subscribers fire on the publish.

The matching logic in `StreamVisitor.visit` (in `faststream/redis/testing.py`) compares `stream` against `sub.stream_sub.name` only, never inspecting `sub.stream_sub.min_idle_time`, so any subscriber sharing the stream name is treated as eligible.

**Screenshots**

N/A.

**Environment**

- FastStream `0.6.7`
- Python `3.14.3`
- macOS (Darwin 25.3.0)
- Backend: redis

**Additional context**

This pattern (two stacked subscribers, second one with `min_idle_time`, on the same stream/group) is what's needed to combine new-message processing with idle-message reclaim — the single-subscriber Pattern 2 in [docs/redis/streams/claiming](https://faststream.ag2.ai/latest/redis/streams/claiming/) doesn't work in production because `min_idle_time` switches the read command from XREADGROUP to XAUTOCLAIM, so such a subscriber never receives new messages at all (maybe a separate doc-correctness issue?).

Workaround used: monkey-patch `StreamVisitor.visit` in test setup to skip subscribers with `min_idle_time`:

```python
from faststream.redis.testing import StreamVisitor

_original = StreamVisitor.visit

def _visit(self, *, sub, **kw):
    if getattr(sub, 'min_idle_time', None) is not None:
        return None
    return _original(self, sub=sub, **kw)

StreamVisitor.visit = _visit
```

ApusBerliozi:
# Description

So, in order to "emulate" xautoclaim, I decided to add `asyncio.wait_for` in `FakeProducer.publish`, with timeout of 30 seconds (cause it's the maximum for pytest). I think it's a reasonable compromise, cause it will simulate how production environment would handle this situation.  So, if `min_idle_time` is less than 30 seconds we will proceed with the call after some delay and if it isn't - we won't.

Fixes #2848
## Type of change

## Checklist

- [X] I have ensured that static analysis tests are passing by running `just static-analysis`
- [X] I have included code examples to illustrate the modifications


ApusBerliozi:
Hello @Lancetnik!

Can you please take a look at this PR?

Lancetnik:
@ApusBerliozi sorry, I undestand nothing in Redis Streams
We need @powersemmi help to review any Redis PRs

ApusBerliozi:
Oh I see, no worries @Lancetnik, thx 4 the info

powersemmi:
@ApusBerliozi 

I've created an issue that should solve this problem at the root: waiting in tests isn't great, it would be simpler to emulate the PEL instead. That way it also becomes possible to test autoclaim separately from the regular handler.

See #2927

PR #2882, opened 2026-05-21
Forward group_instance_id from the FastAPI Kafka router

00yhj22-debug:
# Description

Follow-up to #2871. The FastAPI `KafkaRouter` accepts `group_instance_id` in its `subscriber()` overload signatures, but the parameter was not included in the `super().subscriber()` call in `faststream/kafka/fastapi/fastapi.py`. As a result the value was silently dropped before reaching the underlying broker subscriber and the consumer never saw it — calling `router.subscriber("topic", group_instance_id="x")` on the FastAPI router had no effect.

This PR forwards `group_instance_id` explicitly in the `super()` call. The fix is a single line. I also added a fastapi-flavored regression test alongside the existing broker and `KafkaRouter` cases in `tests/brokers/kafka/test_group_instance_id.py`.

## Type of change

## Checklist

- [x] My code adheres to the style guidelines (`ruff check` and `ruff format` are clean)
- [x] I have made the necessary changes to the documentation (no docs change needed — the parameter is already documented in the FastAPI router`s docstring; it just was not forwarded)
- [x] Both new and existing unit tests pass successfully on my local environment (`pytest tests/brokers/kafka/test_group_instance_id.py` — all five tests pass)


Lancetnik (dismissed):

00yhj22-debug:
Hi @Lancetnik — gentle nudge on this one. It's been approved and CI is green, just wanted to check there's nothing else you need from me before it can go in. Happy to rebase if main has drifted since. Thanks!

Lancetnik (approved):

PR #2883, opened 2026-05-21
Add consumer_only flag to KafkaBroker

### LINKED ISSUE #2879: Feature: Consumer-Only Mode for KafkaBroker
When deploying a subscriber-only service, it's common to provision Kafka credentials scoped strictly to `READ` + `DESCRIBE` ACLs on specific topics. Today, `KafkaBroker` always starts a producer during `broker.start()`, regardless of whether any `@broker.publisher` is registered. This causes the producer to attempt a connection using consumer-only credentials, which fails or raises errors at the broker/ACL level.

Proposal:
```python
# Option A: broker-level flag
broker = KafkaBroker("localhost:9092", consumer_only=True)

# Option B: skip producer if no publishers are registered (automatic)
# No API change needed — broker inspects registered publishers at start()
```

00yhj22-debug:
# Description

Closes #2879.

When deploying a subscriber-only Kafka service it is common to scope credentials to READ + DESCRIBE ACLs on specific topics. Today `KafkaBroker` always creates a producer and admin client during `connect()`, which fails or raises against such credentials even when no publishers are registered.

Per @Lancetnik`s direction on #2879 ("We just need to disable the creation of producer & admin clients when the flag is set"), this PR adds a `consumer_only` flag to `KafkaBroker` and `KafkaBrokerConfig`:

- When `consumer_only=True`, `KafkaBrokerConfig.connect()` skips creating the `AIOKafkaProducer` and `AIOKafkaAdminClient`, and `disconnect()` leaves the producer alone.
- The consumer builder is still wired up so subscribers continue to work normally.
- Default is `False`, so existing behavior is completely unchanged.

## Type of change

## Checklist

- [x] My code adheres to the style guidelines (`ruff check` and `ruff format` are clean)
- [x] I have made the necessary changes to the documentation (the new argument is documented in both the `KafkaBroker` docstring and `KafkaInitKwargs`)
- [x] I have added tests to validate the effectiveness of my fix (`tests/brokers/kafka/test_consumer_only.py`, 7 cases including mocked `connect()`/`disconnect()` to assert producer/admin clients are not created when the flag is set, and *are* created on the default path)
- [x] Both new and existing unit tests pass successfully on my local environment (`pytest tests/brokers/kafka -m "not connected and not slow"` — 171 passed)

### DIFF
--- a/faststream/kafka/broker/broker.py
+++ b/faststream/kafka/broker/broker.py
@@ -88,6 +88,9 @@ if TYPE_CHECKING:
             client_rack: A rack identifier for this client. Used by the
                 broker for rack-aware fetching, letting a consumer read from
                 the closest replica. Requires aiokafka 0.14.0 or newer.
+            consumer_only: When True the broker skips creating the
+                producer and admin clients, letting deployments use Kafka
+                credentials scoped to read-only ACLs. Defaults to False.
             acks: One of ``0``, ``1``, ``all``. The number of acknowledgments
                 the producer requires the leader to have received before considering a
                 request complete. This controls the durability of records that are
@@ -162,6 +165,7 @@ if TYPE_CHECKING:
         client_id: str | None
         # consumer args
         client_rack: str | None
+        consumer_only: bool
         # publisher args
         acks: Literal[0, 1, -1, "all"] | object
         key_serializer: Callable[[Any], bytes] | None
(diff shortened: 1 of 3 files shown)


dzeveloper (approved):
Thank you, that was super fast 👍

Lancetnik (approved):

PR #2885, opened 2026-05-21
Document the concurrency model for subscribers

### LINKED ISSUE #2749: Feature: Concurrency under-the-hood explanation in Docs

It would be great for users of the library if you could explain prefetch vs workers in the docs for controlling the concurrency of these brokers. In particular, if n prefetch means n processed at once etc. or just n 'held locally' by the consumer

Knowing the under-the-hood working can help us in fine-tuning the brokers

Many thanks

00yhj22-debug:
# Description

Closes #2749.

Adds a dedicated docs page (`docs/docs/en/getting-started/subscription/concurrency.md`) that explains the two independent levers FastStream offers for concurrent processing — `max_workers` and RabbitMQ's `prefetch_count` — and clears up the specific confusion called out in the issue:

> Does `prefetch_count=10` mean ten messages are processed at the same time?
> → No. It means the broker may deliver up to ten before requiring acks; how many are processed in parallel is determined by `max_workers`.

The page covers:

- What each setting does on its own.
- How they combine (with a small table of the four common combinations).
- A note that Kafka/Confluent have no `prefetch_count` and link to the existing "Concurrent processing" notes on the Kafka subscriber docs.
- A note that Redis/NATS/MQTT only have `max_workers`.

Also wires the new page into `docs/docs/navigation_template.txt` after the existing "Dynamic Subscribers" entry.

## Type of change

## Checklist

- [x] I have made the necessary changes to the documentation (this *is* the docs change)
- [ ] I have added tests to validate the effectiveness of my fix

No tests added — the change is a new docs page plus a single-line nav entry.

### DIFF
--- a/docs/docs/en/getting-started/subscription/concurrency.md
+++ b/docs/docs/en/getting-started/subscription/concurrency.md
@@ -10,9 +10,9 @@ search:
 
 # Concurrency
 
-**FastStream** gives you two independent levers for processing messages concurrently. Mixing them up is a common source of confusion, so it helps to keep their roles straight.
+By default, a subscriber processes one message at a time: the next message is not handled until the current one is finished. `max_workers` lifts that restriction.
 
-## `max_workers` — handler-level parallelism
+## `max_workers`
 
 `max_workers` is the **number of concurrent handler invocations** a single subscriber may run. It is available on every broker's subscriber decorator:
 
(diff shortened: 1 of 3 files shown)


00yhj22-debug:
Gentle ping on this one - it's been open a couple of months without a review, so I wanted to check whether it's still wanted.

For context: it's docs-only - a new `concurrency.md` page plus a single nav line (+70/-0, no code changes) - covering the `max_workers` vs `prefetch_count` confusion raised in #2749. The branch is still clean against `main` and CI is green.

Completely understand if docs are low priority right now. I'm happy to rework the structure, narrow the scope, or simply close it if it isn't a direction you want - whichever is least work for you. Just let me know either way.

IvanKirpichnikov:
Hi. I'm sorry for waiting so long and thank you for the Pull Request!

In fact, the whole article is a comparison of max_workers and prefetch_count, that is, the whole article is about RMQ.

Let's split this article into two.
We'll leave one part about max_workers in the 'Get Started' section.
The second part, which relates to RMQ, that is, about prefetch_count itself and the comparison of prefetch_count and max_workers, will be included in the documentation about RMQ integration.

00yhj22-debug:
Thanks for the review — and no worries about the wait.

You are right, and it is a better split than what I had. `max_workers` is broker-agnostic, so putting it next to a RabbitMQ-only setting made the whole page read as RabbitMQ-specific. Done:

**`getting-started/subscription/concurrency.md`** — now only `max_workers`: what it does, the default, and that it is handler-level parallelism meaning the same thing on every broker. A closing note points at each broker's own fetch-side knob (`prefetch_count` for RabbitMQ, `max_poll_records` for Kafka, none for NATS/Redis/MQTT), so the distinction between "processed at once" and "fetched from the broker" is still made without this page owning the details.

**`rabbit/prefetch.md`** (new) — `prefetch_count` itself, the `max_workers` x `prefetch_count` combination table, the rule of thumb about keeping `prefetch_count >= max_workers`, and the "does `prefetch_count=10` mean ten at a time?" section. Added to the nav under RabbitMQ, after Acknowledgement.

The two pages cross-link, so a reader landing on either one can reach the other. Relative paths verified against the tree.

Happy to move the RabbitMQ page elsewhere in the nav, or trim the note in Get Started further if you would rather it stayed minimal — just say the word.

00yhj22-debug:
Heads-up on the red check — `check-linting-and-docs-changes` is failing on something unrelated to this PR, so please don't read it as a problem with the docs split.

`pre-commit` runs with `--all-files`, and the `typos` hook wants to rewrite a GitHub handle in `docs/docs/en/release.md`:

```
-* feat(rabbit): support EXTERNAL authentication by [@WellFREEzZ](...)
+* feat(rabbit): support EXTERNAL authentication by [@WellFREEEzZ](...)
```

It only changes the link text, not the URL, and it hits both the changelog line and the "New Contributors" entry for 0.7.4. The string is on `main` today, so any PR whose CI runs now picks it up — #2874 went green because its run finished on Aug 9, before those release notes landed.

I've left it alone: the handle is correct as written, and rewriting an unrelated file from a docs PR seemed like the wrong call. If you'd like, the usual fix is a `_typos.toml` with something like:

```toml
[default.extend-words]
WellFREEzZ = "WellFREEzZ"
```

Happy to send that as a separate PR if it's useful — just say the word. This PR itself only touches the two docs pages and the nav.

IvanKirpichnikov:
about the typos error. If it's not difficult, it would be good if you make a separate request pool with a bug fix.

Otherwise, everything is fine. We are waiting for the final review from @Lancetnik

IvanKirpichnikov (changes requested):
Hi! Please make a couple of changes to the documentation.

1. There is no max_workers parameter in the RMQ integration. At the moment, the documentation is incorrect on this point. This needs to be fixed.
2. Let’s add a section to RMQ stating that messages are already being read in parallel and, on the contrary, they need to be limited via Channel; otherwise, it will keep reading indefinitely.

00yhj22-debug:
Thanks — you were right on both counts, and the first one was a real error on my part. I checked before rewriting: `max_workers` appears in the subscriber factories for Kafka, Confluent, NATS, Redis and MQTT, and not once anywhere under `faststream/rabbit/`. The page claimed it was available on every broker and meant the same thing for RabbitMQ, which was simply wrong.

**1. `max_workers` and RabbitMQ**

`concurrency.md` now scopes the feature to the brokers that actually have it, and says up front that RabbitMQ works the other way around, with a note pointing at the prefetch page.

**2. RabbitMQ is already parallel and needs bounding**

`prefetch.md` now leads with that instead of the comparison table it had before:

> RabbitMQ subscribers are concurrent by default. A delivery is handled as it arrives, so several invocations of your handler can be in flight at the same time — you do not opt into that, and there is no `max_workers` option here.
>
> What you tune instead is the opposite: how many messages the broker is allowed to push at you. Without a limit it will keep sending as fast as it can, which is rarely what you want.

It then covers why an unbounded consumer hurts (memory, and messages sitting in one consumer's buffer can't be picked up by another, so replicas stop helping), how to choose a value, per-subscriber channels, and `global_qos`.

**One more fix while I was in there**

The broker-level argument was written as `RabbitBroker(channel=...)`. It is `default_channel=`. I ran every code block in the page to confirm the arguments are real:

- `RabbitBroker(..., default_channel=Channel(prefetch_count=10))`
- `@broker.subscriber("heavy-queue", channel=Channel(prefetch_count=1))`
- `RabbitBroker(..., default_channel=Channel(prefetch_count=10, global_qos=True))`

Pushed. Happy to adjust the wording if the framing still isn't quite how you'd put it.

IvanKirpichnikov (approved):

PR #2894, opened 2026-05-31
feat(docs): add json endpoint and fix content-type header

### LINKED ISSUE #2803: Feature: host the asyncapi json on a seperate endpoint/route like FastAPI
We use [backstage](https://backstage.io/) to collect project information and techdocs. Backstage has the option to render openapi and asyncapi docs from files. We do this by refering to the hosted version of openapi.json on fastapi projects and we like to do the same for faststream projects.

Currently the hosted apidocs consist of a singe page which includes in the javascript the asyncapi spec. We want this on a separate router so we can directly reference it.

**Feature code example**
I started a branch https://github.com/hmvp/faststream/tree/better_asyncapi ran into some tests that I did not understand and I was not sure if the approach was valid...

Generating the json with the commandline client introduces extra complexities...

Cool-Cat09:
# Description

Create new endpoint which  sends raw json documentation. Fix fastapi integration endpoint header "content-type: application/json"

Fixes #2803
## Type of change

## Checklist

### DIFF
--- a/docs/docs/en/getting-started/asgi.md
+++ b/docs/docs/en/getting-started/asgi.md
@@ -215,5 +215,5 @@ async def start_broker(app):
 app = FastAPI(lifespan=start_broker)
 
 app.mount("/health", make_ping_asgi(broker, timeout=5.0))
-app.mount("/asyncapi", make_asyncapi_asgi(asyncapi, try_it_out=False))
+app.mount("/asyncapi", make_asyncapi_asgi(asyncapi, try_it_out_path=None))
 ```
--- a/docs/docs/en/getting-started/asyncapi/hosting.md
+++ b/docs/docs/en/getting-started/asyncapi/hosting.md
@@ -114,7 +114,7 @@ broker = NatsBroker()
 app = AsgiFastStream(broker, asyncapi_path="/docs/asyncapi")
 ```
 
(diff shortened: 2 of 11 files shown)


ded-furby (commented):
I found one API-compatibility concern in `AsyncAPIRoute.__init__`: the patch moves `asyncapi_json_path` from the second positional parameter slot to after `include_in_schema`. Because `AsyncAPIRoute` is a public type, callers that currently do something like `AsyncAPIRoute("/docs", "/docs.json")` would silently start binding that second argument to `description` instead of the JSON path.

Could we preserve the existing positional order for `asyncapi_json_path`, or make the new try-it-out option keyword-only without reordering the old positional parameters? That would avoid a breaking change while still getting the new `try_it_out_path` behavior.

Lancetnik (dismissed):

Lancetnik:
@ded-furby `asyncapi_json_path` is not part of the public API yet, so we should place it after the current positional arguments — that's exactly what preserves backward compatibility.

As for `asyncapi_json_path` — we're planning the 0.7 release today, so this is actually a good place to break backward compatibility a little.

Lancetnik (dismissed):

Lancetnik (approved):

Issue #1812, opened 2024-09-24
Check Redis Sentinel support


kumaranvpl:

davorrunje:
Please link the original post

DABND19:
I'm taking this task to work.

NikolayBaranovv:
Great! I'm waiting for this feature.
Maybe you need some help with implementation?

kumaranvpl:
@NikolayBaranovv I am not working on it. @DABND19 probably is. 

@DABND19 Any update on this? @NikolayBaranovv is willing to help.

pbonneaudiabolocom:
Hi,

I'm wondering what is the status here.
We are going to need it in the company, and as 3 developers already tried, what would be their feedbacks.

Did they fail to find time, or was it too hard to be implemented ?
What was the difficulty ?

Here is the path CODEX would suggest. Does it seem relevant ?

> 1.Add an explicit Sentinel config path on RedisBroker. For example in broker.py:
> ◦sentinels: Sequence[tuple[str, int]] | None = None
> ◦sentinel_service_name: str | None = None
> ◦sentinel_kwargs: Mapping[str, Any] | None = None
> ◦maybe sentinel_role: Literal["master", "slave"] = "master"
> 
> 2.Stop passing only raw Redis kwargs into ConnectionState. Right now broker.py builds connection_options and does:
> ◦connection_state = ConnectionState(connection_options) That is too narrow. ConnectionState should either:
> ◦accept a client factory, or
> ◦support two modes: direct Redis and Sentinel.
> 
> 3.Implement Sentinel-aware connection creation in state.py. Roughly:
> ◦direct mode: keep current ConnectionPool(...)
> ◦sentinel mode: create redis.asyncio.sentinel.Sentinel(...), then sentinel.master_for(...) Important: Sentinel constructor kwargs and Redis client kwargs are not the same, so you need to split them.
> 
> 4.Keep _resolve_url_options() for direct Redis only. Sentinel usually won’t fit cleanly into the current single-URL model in broker.py. I would not try to overload redis://... parsing for Sentinel first. Add explicit params instead.
> 
> 5.Handle reconnection for subscribers, not only publishers. This is the main trap. Subscribers keep long-lived objects:
> ◦pub/sub in channel_subscriber.py
[...]

PR #2854, opened 2026-05-04
Add Redis Cluster broker support

powersemmi:
# Description

This PR introduces `RedisClusterBroker` — a new broker for Redis Cluster, inheriting from `RedisBroker` with the same constructor API for drop-in compatibility.

Key features:
- **Redis Cluster** connection via `startup_nodes` with auto-discovery
- **Standard Pub/Sub** (`PUBLISH`/`SUBSCRIBE`, `sharded=False`) for single-node delivery
- **List** and **Stream** operations with automatic hash-slot routing via `RedisCluster` client
- **XAUTOCLAIM** support for cyclic re-consumption of pending stream messages (`min_idle_time`)
- Full `RedisBroker` API compatibility — migrate by changing the class name
- Pipeline warning — raises `RuntimeWarning` if `pipeline=` is passed (not supported in cluster)

Documentation:
- New `docs/docs/en/redis/cluster.md` — feature comparison, migration guide, limitations
- Cross-references from Pub/Sub, Streams, Pipeline, and Claiming pages

Fixes #1170

## Type of change

- [x] New feature (a non-breaking change that adds functionality)
- [x] This change requires a documentation update

## Checklist

- [x] My code adheres to the style guidelines of this project (`just lint` shows no errors)
- [x] I have conducted a self-review of my own code
- [x] I have made the necessary changes to the documentation
- [x] My changes do not generate any new warnings
- [x] I have added tests to validate the effectiveness of my fix or the functionality of my new feature
- [x] Both new and existing unit tests pass successfully on my local environment by running `just test-coverage`
- [x] I have ensured that static analysis tests are passing by running `just static-analysis`
- [x] I have included code examples to illustrate the modifications

PR #2895, opened 2026-05-31
feat(redis): add Redis Sentinel support

PAzter1101:
## Summary

Adds connection-level Redis Sentinel support to `RedisBroker`, addressing #1812.

```python
broker = RedisBroker(
    sentinels=[("sentinel-1", 26379), ("sentinel-2", 26379)],
    sentinel_master_name="mymaster",
)
```

## What changed

- **New broker params** (`RedisConnectionParams`): `sentinels`, `sentinel_master_name`, `sentinel_kwargs`. Extracted in `RedisBroker.__init__`; a missing `sentinel_master_name` raises `SetupError`. Excluded from cluster params.
- **`RedisConnectionState`** gains an optional `SentinelConfig`. When set, `connect()` builds the client via `redis.asyncio.sentinel.Sentinel(...).master_for(...)` instead of a plain `ConnectionPool`. The resulting `SentinelConnectionPool` re-discovers the current master on every reconnect, so **publishers and stream (`XREADGROUP`) consumers fail over for free** — both already go through `connection.client`.
- **Stream subscriber busy-loop fix**: the stream `_consume` loop had no backoff in its `except` (unlike the base channel/list loop), so it would hot-loop on a persistent `ConnectionError` (e.g. during a master failover). Extracted the existing `5`s backoff into a shared `CONSUME_ERROR_BACKOFF_SECONDS` and applied it in both loops.
- **Tests**: unit tests that need no running Sentinel — assert the broker builds the right `SentinelConfig`, that direct mode is unaffected, validation, and that `connect()` actually yields a `SentinelConnectionPool`. New `redis_sentinel` pytest marker.

`mypy --strict`, `ruff check`, `ruff format --check` and the new unit tests pass locally.

## Scope / follow-ups

This is the connection-level MVP outlined in #1812 (publishers + simple commands + streams). Intentionally **not** in this PR, to keep it focused:

- **Pub/Sub (channel) resubscribe after failover.** A channel subscriber holds a long-lived `pubsub()` subscription; rebuilding and resubscribing it on failover is a separate change and is left as a follow-up. (Stream consumers recover because each `xreadgroup` re-acquires a connection from the Sentinel pool.)
- **Integration tests against a real Sentinel** (a compose service + CI job mirroring the existing `redis-cluster` setup). The tests here are unit-level.

Refs #1812.


pbonneaudiabolocom:
Hi,

Thanks for the hard work.

Seeing the PR for redis cluster, I wonder if it would not be "better" to create a RedisSentinelBroker than to add parameters to the generic RedisBroker.

It would prevent these param to be available in redisCluster as well, and it would somehow keep the same logic where you just switch from one broker to the other one when changing technology.

What do you think ?

PAzter1101:
Thanks for the suggestion — agreed. I've reworked it into a dedicated `RedisSentinelBroker` (plus `RedisSentinelRouter` for the FastAPI integration), mirroring `RedisClusterBroker` from #2854. The Sentinel params now live only on the Sentinel classes and no longer leak into `RedisBroker`/`RedisClusterBroker`. The earlier "params on `RedisBroker`" commits are kept in the history for context.

PAzter1101:
Hi @Lancetnik 👋

Thanks for the 👍 on reworking this into a dedicated `RedisSentinelBroker` — glad the direction looks right.

Quick status: the only red check was `check-linting-and-docs-changes`, and it turned out to be a tiny `ruff format` nit in `tests/brokers/redis/test_sentinel.py` (two 2-arg calls ruff wanted on a single line). Fixed in the latest commit (`7c95d67c`) — `ruff format --check` and `ruff check` both pass locally now. All the actual test jobs (redis-real, redis-cluster, every broker, Python 3.10–3.14) were already green.

The new run is sitting in "waiting for approval" — could you kick it off when you have a moment? Happy to address anything else.

Thanks!

pbonneaudiabolocom:
@Lancetnik would you agree on integrating tha feature soon ?
Our infra team is pushing us into sentinel adoption more and more and it will greatly help us if that PR was available into faststream.

Lancetnik (dismissed):

powersemmi (changes requested):
@PAzter1101 hi!
`RedisSentinelBroker.__init__` is the third near-verbatim copy of `RedisBroker.__init__` (after the cluster one). By the rule of three, could you extract this into a shared hook on `RedisBroker` (e.g. `_make_connection_state()` + a param-filter set) and refactor all three brokers to use it? The subclasses would shrink to a few lines.

PAzter1101:
@powersemmi good call — done. Pulled the shared setup into `RedisBroker._init_broker` and moved the per-broker bits into `_make_connection_state` / `_make_producer` / `_validate_init_params` hooks, plus a `_EXTRA_NON_CONNECTION_PARAMS` class attr for the param filter. `RedisClusterBroker` and `RedisSentinelBroker` now keep only a thin typed `__init__` (to preserve the `RedisClusterParams` / `RedisSentinelParams` signatures for users and the mypy tests) and override the relevant hooks — net ~-90 lines.

mypy (strict), ruff and the redis test suite are green locally; rebased onto the latest main. PTAL 🙏

powersemmi (approved):

PR #2919, opened 2026-06-22
fix: declare nats-py[nkeys] extra so NKEY auth works

### LINKED ISSUE #2673: Bug: nkey support in NATS needs proper dependency declaration
You support NKEY authentication in NATS via Faststream (https://dummy.faststream.airt.ai/latest/public_api/faststream/nats/NatsBroker/) via the nkeys_seed argument. 

That in turn depends on nats-py to be installed with nkeys support (cf: https://github.com/nats-io/nats.py/blob/main/nats/pyproject.toml) i.e., you have to depend on

```python
nats-py[nkeys]
```

in your pyproject.toml file here https://github.com/ag2ai/faststream/blob/main/pyproject.toml:

I suggest to change 

```toml
nats = ["nats-py>=2.12.0,<=3.0.0"]
````

to

```toml
nats = ["nats-py[nkeys]>=2.12.0,<=3.0.0"]
```

or make this another dependency group. Otherwise you'll run into:

```python
ERROR:    Traceback (most recent call last):
  File "/mycode", line 391, in _start_messagebus_connection
    await self.router.broker.connect()
  File "/srv/nacl/lib/python3.12/site-packages/faststream/nats/broker/broker.py", line 602, in connect
    return await super().connect(**connect_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/srv/nacl/lib/python3.12/site-packages/faststream/broker/core/usecase.py", line 231, in connect
    self._connection = await self._connect(**connection_kwargs)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/srv/nacl/lib/python3.12/site-packages/faststream/nats/broker/broker.py", line 606, in _connect
    connection = await nats.connect(**kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/srv/nacl/lib/python3.12/site-packages/nats/__init__.py", line 46, in connect
    await nc.connect(servers, **options)
  File "/srv/nacl/lib/python3.12/site-packages/nats/aio/client.py", line 519, in connect
    self._setup_nkeys_connect()
  File "/srv/nacl/lib/python3.12/site-packages/nats/aio/client.py", line 565, in _setup_nkeys_connect
    self._setup_nkeys_seed_connect()
  File "/srv/nacl/lib/python3.12/site-packages/nats/aio/client.py", line 679, in _setup_nkeys_seed_connect
    import nkeys
ModuleNotFoundError: No module named 'nkeys'
```

cycsmail:
# Description

The `nats` optional-dependency only pulls in `nats-py`, so NKEY authentication (e.g. passing `nkeys_seed`) fails at runtime with `ModuleNotFoundError: No module named 'nkeys'`. nats-py ships its NKEY support behind an `nkeys` extra, so this declares `nats-py[nkeys]` in the `nats` extra to pull it in.

Fixes #2673

## Type of change

## Checklist

- [x] I have added a test to validate the fix (the new test in `tests/brokers/nats/test_nkeys_dependency.py` asserts the `nats` extra requires `nats-py[nkeys]`, and it passes locally)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,7 +73,7 @@ confluent = [
     "confluent-kafka>=2.6,!=2.8.1,<3; python_version >= '3.13'",
 ]
 
-nats = ["nats-py>=2.12.0,<=3.0.0"]
+nats = ["nats-py[nkeys]>=2.12.0,<=3.0.0"]
 
 redis = ["redis>=5.0.0,<8.0.0"]
 
--- a/tests/brokers/nats/test_nkeys_dependency.py
+++ b/tests/brokers/nats/test_nkeys_dependency.py
@@ -1,0 +1,35 @@
+import re
+from pathlib import Path
+
+import pytest
+
+
+def _read_pyproject() -> str:
+    for parent in Path(__file__).resolve().parents:
+        pyproject = parent / "pyproject.toml"
+        if pyproject.exists():
+            return pyproject.read_text(encoding="utf-8")
+    msg = "pyproject.toml not found"
+    raise FileNotFoundError(msg)
+
+
+@pytest.mark.nats()
+def test_nats_extra_requires_nkeys() -> None:
+    """The ``nats`` extra must install ``nats-py`` with its ``nkeys`` extra.
+
+    Without it, NKEY authentication (e.g. the ``nkeys_seed`` argument) fails at
+    runtime with ``ModuleNotFoundError: No module named 'nkeys'`` (issue #2673).
+    """
+    content = _read_pyproject()
+
+    match = re.search(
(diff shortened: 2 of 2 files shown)


Lancetnik (changes requested):
Thanks for the PR, but I don't think we should go this route. We shouldn't install nats-py with specific extras by default for all users. Not everyone needs nkeys, and it's correct behavior that using this parameter raises an error when the required dependencies aren't installed.

The only thing that could be improved here is the error message. But I'm not sure FastStream is the best place to do that — it might be better to send a PR to nats-py instead.

PR #2941, opened 2026-07-15
fix(confluent): retry on BufferError so a full queue doesn't cancel the whole batch

### LINKED ISSUE #2836: Bug: confluent publish_batch — one message over queue.buffering.max.messages cancels entire batch via TaskGroup cascade
## `publish_batch` can trigger cascading `BufferError` when batch size exceeds librdkafka's `queue.buffering.max.messages`

**NOTE:** This issue is generated with AI assistance, but:
- It is thorough because 
  - I cloned down confluent-kafka-python, librdkafka, faststream to local dir
  - checked out all deps' repos at relevant levels, and faststream at 0.6.7
- I'm doing this only after I hit this exact issue in a real env.; and so want to give a heads up about this... for when `publish_batch` encounters a super lumpy produce request of msgs.  

### Summary

Confluent backend's `publish_batch` with a batch larger than librdkafka's producer queue capacity causes a single `BufferError` on the overflowing message to cancel **all** sibling tasks via `anyio.TaskGroup` — including those already enqueued in librdkafka.  
Orphaned delivery callbacks then try to resolve dead `asyncio.Future`s, and the entire batch is effectively lost even though only one (or how-many-ever above the max msgs config) message failed to enqueue.

### The config parameter

- **`queue.buffering.max.messages`** — default **100,000**, max messages allowed on the producer queue
- **`queue.buffering.max.kbytes`** — default **1,048,576 KB** (1 GB), *higher priority* than message count
- Either limit being hit triggers the same `QUEUE_FULL → BufferError` path
- Both are known to faststream's `ConfluentFastConfig` ([config.py L279-L280](https://github.com/ag2ai/faststream/blob/0.6.7/faststream/confluent/helpers/config.py#L279-L280))

### Reproduction

Requires a running Kafka broker on `localhost:9092`. Run with `uv run repro.py`:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "faststream[confluent]>=0.6.7,<0.8",
# ]
# ///
"""Reproduces BufferError cascade in faststream's confluent publish_batch.

The default librdkafka queue.buffering.max.messages is 100000.
Sending 100001 messages triggers BufferError on the last enqueue,
which cascades through anyio's TaskGroup and cancels all 100000
in-flight tasks — effectively losing the entire batch.
"""
import asyncio
import logging

from faststream.confluent import KafkaBroker

logging.basicConfig(level=logging.WARNING)

invalid_state_count = 0
_default_handler = None


def counting_exception_handler(loop, context):
    global invalid_state_count
    exc = context.get("exception")
    if isinstance(exc, asyncio.InvalidStateError):
        invalid_state_count += 1
    elif _default_handler:
        _default_handler(context)


async def main() -> None:
    global _default_handler
    loop = asyncio.get_running_loop()
    _default_handler = loop.get_exception_handler()
    loop.set_exception_handler(counting_exception_handler)

    async with broker:
        batch_size = 100_001  # one over the default limit
        msgs = [f"msg-{i}" for i in range(batch_size)]
        print(f"Publishing batch of {batch_size} messages...")
        try:
            await broker.publish_batch(
                *msgs,
                topic="faststream-buffererror-repro",
            )
            print("Batch published successfully (unexpected)")
        except BufferError as e:
            print(f"BufferError: {e}")
        except ExceptionGroup as eg:
            print(f"ExceptionGroup with {len(eg.exceptions)} sub-exceptions:")
            for exc in eg.exceptions[:5]:
                print(f"  {type(exc).__name__}: {exc}")
        except Exception as e:
            print(f"{type(e).__name__}: {e}")

    # Give poll loop time to fire orphaned callbacks
    await asyncio.sleep(2)
    print(f"InvalidStateError from orphaned delivery callbacks: {invalid_state_count}")


if __name__ == "__main__":
    asyncio.run(main())
```

Output:
```python
Publishing batch of 100001 messages...
ExceptionGroup with 1 sub-exceptions:
  BufferError: Local: Queue full
InvalidStateError from orphaned delivery callbacks: 100000
```

Also triggers with fewer messages if they're large enough to exceed `queue.buffering.max.kbytes`.

### Root cause trace

Key code locations:
- `publish_batch` → `send_batch`: [producer.py L154-L183](https://github.com/ag2ai/faststream/blob/0.6.7/faststream/confluent/publisher/producer.py#L154-L183)
- `send_batch` (one task per message in TaskGroup): [client.py L142-L162](https://github.com/ag2ai/faststream/blob/0.6.7/faststream/confluent/helpers/client.py#L142-L162)
- `send` (sync `produce()` + `await result_future`): [client.py L70-L136](https://github.com/ag2ai/faststream/blob/0.6.7/faststream/confluent/helpers/client.py#L70-L136)
- librdkafka queue check (`QUEUE_FULL` → `BufferError`): [rdkafka_int.h L792-L827](https://github.com/confluentinc/librdkafka/blob/v2.13.2/src/rdkafka_int.h#L792-L827), [Producer.c L334-L336](https://github.com/confluentinc/confluent-kafka-python/blob/v2.13.2/src/confluent_kafka/src/Producer.c#L334-L336)

```mermaid
sequenceDiagram
    participant FS as FastStream send_batch()
    participant TG as anyio TaskGroup
    participant S1 as send() task 1..100000
    participant SN as send() task 100001
    participant P as confluent Producer.produce()
    participant RD as librdkafka queue

    FS->>TG: create_task_group()
    TG->>S1: start_soon(send, msg_1)
    Note over S1: ...repeated 100000 times
    TG->>SN: start_soon(send, msg_100001)

    S1->>P: produce() [sync, enqueues]
    P->>RD: rd_kafka_curr_msgs_add(cnt <= 100000)
    RD-->>P: OK
    P-->>S1: None
    S1->>S1: await result_future [blocks on delivery cb]

    SN->>P: produce() [sync]
    P->>RD: rd_kafka_curr_msgs_add(cnt=100001 > max=100000)
    RD-->>P: QUEUE_FULL
    P-->>SN: BufferError!

    Note over TG: TaskGroup sees unhandled exception
    TG->>TG: Cancel ALL sibling tasks
    TG->>S1: CancelledError to 100000 tasks

    Note over S1: result_future abandoned mid-await
    Note over RD: Messages still in librdkafka queue
    Note over RD: _poll_loop fires ack_callbacks
    Note over S1: set_result() on cancelled/dead futures
    Note over S1: InvalidStateError (100000 of them)
```

### Proposed fix: chunk the batch to respect queue capacity

Split the batch into sub-batches that fit within the queue limit, flushing between each chunk.

- `self.config` ([client.py L45](https://github.com/ag2ai/faststream/blob/0.6.7/faststream/confluent/helpers/client.py#L45)) is a `dict[str, Any]` — only contains user-set keys, not librdkafka defaults
- No Python API to read effective config back → use `self.config.get("queue.buffering.max.messages", 100000)` with hardcoded default (stable since librdkafka v0.9, [CONFIGURATION.md](https://github.com/confluentinc/librdkafka/blob/v2.13.2/CONFIGURATION.md))
- `flush()` already exists ([client.py L67-L68](https://github.com/ag2ai/faststream/blob/0.6.7/faststream/confluent/helpers/client.py#L67-L68)), already used by `autoflush` ([factory.py L62-L74](https://github.com/ag2ai/faststream/blob/0.6.7/faststream/confluent/publisher/factory.py#L62-L74)) and `stop()` ([client.py L65](https://github.com/ag2ai/faststream/blob/0.6.7/faststream/confluent/helpers/client.py#L65))

#### Sketch of the change

```python
async def send_batch(self, batch, topic, *, partition, no_confirm=False):
    max_msgs = int(self.config.get("queue.buffering.max.messages", 100000))
    msgs = batch._builder

    for i in range(0, len(msgs), max_msgs):
        chunk = msgs[i : i + max_msgs]
        async with anyio.create_task_group() as tg:
            for msg in chunk:
                tg.start_soon(
                    self.send,
                    topic,
                    msg["value"],
                    msg["key"],
                    partition,
                    msg["timestamp_ms"],
                    msg["headers"],
                    no_confirm,
                )
        # Drain the librdkafka queue before enqueuing the next chunk.
        if i + max_msgs < len(msgs):
            await self.flush()
```

No new deps, no behavior change for batches that fit, flush only between chunks.  
Expected output with this fix (same 100,001 messages):
```python
Publishing batch of 100001 messages...
Batch published successfully
InvalidStateError from orphaned delivery callbacks: 0
```
**Note:** this only chunks by message count, not by `queue.buffering.max.kbytes`.  
Large messages that hit the bytes limit before the count limit would still trigger the error.  
Handling both would require tracking cumulative message sizes per chunk.

### Further improvement: graceful partial failure with `aioresult`

Even with chunking, other errors inside `send()` still cancel all siblings.  
[`aioresult.ResultCapture`](https://github.com/arthur-tacca/aioresult) captures per-task results without propagating into the TaskGroup:

```python
from aioresult import ResultCapture

async def send_batch(self, batch, topic, *, partition, no_confirm=False):
    max_msgs = int(self.config.get("queue.buffering.max.messages", 100000))
    msgs = batch._builder

    for i in range(0, len(msgs), max_msgs):
        chunk = msgs[i : i + max_msgs]
        results = []
        async with anyio.create_task_group() as tg:
            for msg in chunk:
                rc = ResultCapture.start_soon(
                    tg,
                    self.send,
                    topic,
                    msg["value"],
                    msg["key"],
                    partition,
                    msg["timestamp_ms"],
                    msg["headers"],
                    no_confirm,
                    suppress_exception=True,
                )
                results.append(rc)

        errors = [r.exception() for r in results if r.exception() is not None]
        if errors:
            raise ExceptionGroup("publish_batch partial failure", errors)

        if i + max_msgs < len(msgs):
            await self.flush()
```
With aioresult, if a message still fails (e.g. user set a lower limit), only that one is reported — the rest succeed:

```python
ExceptionGroup: publish_batch partial failure (1 sub-exception)
  BufferError: Local: Queue full
# but 100000 messages delivered successfully
```

### Versions tested

| Component | Version |
|---|---|
| faststream | 0.6.7 (also checked 0.7.0rc1 — `send_batch` is identical) |
| confluent-kafka-python | 2.13.2 |
| librdkafka | 2.13.2 |

cycsmail:
# Description

Fixes #2836.

When a batch pushes librdkafka's local produce queue past `queue.buffering.max.messages`, `produce()` raises `BufferError` ("Local: Queue full"). That error was bubbling up out of a single `send` task and cancelling the whole `send_batch` task group, so one overflowing message took down every sibling message in the batch.

This catches `BufferError` in the produce path, serves delivery reports with `poll()` to drain the queue, then retries that same message. It's the usual confluent-kafka backpressure pattern, so the rest of the batch goes through instead of being cancelled.

## Type of change

## Checklist

Added `tests/brokers/confluent/test_buffer_full.py`, which fakes a full queue on the first produce call and checks the batch still completes (it fails without the fix). That test and the existing confluent unit tests pass locally.

### DIFF
--- a/faststream/confluent/helpers/client.py
+++ b/faststream/confluent/helpers/client.py
@@ -50,6 +50,11 @@ class _LazyLoggerProxy(logging.Logger):
             real_logger.handle(record)
 
 
+# How long (in seconds) to wait while draining librdkafka's local produce
+# queue before retrying a message that hit ``BufferError`` ("Local: Queue full").
+_BUFFER_FULL_POLL_TIMEOUT = 1.0
+
+
 class AsyncConfluentProducer:
     """An asynchronous Python Kafka client using the "confluent-kafka" package."""
 
@@ -148,7 +153,17 @@ class AsyncConfluentProducer:
             produce_kwargs["partition"] = kwargs["partition"]
         if kwargs.get("timestamp") is not None:
             produce_kwargs["timestamp"] = kwargs["timestamp"]
-        self.producer.produce(topic, **produce_kwargs)
+        while True:
+            try:
+                self.producer.produce(topic, **produce_kwargs)
+                break
+            except BufferError:
+                # librdkafka's local produce queue is full
+                # (queue.buffering.max.messages / queue.buffering.max.kbytes).
+                # Serve delivery reports to drain it and retry this message,
+                # instead of raising and cancelling the whole batch through the
+                # task group in ``send_batch``.
+                await call_or_await(self.producer.poll, _BUFFER_FULL_POLL_TIMEOUT)
 
         if no_confirm:
             return result_future
--- a/tests/brokers/confluent/test_buffer_full.py
+++ b/tests/brokers/confluent/test_buffer_full.py
@@ -1,0 +1,80 @@
+import time
+from typing import Any
+from unittest.mock import patch
+
+import pytest
+
+from faststream._internal.logger.state import LoggerState
+from faststream.confluent.helpers.client import AsyncConfluentProducer
+from faststream.confluent.helpers.config import ConfluentFastConfig
+
+
+class _FakeProducer:
+    """Stand-in for ``confluent_kafka.Producer``.
+
+    ``produce`` raises ``BufferError`` ("Local: Queue full") on the call
+    indices listed in ``fail_calls`` (mimicking librdkafka rejecting a message
+    once its local queue is full), and otherwise immediately fires the delivery
+    callback so the awaiting future resolves.
+    """
+
+    def __init__(self, fail_calls: set[int]) -> None:
+        self._fail_calls = fail_calls
+        self.produce_calls = 0
+        self.produced: list[tuple[str, dict[str, Any]]] = []
+
+    def produce(self, topic: str, *, on_delivery: Any = None, **kwargs: Any) -> None:
+        self.produce_calls += 1
+        if self.produce_calls in self._fail_calls:
+            msg = "Local: Queue full"
+            raise BufferError(msg)
+        self.produced.append((topic, kwargs))
+        if on_delivery is not None:
+            on_delivery(None, None)
+
+    def poll(self, timeout: float = 0) -> int:
+        # Throttle the background poll loop a little so it does not busy-spin.
+        time.sleep(0.01)
+        return 0
+
+    def flush(self, *args: Any, **kwargs: Any) -> int:
+        return 0
+
+
+@pytest.mark.confluent()
+@pytest.mark.asyncio()
+async def test_send_batch_survives_buffer_full() -> None:
+    """A single ``BufferError`` must not cancel the whole batch.
+
+    Regression test for #2836: a message over ``queue.buffering.max.messages``
+    used to raise ``BufferError`` inside one ``send`` task, which cascaded
+    through the ``send_batch`` task group and cancelled every sibling message.
+    The overflowing message should instead be retried after draining the queue.
+    """
+    # The first produce call hits a full queue; the retry must succeed.
+    fake = _FakeProducer(fail_calls={1})
+
+    with patch(
+        "faststream.confluent.helpers.client.Producer",
+        return_value=fake,
+    ):
+        producer = AsyncConfluentProducer(
+            logger=LoggerState(),
+            config=ConfluentFastConfig(),
+        )
+
+    try:
+        batch = producer.create_batch()
+        batch_size = 5
+        for i in range(batch_size):
+            batch.append(value=f"msg-{i}".encode())
+
+        # Must not raise an ExceptionGroup / BufferError.
+        await producer.send_batch(batch, "topic", partition=None)
+
+        # Every message is enqueued, with exactly one extra produce call for the
+        # retry that followed the BufferError.
+        assert len(fake.produced) == batch_size
+        assert fake.produce_calls == batch_size + 1
+    finally:
+        await producer.stop()


cycsmail:
signed the CLA now, sorry for the noise. mind reopening? happy to rebase if it's drifted.

PR #2975, opened 2026-07-31
fix(cli): relax Typer version constraint

### LINKED ISSUE #2969: Feature: relax typer version
if no incompatibility, it will be great to relax the cap of `typer`, easy for projects using faststream cli to upgrade `typer`

lntutor:
# Description

Removes the Typer upper bound from the CLI extra and refreshes the lockfile to Typer 0.27.0. The existing exclusion for Typer 0.12 remains unchanged.

Typer 0.27 vendors Click, so the startup-validation renderer now selects the matching Click exception classes while retaining a fallback for older Typer releases.

Fixes #2969

## Type of change

## Checklist

- [x] Relevant CLI tests pass locally

## Verification

- `uv lock --check`
- `uv run pytest -q tests/cli`
- `just mypy`
- Verified with Typer 0.27.0


lntutor:
Current head `0127a5435759f6bbafc20bae2fa8387c37026518` includes the follow-up fix for the cross-platform `tests/cli/utils/test_errors.py::test_draw_startup_errors` failure by stripping ANSI escape codes before asserting on the rendered option names. Focused exact-head local verification passed for that regression test in both normal and forced-color modes.

On GitHub, the remaining blockers visible on this exact SHA are both external to the code diff:
- the contributor workflows `Run all tests`, `Run linters`, `Dependency Review`, and `CodeQL` are currently `action_required` with no jobs started, so they appear to need maintainer workflow approval for this fork head before they can run;
- `license/cla` is still pending.

Once those two gates are cleared, the previous attributable test failure should already be addressed on the current head.

Lancetnik:
Sorry, we don't accept PRs from automatic AI accounts

PR #2976, opened 2026-08-01
fix(redis): consume existing entries in new groups

### LINKED ISSUE #2694: Bug: consumers of a redis StreamSub consumer groups are unable to read messages from before the creation of the group.
**Describe the bug**
When using redis StreamSub with consumer groups, consumers of a group are unable to read messages from before the creation of the group.

**How to reproduce**
Include source code:

```python
...
```

1. Ensure a redis stream has existing messages
2. Start a faststream app with a redis StreamSub -
    broker.subscriber(stream=StreamSub(stream="my_stream", group="my_group", consumer="my_consumer", last_id="0"))

**Expected behavior**
I expect the handler to receive the existing messages in the stream.

**Observed behavior**
It does not.

**Screenshots**
If applicable, attach screenshots to help illustrate the problem.

**Environment**
Running FastStream 0.6.2 with CPython 3.12.12 on Linux

**Additional context**
The issue is how `last_id` is used with xgroupcreate and xreadgroup.
If am not wrong, to be able to read messages from the beginning of a redis stream, xgroupcreate must be created with last_id "0" and xreadgroup with last_id ">". But when StreamSub is created with last_id = "0", even the xreadgroup is called with last_id="0" and this does not return any messages.

dk3yyyy:
## Summary

- separate the consumer-group creation cursor from the initial `XREADGROUP` selector
- read undelivered entries with `>` after creating a new group from an explicit stream ID
- preserve explicit-ID reads for pending entries when the consumer group already exists
- add connected Redis regressions for both new-group history and existing-group pending entries

Fixes #2694

## Type of change

## Verification

- Fast coverage suite: `3,880 passed, 37 skipped, 7 xfailed`; 73.05% total coverage
- Connected Redis regression and preservation tests on Python 3.10 and Python 3.14
- Full affected Redis consumer module and all top-level standalone Redis tests
- Ruff format/check and codespell: passed
- Mypy: no issues in 510 source files
- Bandit: no issues
- Semgrep: 0 findings across 504 files / 290 rules
- Source distribution and wheel build: passed

## Checklist

- [x] Documentation changes are not required for this internal bug fix
- [x] I have added tests to validate the fix
- [x] New and existing fast tests pass with coverage
- [x] Static analysis passes
- [x] The issue reproduction and regression tests demonstrate the behavior


faststream/redis/subscriber/usecases/stream_subscriber.py
```diff
@@ -139,6 +140,10 @@ async def start(self) -> None:
             except ResponseError as e:
                 if "already exists" not in str(e):
                     raise
+            else:
+                read_id = ">"
```
IvanKirpichnikov: may be just `self.last_id = ">"`?

faststream/redis/subscriber/usecases/stream_subscriber.py
```diff
@@ -148,7 +153,7 @@ def read(
                     return client.xreadgroup(
                         groupname=stream.group,
                         consumername=stream.consumer,
-                        streams={stream.name: stream.last_id},
+                        streams={stream.name: read_id},
```
IvanKirpichnikov: then it's just `streams={stream.name: self.read_id},`

faststream/redis/subscriber/usecases/stream_subscriber.py
```diff
@@ -139,6 +140,10 @@ async def start(self) -> None:
             except ResponseError as e:
                 if "already exists" not in str(e):
                     raise
+            else:
+                read_id = ">"
+
+            self.last_id = read_id
```
IvanKirpichnikov: And it won't be necessary.

tests/brokers/redis/test_consume.py
```diff
@@ -662,6 +662,55 @@ async def handler(msg: RedisMessage) -> None: ...
...
+        @consume_broker.subscriber(
+            stream=StreamSub(queue, group="group", consumer=queue, last_id="0"),
+        )
+        async def handler(msg: RedisMessage) -> None:
+            mock(msg)
+            event.set()
+
+        async with self.patch_broker(consume_broker) as br:
+            await br.publish({"data": "before-group"}, stream=queue)
+            await br.start()
```
IvanKirpichnikov, line 683: may be at first `br.start` and then `br.publish`?

tests/brokers/redis/test_consume.py
```diff
@@ -662,6 +662,55 @@ async def handler(msg: RedisMessage) -> None: ...
...
+        queue: str,
+        event: asyncio.Event,
+        mock: MagicMock,
+    ) -> None:
+        consume_broker = self.get_broker()
+
+        @consume_broker.subscriber(
+            stream=StreamSub(queue, group="group", consumer=queue, last_id="0"),
+        )
+        async def handler(msg: RedisMessage) -> None:
```
IvanKirpichnikov: use `msg: Any`. so get msg body, but this is how the RedisMessage instance turns out.

tests/brokers/redis/test_consume.py
```diff
@@ -662,6 +662,55 @@ async def handler(msg: RedisMessage) -> None: ...
...
```
IvanKirpichnikov: use `msg: Any`. so get msg body, but this is how the RedisMessage instance turns out.

IvanKirpichnikov (commented):
Thanks for Pull Request!

faststream/redis/subscriber/usecases/stream_subscriber.py
dk3yyyy: Addressed in `76690ca` — I adopted the simplification (dropped the throwaway `read_id` local, set the selector once in the branch, referenced it directly in `streams=`) but used a dedicated `self.read_id` attribute rather than `self.last_id`:

```python
self.read_id = self.last_id          # __init__
...
else:
    self.read_id = ">"               # fresh group: consume existing entries
self.last_id = self.read_id           # keep cursor sane for get_one/__aiter__
...
streams={stream.name: self.read_id}   # pinned — never mutated
```

`self.last_id` cannot double as the selector because `_get_msgs` overwrites it with the last consumed message ID after every batch (`self.last_id = msgs[-1][0].decode()`), and `XREADGROUP` with an explicit (non-`>`) ID only reads pending entries from the PEL — never new messages. So once the first batch is consumed the selector would drift to an explicit ID and new-message delivery would silently stall.

Verified against a real Redis (fresh group, `last_id="0"`, backlog + one message published after the first poll):

- poll 1 with `>` → delivers backlog (…-0, …-1, …-2)
- `self.last_id` becomes …-2
- new message published
- next poll with `self.last_id` (…-2) → **no messages** — the new message is never delivered
- next poll with pinned `>` → delivers it

The trailing `self.last_id = self.read_id` is kept on purpose: for a freshly created group it resets `self.last_id` to `>` so `get_one`/`__aiter__` keep reading new messages after `start()`; for an existing group it is a no-op.

tests/brokers/redis/test_consume.py
dk3yyyy: Intentional — the message has to exist *before* the group is created, so this test exercises the new-group-history path the PR fixes. Starting the broker first would create the group before the message exists and the entry would be consumed as an ordinary new message, so the regression would no longer be covered. (The assertion `mock.assert_called_once_with({"data": "before-group"})` confirms the handler receives the raw body.)

tests/brokers/redis/test_consume.py
dk3yyyy: Done in `76690ca` — handler is now `msg: Any` (it receives the decoded body, not a `RedisMessage`).

tests/brokers/redis/test_consume.py
dk3yyyy: Done in `76690ca` — handler is now `msg: Any` (it receives the decoded body, not a `RedisMessage`).

IvanKirpichnikov (approved):

PR #2977, opened 2026-08-02
feat: #2101 Generate AsyncAPI schema for msgspec Structs

### LINKED ISSUE #2101: Feature: Generate AsyncAPI from msgspec.Struct
To suggest an idea or inquire about a new Message Broker supporting feature or any other enhancement, please follow this template:

Since 0.6, FastStream is going to support MsgSpec (according to https://github.com/Lancetnik/FastDepends/blob/51524261c847998bb3dafe7fc8dff76c23321396/fast_depends/msgspec/serializer.py). But for now, AsyncAPI is build only for Pydantic models, although MsgSpec also supports [generating JSON Schema](https://jcristharif.com/msgspec/jsonschema.html) for specific model.

It would be nice if FastStream could generate AsyncAPI for MsgSpec structs.

-

**Feature code example**
-

-

**Additional context**
-

pelazas:
Closes https://github.com/ag2ai/faststream/issues/2101

# Description

AsyncAPI generation builds payload schemas through Pydantic. Pydantic cannot describe a `msgspec.Struct` and raises rather than degrading, so today *any* handler annotated with a Struct takes down the whole schema build:

```python
class User(msgspec.Struct):
    name: str

@broker.subscriber("test")
async def handler(user: User) -> None: ...

AsyncAPI(broker, schema_version="3.0.0").to_specification()
# pydantic.errors.PydanticSchemaGenerationError: Unable to generate
# pydantic-core schema for <class 'User'>
```

Return annotations fail the same way through the publisher path, with `PydanticInvalidForJsonSchema` instead. So `faststream docs gen` and `docs serve` are unusable for anyone on msgspec, even though 0.6 shipped msgspec serialization.

The fix keeps Structs out of the Pydantic model entirely and builds their schema with `msgspec.json.schema_components()`, using the same `#/$defs/{name}` reference template Pydantic emits. That matters because the v2.6.0 and v3.0.0 generators already call `move_pydantic_refs()` and hoist `$defs` into `components/schemas` — feeding them Pydantic-shaped dicts means nested Structs get hoisted and referenced with no changes to either generator.

Output for a Struct is now the same shape as for the equivalent Pydantic model:

```json
"User": {
  "title": "User", "type": "object",
  "properties": {
    "name": {"type": "string"},
    "address": {"$ref": "#/components/schemas/Address"}
  },
  "required": ["name", "address"]
},
"Address": {"title": "Address", "type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}
```

Covered: a lone Struct argument (described by its own schema, like a lone Pydantic model), a Struct next to other arguments, nested Structs, Struct return annotations, and with `MsgSpecSerializer` configured. `msgspec` stays an optional import behind `HAS_MSGSPEC`, so nothing changes when it is not installed, and handlers with no Struct take an early exit before any of this runs.

Not covered, and worth saying: I only special-case Structs that appear *directly* as a parameter or return annotation. A Struct nested inside a Pydantic model, or inside a generic like `list[User]`, still goes through Pydantic and will still raise. Happy to widen it if you'd rather, but that seemed like a bigger design question than the issue was asking.

## Type of change

## Checklist

Notes on the boxes above, so they mean something:

- 12 new tests in `tests/asyncapi/test_msgspec.py`, parametrised over AsyncAPI 2.6.0 and 3.0.0. All 12 fail on `main` and pass here. The whole `tests/asyncapi` suite and `tests/integrations/msgspec` pass unchanged.
- `ruff check`, `ruff format --check`, `mypy` (510 files) and `bandit` are all clean; `mypy` on the touched file is clean on `main` too, so that is not a pre-existing-error situation.
- Docs unticked because I did not find a page describing which types AsyncAPI can render — the msgspec docs cover serialization only. Point me at the right page and I will add it.
- I ran tests directly with `pytest` rather than `just test-coverage`, since the justfile runs it inside the Docker Compose stack and I ran locally.

### DIFF
--- a/faststream/specification/asyncapi/message.py
+++ b/faststream/specification/asyncapi/message.py
@@ -11,8 +11,38 @@ from faststream._internal._compat import (
     model_schema,
 )
 
+try:
+    import msgspec
+
+    HAS_MSGSPEC = True
+except ImportError:  # pragma: no cover
+    HAS_MSGSPEC = False
+
 if TYPE_CHECKING:
     from fast_depends.core import CallModel
+
+
+def is_msgspec_struct(annotation: Any) -> bool:
+    """Whether `annotation` is a msgspec Struct type."""
+    return HAS_MSGSPEC and isclass(annotation) and issubclass(annotation, msgspec.Struct)
+
+
+def get_msgspec_schema(struct: Any) -> tuple[dict[str, Any], dict[str, Any]]:
+    """Build a JSON Schema for a msgspec Struct, shaped like Pydantic's.
+
+    Returns the struct's own schema inline plus the definitions any nested
+    structs live in, using the same `#/$defs/{name}` references Pydantic emits,
+    so the generators can hoist them into `components/schemas` unchanged.
+    """
+    (schema,), definitions = msgspec.json.schema_components(
+        [struct],
+        ref_template=f"#/{DEF_KEY}/{{name}}",
+    )
+    # A Struct is always emitted as a reference into the definitions, but stay
+    # defensive: an inline schema is still usable as-is.
+    name = schema.get("$ref", "").rsplit("/", 1)[-1]
+    body = dict(definitions.pop(name)) if name in definitions else dict(schema)
+    return body, definitions
 
 
 def parse_handler_params(call: "CallModel", prefix: str = "") -> dict[str, Any]:
@@ -21,13 +51,29 @@ def parse_handler_params(call: "CallModel", prefix: str = "") -> dict[str, Any]:
     model = cast("type[BaseModel] | None", getattr(model_container, "model", None))
     assert model
 
+    # Pydantic cannot build a schema for a msgspec Struct and raises rather than
+    # degrading, so structs are kept out of the model entirely and their schema
+    # is spliced back in by get_model_schema().
+    structs = {
+        p.field_name: p.field_type
+        for p in call.flat_params
+        if is_msgspec_struct(p.field_type)
+    }
+
     body = get_model_schema(
         create_model(
             model.__name__,
-            **{p.field_name: (p.field_type, p.default_value) for p in call.flat_params},  # type: ignore[call-overload]
+            **{  # type: ignore[call-overload]
+                p.field_name: (
+                    Any if p.field_name in structs else p.field_type,
+                    p.default_value,
+                )
+                for p in call.flat_params
+            },
         ),
         prefix=prefix,
         exclude=tuple(call.custom_fields.keys()),
+        structs=structs,
     )
 
     if body is None:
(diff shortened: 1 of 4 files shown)


faststream/specification/asyncapi/message.py
```diff
@@ -11,23 +11,69 @@
...
+
+
+def get_msgspec_schema(struct: Any) -> tuple[dict[str, Any], dict[str, Any]]:
+    """Build a JSON Schema for a msgspec Struct, shaped like Pydantic's.
+
+    Returns the struct's own schema inline plus the definitions any nested
+    structs live in, using the same `#/$defs/{name}` references Pydantic emits,
+    so the generators can hoist them into `components/schemas` unchanged.
+    """
+    (schema,), definitions = msgspec.json.schema_components(
```
sobolevn: Why not `msgspec.json.schema`?

sobolevn (commented):
I would suggest splitting pydantic / msgspec schema generation. Currently adding `structs` parameter in many places seems strange :)

IvanKirpichnikov:
Hello, @pelazas! Will you comeback?

pelazas:
Hi ! I rebased on main, and did the split @sobolevn suggested: 
- Everything that touches msgspec is under `specification/asyncapi/_msgspec.py`
- now pydantic never gets given a struct
- no change on get_model_schema

Also, I think for nested cases, having it repairing the schema afterwards (Any in struct slot and overwriting once pydantic is done) wouldn't work.

For the comment: Why not msgspec.json.schema?
I changed it now, the difference is that apparently on clients that use pydantic v1 it would generate broken links on nested structs. On pydantic v2 works the same. :)

tests/asyncapi/test_msgspec.py
```diff
@@ -0,0 +1,157 @@
...
+    @broker.subscriber("test")
+    async def handler(user: User) -> None: ...
+
+    assert schemas(broker, version)["User"] == {
+        "title": "User",
+        "type": "object",
+        "properties": {
+            "name": {"type": "string"},
+            "age": {"type": "integer"},
+            "address": {"$ref": "#/components/schemas/Address"},
```
sobolevn: If there's a ref, test that component under ref exists

tests/asyncapi/test_msgspec.py
```diff
@@ -0,0 +1,157 @@
...
+        "required": ["name", "age", "address"],
+    }
+
+
+@pytest.mark.parametrize("version", VERSIONS)
+def test_nested_struct_is_hoisted_into_components(version: str) -> None:
+    broker = NatsBroker()
+
+    @broker.subscriber("test")
+    async def handler(user: User) -> None: ...
```
sobolevn: Why do you split a single test into multiple tests? just assert everything in a single place.

sobolevn (commented):
One last thing, `msgspec.Meta` is not tested, it should be. It can provide custom json-schema for fields. Please, test different versions of this feature.

PR #2983, opened 2026-08-03
bug(kafka/confluent): batch per-message key alignment raises when a Response isn't first, or on duplicate bodies

### LINKED ISSUE #2982: Bug (kafka/confluent): batch per-message key alignment raises when a Response isn't first, or on duplicate bodies
### Describe the bug

`KafkaPublishCommand._align_keys()` (added in #2951 for #2943) raises on any batch where a `Response` object is **not** the first element, and raises on batches containing duplicate bodies. Both `faststream/kafka/response.py` and `faststream/confluent/response.py` are affected — the code is duplicated verbatim.

Two failure modes, both on plain `str` bodies, no custom middleware involved:

- `ValueError: tuple.index(x): x not in tuple` — a `Response` anywhere but index 0
- `IndexError: tuple index out of range` — two equal bodies where one is wrapped in a `Response`

Root cause is in `_align_keys` / `_form_indexes`:

```python
def _align_keys(self, value: Sequence["Any"]) -> None:
    if len(self.batch_bodies) == 0:
        return
    if isinstance(self.batch_bodies[0], KafkaResponse):   # <-- only checks slot 0
        return

    new_indexes = self._form_indexes(value)
    self._per_message_keys = tuple(self._per_message_keys[i] for i in new_indexes)

def _form_indexes(self, value: Sequence["Any"]) -> tuple[int, ...]:
    bodies_seen: dict[int, Any] = {}
    for body in value:
        index = self.batch_bodies.index(body)             # <-- raw bodies, not unwrapped
        self._update_bodies(bodies_seen, index, body)
    return tuple(i for i in bodies_seen)
```

1. `__init__` normalizes the batch via `extract_per_message_keys_and_bodies()`, then assigns `self.batch_bodies = normalized`. Inside the setter, `self.batch_bodies` is still the **raw** tuple (holding `KafkaResponse` wrappers) while `value` holds the **unwrapped** bodies. `.index(body)` therefore searches for an unwrapped body in a tuple that contains the wrapper, and misses.
2. The `isinstance(self.batch_bodies[0], KafkaResponse)` check happens to make this work when a `Response` is first — that early return is doing double duty as an "are we in `__init__`?" guard (at that point `self._per_message_keys` is not even assigned yet). Any mixed batch that starts with a plain body skips the guard and hits the broken lookup.
3. `_update_bodies` recurses with `index += 1` and has no bound check, so equal bodies walk off the end of the tuple into `IndexError`.

### How to reproduce

On current `main` (`d14ec560`), no broker needed:

```python
from faststream.kafka.response import KafkaPublishCommand, KafkaResponse
from faststream.response.publish_type import PublishType

cases = {
    "response first": (KafkaResponse("b1", key=b"k1"), "plain"),
    "response last": ("plain", KafkaResponse("b2", key=b"k2")),
    "response middle": ("a", KafkaResponse("b", key=b"k"), "c"),
    "all responses": (
        KafkaResponse("b1", key=b"k1"),
        KafkaResponse("b2", key=b"k2"),
    ),
    "duplicate bodies": ("same", KafkaResponse("same", key=b"k1")),
}

for name, bodies in cases.items():
    try:
        cmd = KafkaPublishCommand(*bodies, topic="t", _publish_type=PublishType.PUBLISH)
        keys = [cmd.key_for(i) for i in range(len(cmd.batch_bodies))]
        print(f"OK   {name:18} bodies={list(cmd.batch_bodies)} keys={keys}")
    except Exception as e:
        print(f"FAIL {name:18} {type(e).__name__}: {e}")
```

Output:

```python
OK   response first     bodies=['b1', 'plain'] keys=[b'k1', None]
FAIL response last      ValueError: tuple.index(x): x not in tuple
FAIL response middle    ValueError: tuple.index(x): x not in tuple
OK   all responses      bodies=['b1', 'b2'] keys=[b'k1', b'k2']
FAIL duplicate bodies   IndexError: tuple index out of range
```

Identical output when the first import is swapped to `faststream.confluent.response`.

The same thing through the public API fails too:

```python
await broker.publish_batch("plain", KafkaResponse("b2", key=b"k2"), topic=queue)
```

### Expected behavior

Per-message keys stay attached to their bodies regardless of where `Response` objects sit in the batch, and duplicate bodies don't raise. The `"response last"` case above should behave like `"response first"`: `keys=[None, b'k2']`.

### Why CI didn't catch it

The mixed-batch test in `tests/brokers/kafka/test_publish.py::TestPublish::test_batch_publisher_kafka_response_with_default_key` publishes `KafkaResponse("message1", key=b"explicit_key"), "message2", KafkaResponse("message3", key=b"another_key")` — a `Response` in slot 0, so it takes the early return and never exercises the alignment path. Moving the plain body to the front is enough to make it fail.

### Environment

`main` @ `d14ec560`, reproduced on Python 3.12 with `aiokafka` and `confluent-kafka` installed. Affects both the aiokafka and confluent implementations.

I hit this from #2933, where a batch of `(b"hi", KafkaResponse(TOMBSTONE, key=...))` started failing after rebasing onto `main`. Happy to open a PR — matching by position/unwrapped body with a first-unused-slot scan fixes all three cases and lets `_update_bodies` go away.

ApusBerliozi:
# Description

Applied changes proposed in #2982. 
1 - Replaced check with `if not self._per_message_keys`
2 - `_align_keys()` in both `kafka/response.py` and `confluent/response.py` collapses to a call into it; `_form_indexes() / _update_bodies()` are deleted from both copies
3 - adds `realign_keys()` to `faststream/response/response.py`, matching each new body against the unwrapped form of the current bodies and claiming every slot at most once, so Response wrappers and duplicate bodies both resolve

Fixes #2982

## Type of change

## Checklist

### DIFF
--- a/faststream/confluent/response.py
+++ b/faststream/confluent/response.py
@@ -100,10 +100,10 @@ class KafkaPublishCommand(BatchPublishCommand):
         self.timeout = timeout
 
         # per-message keys support
+        self._per_message_keys = keys        
         keys, normalized = extract_per_message_keys_and_bodies(self.batch_bodies)
         if normalized is not None:
             self.batch_bodies = normalized
-        self._per_message_keys = keys
 
     @classmethod
     def from_cmd(
@@ -154,17 +154,13 @@ class KafkaPublishCommand(BatchPublishCommand):
 
     def _align_keys(self, value: Sequence["Any"]) -> None:
         """Align the per-message keys with the batch_bodies."""
+        if not self._per_message_keys:
+            return
         self._per_message_keys = realign_keys(
             self._per_message_keys, self.batch_bodies, value
         )
 
-    def _update_bodies(self, bodies_seen: dict[int, Any], index: int, body: Any) -> None:
-        """Update the bodies_seen dictionary with the given index and body."""
-        if self.batch_bodies[index] == body and bodies_seen.get(index) is None:
-            bodies_seen.update({index: body})
-        else:
-            index += 1
-            self._update_bodies(bodies_seen, index, body)
+
 
 
 # Semantic alias for publish operations
--- a/faststream/confluent/response.py
+++ b/faststream/confluent/response.py
@@ -10,6 +10,7 @@ from faststream.response.response import (
     Response,
     extract_per_message_keys_and_bodies,
     key_for_index,
+    realign_keys,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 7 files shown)


ApusBerliozi:
@aradng, will mention you here as well, in case there is anything you want to change

Lancetnik (dismissed):

Lancetnik (approved):

PR #3010, opened 2026-08-11
fix: skip ack action on handler cancellation

### LINKED ISSUE #3000: Suggestion: handle framework-driven cancellation separately from handler failures
## Summary

When `graceful_timeout` expires while a non-concurrent Kafka handler is still running, FastStream cancels the consume-loop task. The acknowledgement middleware treats the resulting `CancelledError` as a handler failure. With `AckPolicy.REJECT_ON_ERROR`, `reject()` delegates to `ack()`, which calls `consumer.commit()`. Therefore the message will not be delivered again. This violates at-least-once processing semantics: FastStream marks the record as consumed before its processing has completed fully (successfully or not).

In the reproduction below, processing of a record published at offset `0` is interrupted while the handler is sleeping, but the consumer group's committed offset advances from unset to `1`. The same was reproduced with the Confluent integration.

## aiokafka reproduction

Environment used:

- FastStream `0.7.1`, current `main` at `5c392e2b9e4dfb9b546e22c0dc8aa7af9e020231`
- aiokafka `0.13.0`
- Python `3.12.3`
- Kafka broker at `127.0.0.1:9092`, with topic auto-creation (the broker image was `confluentinc/cp-kafka:8.0.0`; the FastStream client in this reproduction is aiokafka)
- one subscriber, no `max_workers`

```python
import asyncio
import uuid

from aiokafka import AIOKafkaConsumer
from aiokafka.structs import TopicPartition

BOOTSTRAP = "127.0.0.1:9092"
TOPIC = f"fs-cancelled-ack-{uuid.uuid4().hex}"
GROUP = f"fs-cancelled-ack-{uuid.uuid4().hex}"

started = asyncio.Event()
broker = KafkaBroker(BOOTSTRAP, graceful_timeout=2.0)


@broker.subscriber(
    TOPIC,
    group_id=GROUP,
    ack_policy=AckPolicy.REJECT_ON_ERROR,
)
async def handle(_: str) -> None:
    print("HANDLER_STARTED", flush=True)
    started.set()
    await asyncio.sleep(60)
    print("HANDLER_FINISHED", flush=True)


async def committed() -> int | None:
    consumer = AIOKafkaConsumer(
        bootstrap_servers=BOOTSTRAP,
        group_id=GROUP,
        enable_auto_commit=False,
    )
    await consumer.start()
    try:
        return await consumer.committed(TopicPartition(TOPIC, 0))
    finally:
        await consumer.stop()


async def main() -> None:
    print(f"COMMITTED_BEFORE={await committed()}")
    await broker.start()
    metadata = await broker.publish("one", TOPIC)
    print(f"PUBLISHED_OFFSET={metadata.offset}")
    await asyncio.wait_for(started.wait(), timeout=20)
    await broker.stop()
    print(f"COMMITTED_AFTER={await committed()}")


asyncio.run(main())
```

Observed:

```text
COMMITTED_BEFORE=None
PUBLISHED_OFFSET=0
HANDLER_STARTED
...
ERROR ... CancelledError:
Traceback (most recent call last):
  ...
  File "repro_aiokafka_cancelled_ack.py", line 48, in handle
    await asyncio.sleep(60)
asyncio.exceptions.CancelledError
...
COMMITTED_AFTER=1
```

In both integrations, restarting the same consumer group skips that message.

## Expected behavior

Cancellation caused by FastStream's own graceful-shutdown timeout should not be classified as handler failure and should not apply the normal error ack policy. The offset should remain uncommitted so Kafka can redeliver the interrupted record.

Shutdown cancellation should also not produce an ERROR-level handler traceback for each in-flight message.

## Broker-independent concern

This is not specific to Kafka or FastStream's built-in brokers. Framework-driven cancellation means processing was interrupted; it is not a handler failure. Applying the same acknowledgement policy to both is problematic for downstream broker integrations, where `reject` or `nack` may render the message as unretryable. This would prevent the broker from reliably preserving its delivery guarantees. Cancellation should therefore be handled separately at the common acknowledgement-policy layer.

Arseniy-Popov:
# Description

When a handler is cancelled (for example after `graceful_timeout`), acknowledgement middleware skips `ack` / `nack` / `reject` so the message stays unacknowledged and can be redelivered. Covered by shared consume tests and an aiokafka restart check.

Fixes #3000 

## Type of change

## Checklist

### DIFF
--- a/tests/brokers/base/consume.py
+++ b/tests/brokers/base/consume.py
@@ -1,12 +1,14 @@
 import asyncio
-from unittest.mock import MagicMock, call
+from unittest.mock import MagicMock, call, patch
 
 import anyio
 import pytest
 from pydantic import BaseModel
 
-from faststream import Context, Depends, FastStream, TestApp
+from faststream import AckPolicy, Context, Depends, FastStream, TestApp
 from faststream.exceptions import StopConsume
+from faststream.message import StreamMessage
+from tests.tools import spy_decorator
 
 from .basic import BaseTestcaseConfig
 
@@ -403,6 +405,80 @@ class BrokerConsumeTestcase(MultibrokerTestcase, BaseTestcaseConfig):
 
             with pytest.raises(AssertionError):
                 await subscriber.get_one(timeout=1e-24)
+
+    @pytest.mark.parametrize(
+        "ack_policy",
+        (
+            pytest.param(AckPolicy.ACK, id="ack"),
+            pytest.param(
+                AckPolicy.REJECT_ON_ERROR,
+                id="reject_on_error",
+                marks=[
+                    pytest.mark.filterwarnings(
+                        "ignore:AckPolicy.REJECT_ON_ERROR has the same effect"
+                    )
+                ],
+            ),
+            pytest.param(AckPolicy.NACK_ON_ERROR, id="nack_on_error"),
+        ),
+    )
+    async def test_consume_cancel_skips_ack_nack_reject(
+        self,
+        queue: str,
+        ack_policy: AckPolicy,
+    ) -> None:
+        if not self.supports_cancel_ack_skip:
+            pytest.skip("broker default subscriber has no acknowledgement middleware")
+
+        started = asyncio.Event()
+        broker = self.get_broker(graceful_timeout=0.2)
+        args, kwargs = self.get_subscriber_params(
+            queue,
+            ack_policy=ack_policy,
+            **self.get_cancel_ack_subscriber_kwargs(queue),
+        )
+
+        @broker.subscriber(*args, **kwargs)
+        async def handler(_msg: object) -> None:
+            started.set()
+            await asyncio.sleep(60)
+
+        with (
+            patch.object(
+                StreamMessage,
+                "nack",
+                spy_decorator(StreamMessage.nack),
+            ) as nack,
+            patch.object(
+                StreamMessage,
+                "ack",
+                spy_decorator(StreamMessage.ack),
+            ) as ack,
+            patch.object(
+                StreamMessage,
+                "reject",
+                spy_decorator(StreamMessage.reject),
+            ) as reject,
+        ):
+            async with self.patch_broker(broker) as br:
+                await br.start()
+                publish_task = asyncio.create_task(br.publish("hello", queue))
+                await asyncio.wait_for(started.wait(), timeout=self.timeout)
+
+                # TestBroker awaits the handler inside publish; real brokers
+                # return after enqueue, so cancel publish when still running,
+                # otherwise stop and let graceful_timeout cancel the consume task.
+                if not publish_task.done():
+                    publish_task.cancel()
+                    with pytest.raises(asyncio.CancelledError):
+                        await publish_task
+                else:
+                    await publish_task
+                    await br.stop()
+
+        nack.mock.assert_not_awaited()
+        ack.mock.assert_not_awaited()
+        reject.mock.assert_not_awaited()
 
 
 @pytest.mark.asyncio()
--- a/tests/brokers/base/consume.py
+++ b/tests/brokers/base/consume.py
@@ -1,4 +1,5 @@
 import asyncio
+from typing import Any
 from unittest.mock import MagicMock, call, patch
 
 import anyio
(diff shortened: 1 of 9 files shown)


tests/brokers/base/consume.py
```diff
@@ -404,6 +406,80 @@ async def t() -> None: ...
...
+    )
+    async def test_consume_cancel_skips_ack_nack_reject(
+        self,
+        queue: str,
+        ack_policy: AckPolicy,
+    ) -> None:
+        if not self.supports_cancel_ack_skip:
+            pytest.skip("broker default subscriber has no acknowledgement middleware")
+
+        started = asyncio.Event()
```
IvanKirpichnikov: use the `event` pytest fixture to create `asyncio.Event`

tests/brokers/base/consume.py
```diff
@@ -404,6 +406,80 @@ async def t() -> None: ...
...
+        broker = self.get_broker(graceful_timeout=0.2)
+        args, kwargs = self.get_subscriber_params(
+            queue,
+            ack_policy=ack_policy,
+            **self.get_cancel_ack_subscriber_kwargs(queue),
+        )
+
+        @broker.subscriber(*args, **kwargs)
+        async def handler(_msg: object) -> None:
```
IvanKirpichnikov: `_msg: Any`

tests/brokers/base/consume.py
```diff
@@ -404,6 +406,80 @@ async def t() -> None: ...
...
+                # TestBroker awaits the handler inside publish; real brokers
+                # return after enqueue, so cancel publish when still running,
+                # otherwise stop and let graceful_timeout cancel the consume task.
+                if not publish_task.done():
+                    publish_task.cancel()
+                    with pytest.raises(asyncio.CancelledError):
+                        await publish_task
+                else:
+                    await publish_task
+                    await br.stop()
```
IvanKirpichnikov: `await br.stop()`
There's no need for that.

tests/brokers/kafka/test_cancellation.py
```diff
@@ -0,0 +1,58 @@
...
+)
+class TestKafkaCancellation(KafkaTestcaseConfig):
+    async def test_graceful_shutdown_redelivers_interrupted_message(
+        self,
+        queue: str,
+    ) -> None:
+        group_id = f"{queue}-group"
+        payload = f"{queue}-payload"
+        started = asyncio.Event()
+        redelivered = asyncio.Event()
```
IvanKirpichnikov: use `event` and `event2` fixtures to create `asyncio.Event`

```py
started, redelivered = event, event2
```

tests/brokers/kafka/test_cancellation.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+        async def handle(_: str) -> None:
+            started.set()
+            await asyncio.sleep(60)
+
+        async with self.patch_broker(broker) as br:
+            await br.start()
+            await br.publish(payload, queue)
+            await asyncio.wait_for(started.wait(), timeout=self.timeout)
+            await br.stop()
```
IvanKirpichnikov: `await br.stop()`
There's no need for that.

tests/brokers/kafka/test_cancellation.py
```diff
@@ -0,0 +1,58 @@
...
+        restart_broker = self.get_broker()
+
+        @restart_broker.subscriber(
+            queue,
+            group_id=group_id,
+            ack_policy=AckPolicy.REJECT_ON_ERROR,
+            auto_offset_reset="earliest",
+        )
+        async def handle_retry(msg: str) -> None:
+            assert msg == payload
```
IvanKirpichnikov: Let's check that the msg contains the payload via mock.

Use pytest fixture `mock` to create `MagicMock`

```py
 @restart_broker.subscriber(
     queue,
     group_id=group_id,
     ack_policy=AckPolicy.REJECT_ON_ERROR,
     auto_offset_reset="earliest",
)
async def handle_retry(msg: str) -> None:
     mock(msg)
     redelivered.set()

async with self.patch_broker(restart_broker) as br:
     await br.start()
     await asyncio.wait_for(redelivered.wait(), timeout=self.timeout)

mock.assert_called_once_with(payload)
```

IvanKirpichnikov (commented):
Hi, thanks for the Pull Request!

tests/brokers/kafka/test_cancellation.py
Arseniy-Popov: Looks uglier, but fine.

Arseniy-Popov:
@IvanKirpichnikov Applied the changes.

IvanKirpichnikov (approved):

borisalekseev (approved):

PR #3011, opened 2026-08-12
docs: add RabbitMQ batch subscriber How-To

### LINKED ISSUE #1975: Rabbit Batch Subscriber approach
I am looking to migrate from celery to faststream and I am just wondering what would be the best way to archive a Rabbit Batch Subscriber similar to https://celery-batches.readthedocs.io/en/latest/how_it_works.html?

StepanKatin:
# Description

Add a batch subscriber pattern to the How-To section (issue #1975). Includes docs page, navigation update, code example, and docs tests.

Fixes #1975 

## Type of change

## Checklist

- [] Both new and existing unit tests pass successfully on my local environment by running `just test-coverage`
- [] I have ensured that static analysis tests are passing by running `just static-analysis`


Sehat1137:
I don't think this pattern is suitable even as a How-To example in the FastStream documentation.

Buffering unacknowledged RabbitMQ messages in application memory is inherently risky: it increases per-worker memory usage, weakens the broker-backed durability model while the batch is being accumulated/processed, and makes crash/shutdown handling significantly more complex because the messages remain unacked until the whole batch completes. It also encourages users to move batching semantics into process-local state, which is exactly the kind of failure-prone behavior the broker is normally meant to help avoid.

Documentation examples are likely to be copied into production code, so publishing this pattern implicitly legitimizes process-local buffering of unacked messages. If we want to address #1975, I think the documentation should instead explain why FastStream does not provide this abstraction and point users toward designs where batching is backed by durable storage / broker semantics rather than an in-memory collector.

docs/docs_src/rabbit/subscription/batch.py
```diff
@@ -0,0 +1,88 @@
...
+            self.data.append(message)
+            self.msg.append(raw_message)
+
+            if len(self.data) == 1:
+                self._timer = asyncio.create_task(self._on_timeout(logger))
+            
+            if len(self.data) >= self.batch_size:
+                self._cancel_timer()
+                should_flush = True
+        if should_flush:
```
Sehat1137: **Blocking:** the threshold check and the flush are not atomic. Once this coroutine reaches `BATCH_SIZE`, it releases `self.lock` before `_flush()` reacquires it. Another concurrent delivery can enter `add()` in between and append to the same buffers, so `BATCH_SIZE = 10` can produce a batch of 11+ messages. That directly contradicts the docs, which call `BATCH_SIZE` the maximum batch size.

Please detach/swap the current batch from the shared buffers while still holding the lock when the threshold is reached, then process that detached batch outside the lock. Also add a concurrent test around the batch boundary; the current sequential publish test cannot catch this race.

Sehat1137 (commented):
Blocking correctness issue in `BatchCollector.add()`.

StepanKatin:
Thank you for your review — both points are fair.
On the race in add(): agreed. Releasing the lock between the BATCH_SIZE check and _flush() allows another delivery to append into the same buffers, so the batch can grow past BATCH_SIZE. The correct approach is to detach/swap the current batch while still holding the lock, then process that local batch outside the lock (and cover it with a concurrent boundary test).

On whether this How-To belongs in the docs: I followed the suggestion in #1975 that FastStream should not add a special batch API, but that an example could live in How-To. After this discussion I agree that publishing an in-memory collector for unacked RabbitMQ messages is a poor fit for official docs — examples get copied into production, and this pattern weakens broker-backed durability and complicates crash handling.

I’m happy to either:

Close this PR as out of scope for in-memory batching, or
Replace it with a short page that explains why FastStream does not provide this abstraction and points users toward durable / broker-native designs instead of a process-local buffer.
@Lancetnik — which option do you prefer for #1975?

borisalekseev:
Two issues need to be addressed. First, as @Sehat1137 pointed out, the code has a race condition at the batch boundary. Second, processing and acknowledgment are non-atomic: if the process stops during the ack loop, already-processed but unacknowledged messages will be redelivered. The documentation should clearly mark idempotent batch processing as an important requirement. This does not need to be implemented in the example.

StepanKatin:
Thanks for the feedback on complexity.

I kept the class-based collector on purpose: a module-level `global` buffer is shorter, but it is also the kind of shortcut people copy into production, and shared process state becomes harder to reason about (imports, tests, multiple collectors). The class makes the in-process buffer explicit without pretending this is a library API.

I also kept the size + timeout flush path. Size-only would be simpler, but then the How-To would not show the usual production case — draining a partial batch when traffic is sparse. Dropping the timer would make the example look cleaner while quietly teaching an incomplete pattern.

What I did address from review:

- atomic detach under the lock so `BATCH_SIZE` is a hard max (plus a concurrent boundary test)
- docs warnings that unacked in-memory buffering is risky and batch handling must be **idempotent**

I’m happy to simplify further if you have a concrete direction — for example trimming demo noise (`sleep`, verbose logs) — but I’d rather not trade correctness or completeness for brevity in a pattern that docs readers may reuse.

@Lancetnik — any specific parts you’d like cut?

borisalekseev (dismissed):

borisalekseev (approved):

PR #3018, opened 2026-08-14
feat, refactor (publisher): Add `skip_none` param to publisher configuration

Oidaho:
# Description

This functionality arose during the design of the DLQ pattern for message processing in a product application at a FinTech organization. While implementing the DLQ, the non-obvious behavior of FastStream publishers came to light: `None` values are sent as empty messages — or even as tombstones. That led to the idea of the `skip_none` parameter, which controls this behavior explicitly.

## DLQ Pattern

With `FastStream v0.7.4` there are only two ways to set up a DLQ. Both examples use `faststream[kafka]`.

**First** — use `broker.publisher()` passed into `ContextRepo`:

```python
from typing import Annotated

from fastapi import FastAPI
from faststream_fastapi import FastStreamAPI

from faststream import AckPolicy, Context, ContextRepo
from faststream.kafka import KafkaBroker, KafkaMessage, KafkaRouter
from faststream.kafka.publisher import DefaultPublisher

broker = KafkaBroker(...)
fastapi = FastAPI(...)

application = FastStreamAPI(
    broker,
    application=fastapi,
    context=ContextRepo(
        {"dlq_publisher": broker.publisher("my-app-events-dead")},
    ),
)

router = KafkaRouter(prefix="my-app-events-")


@router.subscriber("input", ack_policy=AckPolicy.MANUAL)
async def handle_event(
    msg: KafkaMessage,
    dlq_publisher: Annotated[DefaultPublisher, Context()],
) -> None:

    payload = msg.raw_message.value.decode()

    try:
        InputPydanticModel.model_validate(payload)

    except ValidationError as error:
        await msg.reject()

        dead = DeadPydnticModel(details=str(error), payload=payload)
        await dlq_publisher.publish(dead)

    await msg.ack()
```

This approach is architecturally poor and awkward to use. The publisher is created imperatively from a concrete `KafkaBroker` before the routers exist, so the router is not self-contained: reusing it with another broker (tests, local development, multi-instance deployments) requires re-registering the publisher manually. It does not scale: either a single shared DLQ mixes dead letters from different domains, hurting operations (retention, alerts, incident triage), or every DLQ needs its own publisher pushed into ContextRepo. The full topic name is hardcoded and duplicates the `KafkaRouter(prefix=...)` + `@router.publisher(...)` convention, so after a prefix change the publisher silently keeps writing to the old topic. Finally, the wiring spans three places (creation, registration, usage) and is easy to break during refactoring — the error only surfaces with the first message. This is dirty.

**Second** — exploit the non-obvious behavior of `batch=True`:

```python
from faststream.kafka import KafkaMessage, KafkaRouter
from pydantic import ValidationError

router = KafkaRouter(prefix="my-app-events-")


@router.publisher("dead")
@router.subscriber("input", ack_policy=AckPolicy.MANUAL)
async def handle_event(msg: KafkaMessage) -> None | DeadPydnticModel:

    payload = msg.raw_message.value.decode()

    try:
        InputPydanticModel.model_validate(payload)

    except ValidationError as error:
        await msg.reject()
        return DeadPydnticModel(details=str(error), payload=payload)

    await msg.ack()
    return None
```

With this approach the handler never touches the publisher: on a validation error it returns a `DeadPydnticModel`, which `.publisher()` sends to the DLQ topic itself; on success it returns `None`. When the internal `batch_bodies()` method of the command class runs, the publisher receives an empty `tuple` and passes it to Kafka. It works, but relies on undocumented FastStream internals.

The upside: the DLQ topic name is assembled automatically from the router prefix and the publisher name, with no manual registration in ContextRepo.

The downside: `NoneType` objects in the batch become `b""` and are sent as empty messages alongside meaningful ones. So every successfully processed message leaves an empty record in the DLQ topic that has to be filtered out.

## Tombstones

In FastStream, sending `None` to a publisher with a `key` generates a tombstone: `key=123`, `value=None`. `skip_none` adds explicit control over this: `skip_none=False` (the default) keeps the current behavior, while `skip_none=True` stops such `None` values from being sent, so tombstones are only produced when intended. In production this matters most for State Based Architecture, where the compacted topic is the source of truth for the current state: an accidental tombstone silently deletes state from it, and after a restart consumers rebuild from incomplete data with no errors in the logs. `skip_none=True` prevents that.

## The `skip_none` parameter

Usage example:

```python
from pydantic import ValidationError

from faststream.kafka import KafkaMessage, KafkaRouter

router = KafkaRouter(prefix="my-app-events-")


@router.publisher("dead", skip_none=True)
@router.subscriber("input", ack_policy=AckPolicy.MANUAL)
async def handle_event(msg: KafkaMessage) -> None | DeadPydnticModel:

    payload = msg.raw_message.value.decode()

    try:
        InputPydanticModel.model_validate(payload)

    except ValidationError as error:
        await msg.reject()
        return DeadPydnticModel(details=str(error), payload=payload)

    await msg.ack()
    return None
```

The `skip_none` parameter gives explicit control over `Publisher` behavior — the publisher behaves exactly as its `self.specification` dictates. The default is `skip_none=False` for backward compatibility. With `skip_none=True`, `NoneType` objects are skipped in `_basic_publish` and `_basic_publish_batch`, and a debug log entry is written. The skip happens before the middleware chain, so skipped values do not trigger middlewares — a useful optimization, since no work is done for messages that will never be sent. With `skip_none=True` + `batch=True` the behavior stays intuitive: if `_publish()` receives `None`, all batch operations are skipped; `NoneType` objects inside the batch are excluded; and if the batch becomes empty, nothing is sent — just as if the handler had returned `None`.

This solution also eliminates the problems described in the [DLQ examples](#dlq-pattern).

## Breaking change

The return types of `_basic_publish` and `_basic_publish_batch` changed: both can now return `None` at any point, since `skip_none=True` may skip a message. Functionally nothing changes and existing code keeps working, but `mypy` will now complain where the return value is used directly — it can no longer be assumed to be non-`None`.

## Type of change

- [x] Breaking change (a fix or feature that would disrupt existing functionality)

## Checklist


Oidaho:
## Bug description
 
While reviewing the `skip_none` batch filtering, I noticed a bug:
`batch=True` publishers on `main` silently dropped **only the first
`None`** of a batch, while the rest of the messages were preserved. The
cause is the inherited getter, which hides a leading `None` body:
 
```python
# faststream/response/response.py [blocked] (main)
 
class PublishCommand(Response):
    @property
    def batch_bodies(self) -> tuple["Any", ...]:
        if self.body is not None:
            return (self.body,)
        return ()  # <-- a leading None body disappears
 
class BatchPublishCommand(PublishCommand):
    @property
    def batch_bodies(self) -> tuple["Any", ...]:
        return (*super().batch_bodies, *self.extra_bodies)
```
For a batch like (None, "hi") the getter returns ("hi",) — the first message is invisible to producers (and to realign_keys, which caused the key shift fixed in this PR).

## Default behaviour (skip_none=False)

With the default settings this bug already existed on `main`, so the broker published fewer records than requested:

```python
publisher = broker.publisher(queue, batch=True)  # skip_none=False by default

await publisher.publish(None, "hi")
# Expected: both records delivered (b"" and b"hi")
# Actual:   only "hi" is delivered — the first None is lost silently

# BUT: the drop applies only to the FIRST None
await publisher.publish(None, "hi", None)
# Expected: (b"", b"hi", b"")
# Actual:   (b"hi", b"")  # the trailing None survives as b""
```

This position-dependent drop is exactly the "non-obvious behavior of batch=True" that the second DLQ method relies on: a batch consisting of a lone None or is None itself publishes nothing, so no message reaches the DLQ topic. The fix intentionally preserves that trick — an empty command (body is None with no extra bodies) still returns no records; only mixed batches become consistent.

Therefore, I consider it necessary to make the fix commit within the scope of my changes.

Lancetnik:
Thanks for the very thorough write-up — the DLQ use case is a real one, and the empty-message behaviour you describe is real too. But I don't think it justifies a new parameter on every broker's publisher signature: the case is already covered by the current API, in three different ways. Let me go through them, because I think the PR description under-counts the available options.

> **Edited.** This paragraph originally claimed FastStream does not produce real Kafka tombstones. That was wrong — you were right. Corrected below; the rest of the comment is unchanged and the conclusion still holds.

First, on the tombstone motivation: it is real, not hypothetical. Publishing `None` does produce a genuine tombstone (`value=None`). `AioKafkaFastProducerImpl.publish` does it when the body is `None` **and** a key is set ([`kafka/publisher/producer.py#L113`](https://github.com/ag2ai/faststream/blob/main/faststream/kafka/publisher/producer.py#L113)); the Confluent producer does it for any `None`, keyed or not ([`confluent/publisher/producer.py#L142`](https://github.com/ag2ai/faststream/blob/main/faststream/confluent/publisher/producer.py#L142)). `test_publish_none_sends_a_real_tombstone` covers it. So the State-Based-Architecture concern in your description stands.

The batch path is the exception — and it is the one your second example relies on. `publish_batch` has no tombstone branch: every body goes through the codec, and [`encode_message`](https://github.com/ag2ai/faststream/blob/main/faststream/message/utils.py#L50-L59) turns `None` into `(b"", None)`. So a `None` inside a batch becomes an empty record rather than a tombstone, which is the junk-record problem you describe.

Both failure modes have the same cause — a `None` reaching the producer at all — and each option below stops it before that point, so both are prevented.

---

### Option 1 — a router-declared publisher, used as a closure (recommended)

The premise that a publisher has to be created from the broker and pushed through `ContextRepo` isn't accurate. A publisher can be declared on the router and simply closed over in the handler body:

```python
from faststream.kafka import KafkaRouter, KafkaMessage
from pydantic import ValidationError

router = KafkaRouter(prefix="my-app-events-")

dlq = router.publisher("dead")  # -> "my-app-events-dead"


@router.subscriber("input", ack_policy=AckPolicy.MANUAL)
async def handle_event(msg: KafkaMessage) -> None:
    payload = msg.body.decode()

    try:
        InputPydanticModel.model_validate(payload)
    except ValidationError as error:
        await msg.reject()
        await dlq.publish(DeadPydanticModel(details=str(error), payload=payload))
        return

    await msg.ack()
```

This answers every objection raised against the `ContextRepo` variant:

- The topic name is **not** hardcoded. `LogicPublisher.topic` is a property resolved lazily through `self._outer_config.prefix`, so the object created at import time still picks up the router prefix after `include_router`. Change the prefix and the DLQ topic follows.
- The router **is** self-contained and reusable with another broker, because the publisher belongs to the router, not to a `KafkaBroker` instance.
- It scales per domain — each router carries its own DLQ publisher, no shared dead-letter topic and no registry to maintain.
- The wiring lives in one place instead of three.

And crucially: because you call `.publish()` explicitly instead of returning a value through the `@publisher` decorator, **`None` is never published at all**. There is no empty record to filter, so `skip_none` has nothing to do here.

### Option 2 — raise a `HandlerException`

If you'd rather keep the `@publisher` decorator, note that publishers are invoked only after the handler returns normally ([`process_message`](https://github.com/ag2ai/faststream/blob/main/faststream/_internal/endpoint/subscriber/usecase.py#L374-L393)). Any exception short-circuits before `h.handler._publishers` is reached:

```python
from faststream.exceptions import SkipMessage


@router.publisher("dead")
@router.subscriber("input", ack_policy=AckPolicy.MANUAL)
async def handle_event(msg: KafkaMessage) -> DeadPydanticModel:
    payload = msg.body.decode()

    try:
        InputPydanticModel.model_validate(payload)
    except ValidationError as error:
        await msg.reject()
        return DeadPydanticModel(details=str(error), payload=payload)

    await msg.ack()
    raise SkipMessage
```

`SkipMessage` is an `IgnoredException`, so it isn't logged as an error and `consume()` swallows it. I verified this across `{SkipMessage, AckMessage, NackMessage, RejectMessage} x {MANUAL, ACK_FIRST, REJECT_ON_ERROR}` — nothing is published in any of the twelve combinations.

One caveat worth knowing: under `AckPolicy.MANUAL` the `AcknowledgementMiddleware` isn't installed at all, so `AckMessage` won't ack for you — which is fine here, since the example acks manually. With a non-manual policy, `raise AckMessage()` both skips the publish and acks.

I'd treat this option as the escape hatch rather than the default — it uses control flow to express "nothing to send", which reads less clearly than option 1.

### Option 3 — a `publish_scope` middleware on the router (recommended)

If you want the behaviour declaratively, and across a whole set of publishers at once, this is exactly what `publish_scope` is for. Not calling `call_next` means the producer is never reached:

```python
from faststream import BaseMiddleware
from faststream.kafka import KafkaRouter


class SkipNoneMiddleware(BaseMiddleware):
    async def publish_scope(self, call_next, cmd):
        bodies = tuple(body for body in cmd.batch_bodies if body is not None)

        if not bodies:
            return None

        cmd.batch_bodies = bodies
        return await call_next(cmd)


router = KafkaRouter(prefix="my-app-events-", middlewares=(SkipNoneMiddleware,))
```

This is functionally equivalent to `skip_none=True`, including the batch semantics you describe:

- a single `None` body is dropped before it reaches the producer;
- `None` entries inside a batch are filtered out (`BatchPublishCommand.batch_bodies` has a setter);
- a batch that becomes empty is skipped entirely.

Scoping works too. A middleware registered on the router applies only to that router's publishers — I checked that `broker.publish(None, "elsewhere")` does not trigger it. If you need finer granularity than a router, `cmd.destination` is available on the command, so you can filter by topic.

---

### Why I'm closing this

The pattern the PR is built around is already expressible today, and options 1 and 3 are the ones I'd point people to — option 1 when the DLQ send is explicit, option 3 when you want it enforced across a router. Neither needs new surface area.

Weighed against that, `skip_none` adds a parameter to every publisher factory, registrator, router, and FastAPI wrapper across six brokers, and the PR itself flags a typing break: `_basic_publish` and `_basic_publish_batch` may now return `None`, so downstream code that uses the return value stops type-checking. That's a real cost for behaviour that's reachable without it.

So I'm closing this as covered by the current API — but happy to reopen if there's a case here that none of the three options above reach. If the sharper point is really "the `@publisher` decorator turning a `None` return into an empty record is surprising", that's a fair complaint, and I'd rather address it as a documentation change plus a look at the decorator's own semantics than as a per-publisher flag.

Separately, your investigation did surface a genuine bug — a batch publisher whose handler returns `None` sends an empty batch, which crashes Kafka, Confluent, and Redis. Filed as #3056. Thanks for that, and for the depth of the write-up here.

Oidaho:
> Thanks

No, thank **you**! Actually, for the most part, the motivation for introducing skip_none came precisely from the non-obvious behavior of `@publisher()`, as well as the several issues you referenced in #3062 and #3060.

However, as is often the case in practice, I didn't have enough FastStream expertise to weigh all the pros and cons more thoroughly, which is why I went ahead and made the MR instead of using one of the approaches you suggested in your reply.

I'm really glad to have made even a small contribution to the development of this technology!

PR #3026, opened 2026-08-19
feat(confluent): add Topic schema to configure topic creation

### LINKED ISSUE #1827: Ability to configure topic params in confluent create_topics

vyhuholl:
# Description

Every topic FastStream created was hardcoded to `num_partitions=1` and `replication_factor=1`, and the only knob was the broker-wide `allow_auto_create_topics` flag.

Introduce a `Topic` schema accepted by `subscriber()` / `publisher()` alongside plain strings:

```python
    @broker.subscriber(
        Topic("orders", num_partitions=3, replication_factor=2),
        Topic("externally-managed", declare=False),
        "audit",
    )
```

`declare=False` skips the `create_topics` call for a single topic and nothing else - it does not probe the cluster and does not fail if the topic is missing, matching `JStream(declare=False)`. The broker-level `allow_auto_create_topics=False` still wins over everything.

`str` is normalised to `Topic` at the subscriber/publisher factory level, so no `str | Topic` union leaks deeper into the framework. Publishers accept a `Topic` for symmetry, but FastStream never creates publisher topics, so only its name is used.

Also fixes `LogicSubscriber.topic_names` applying the router prefix twice.

Fixes #1827

## Type of change

## Checklist


Lancetnik:
Thanks for the work here — the feature itself looks good, and I pushed a small review commit to the branch (conflicting-declaration warning in the subscriber factory, a `RedisPublisher` signature fix, regression tests for the double-prefix bugfix, test cleanup, docstring). Two more things I'd like to see before merge.

### 1. `declare` is not reachable for partition-assigned subscribers

`TopicPartition` has no `declare` flag, and `topics_to_create` gives partition topics the default one:

```python
for p in self.partitions:
    topics.setdefault(p.topic, Topic(p.topic))

return [t for t in topics.values() if t.declare]
```

`Topic(p.topic)` is `declare=True`, so the filter on the last line never applies to them.

That matters because `topics` and `partitions` are mutually exclusive (`create_subscriber` raises ``"You can't provide both `topics` and `partitions`."``). A subscriber that assigns partitions manually therefore has **no** way to opt out of topic creation short of the broker-wide `allow_auto_create_topics=False`. And that's exactly the case where the opt-out matters most: if you're assigning partition 3 by hand, you expect the topic to already exist with the right partition count, and you definitely don't want FastStream silently creating it with `num_partitions=1`.

This isn't a regression — pre-PR the code created every partition topic too — but the PR introduces `declare` as the public opt-out and half the API doesn't get it.

Could you add `declare: bool = True` to `TopicPartition`? It's `__slots__`, `__init__`, `add_prefix`, plus `Topic(p.topic, declare=p.declare)` in `topics_to_create`, and a test that a `declare=False` partition is not created.

### 2. Same feature for aiokafka

`#1827` scoped aiokafka out on the grounds that it can't create topics. That premise is only half right, and I'd like to fix it here rather than leave the two Kafka brokers diverged.

What's actually true on aiokafka 0.13.0:

* `AIOKafkaAdminClient.create_topics(new_topics, timeout_ms=None, validate_only=False)` exists, and `NewTopic(name, num_partitions, replication_factor, replica_assignments=None, topic_configs=None)` covers everything `Topic` carries today.
* `faststream/kafka` **already holds a started admin client** — `KafkaBrokerConfig._admin_client` / the `admin_client` property in `faststream/kafka/configs/broker.py`, built from `AdminClientConnectionParams` in `connect()`. Nothing in `faststream/kafka/` ever calls `create_topics` on it; the tests do it by hand.
* The part of `#1827` that *is* right: `AIOKafkaConsumer.__init__` has no `allow_auto_create_topics` equivalent (only `auto_offset_reset`, `enable_auto_commit`, `auto_commit_interval_ms`). Confluent gets that switch from librdkafka config; on aiokafka it would have to be a FastStream-side flag on the broker that simply gates the `create_topics` call.

So the work is: a `Topic` schema under `faststream/kafka/schemas/` mirroring the confluent one (with `to_aiokafka()` returning `NewTopic` instead of `to_confluent()`), accepting `str | Topic` in the registrator and publisher, and a `create_topics` call in `KafkaSubscriber.start()` (`faststream/kafka/subscriber/usecase.py:100`) gated by that flag.

One design question I don't have a strong opinion on: aiokafka's `TopicPartition` is their own namedtuple, which we re-export directly, so point 1's `declare` flag can't be bolted onto it there. Either wrap it in a FastStream schema the way confluent does, or leave partition-assigned subscribers on aiokafka without the opt-out and say so in the docstring. Your call — just make it a deliberate one.

No double-prefix fix is needed on the aiokafka side, by the way: `topic_names` there already returns `topics` without re-applying the router prefix. Confluent was the one that had drifted.

Happy to take either of these myself if you'd rather not — just say so.

Lancetnik:
Follow-up on my previous comment — I took both items off your plate, in opposite directions.

**Point 1 (`declare` for partitions): done, pushed to the branch** as `676daa47c3`. `TopicPartition` gets a keyword-only `declare: bool = True` that `add_prefix` carries and `to_confluent` ignores, and `topics_to_create` now builds `Topic(p.topic, declare=p.declare)` instead of hardcoding the default. Three unit tests for the schema plus one asserting a `declare=False` partition produces no creation request. I did not add a `connected` counterpart to `test_not_declared_topic_is_not_created` — the code path past `topics_to_create` is the same one the existing `Topic` test already covers, and I have no broker here to verify a new one against.

**Point 2 (aiokafka): out of scope for this PR.** Filed as #3067 with the findings from my previous comment — that aiokafka *can* create topics, that `faststream/kafka` already holds a started admin client nobody calls `create_topics` on, and that the missing piece is the `allow_auto_create_topics` equivalent, which has to become a FastStream-side flag. The open design question about wrapping aiokafka's `TopicPartition` namedtuple is recorded there too. Nothing for you to do here.

So this PR stays confluent-only. From my side it's ready once CI is green.

Lancetnik (approved):

PR #3029, opened 2026-08-22
feat: add pyright over the public API surface and fix what it found

Lancetnik:
Adds pyright alongside mypy, scoped to the **public API surface** rather than the whole package.

## Scope

```toml
[tool.pyright]
include = ["tests/mypy"]
pythonVersion = "3.10"
typeCheckingMode = "strict"
```

`tests/mypy` exercises the documented API the way a user does. Pyright still follows imports into `faststream/`, so a break in the framework surfaces where that API is used — but internal annotations are not policed. That keeps the signal about what users actually see, and keeps the gate from turning into a style debate about framework internals.

Wiring matches the existing checkers: a `lint`-group dependency, a `just pyright` recipe folded into `just static-analysis`, and a pre-commit hook. Worth noting for CI: mypy has no dedicated GitHub Actions job either — static analysis reaches CI through `pre-commit run --hook-stage manual --all-files` in the "Run linters" step of `pr_autoupdate.yaml`, so the hook is what makes this run there.

## What it found

**Every broker and router supplied only 2 of `Registrator`'s 3 type parameters.** `BrokerConfigType` has a default, so the omission was silent — every broker and router was typed against the base `BrokerConfig` instead of its own config. Now spelled out for kafka, confluent, nats, rabbit, redis and mqtt. Two more slipped in behind that: `RedisRouter` passed `BaseMessage` where the registrator expects `UnifyRedisDict`, and `MQTTRouter` passed the string `"Any"` as its message type.

Internal annotations that hold a broker — `app`, the ASGI factories, the AsyncAPI generators, `TestBroker`, `StreamRouter` — widen to a third `Any`. That part is not cosmetic: `BrokerUsecase` is invariant in the config parameter, so without it the concrete config types are no longer assignable and mypy fails.

**`ListBatchPublisher.publish()` required `list=`** while `ListPublisher.publish()` declares it optional. The `# type: ignore[override]` was hiding a signature users could not satisfy the documented way. Now `list: str | None = None`, and the ignore is gone.

**`SubscriberUsecase.__aiter__` was declared `async def`.** With no `yield` in its body that types as `Coroutine[..., AsyncIterator[...]]` — a shape the `async for` protocol does not accept, so the base class was by its own declaration not async-iterable. All 12 concrete implementations are async generators, and each carried a `# type: ignore[override]` to cover the mismatch. The base is a plain `def` now and all 12 ignores are gone. Since `warn_unused_ignores = true`, mypy staying green is the proof they were redundant rather than load-bearing.

**Four exported parser aliases pointed at names that do not exist.** `NatsParserType`, `ConfluentParserType`, `RabbitParserType` and `RedisParserType` are in `__all__` but were written as `ParserProto["Msg"]` and friends with no matching import, silenced by `# type: ignore[name-defined]` — so the parameter resolved to nothing for every checker and for anyone reading the annotation. Added the `TYPE_CHECKING` imports; all five aliases (Kafka included, which worked only by accident of an unrelated runtime import) are now annotated `TypeAlias`, which is also what makes ruff resolve the reference. Runtime value is unchanged — still a `ForwardRef`.

**`AdminClientConnectionParams` declared `api_version`, which `AIOKafkaAdminClient` does not accept.** This one has a runtime effect worth calling out: `filter_by_dict` selects admin kwargs from this TypedDict's annotations, so the field is precisely what forwarded `protocol_version` into a constructor that rejects it. Removing it stops the forwarding. `ConsumerConnectionParams` and `ProducerConnectionParams` are untouched — see below.

**`Tag.from_spec()` (dict branch) and the v2 ASGI route builder passed specification-level tags and external docs straight into the AsyncAPI models**, skipping the conversion the branch directly above them performs — a `ValidationError` for an `ExternalDocs` object.

Also: `KafkaInitKwargs.partitioner` was annotated with a `Partition` TypeVar that is meaningless in a TypedDict field.

**`aiokafka.ConsumerRecord` was used bare, and it is generic** (`ConsumerRecord[KT, VT]`). Under `strict` that made every derived type partially unknown — 31 errors from a single root cause. Parameterised as `ConsumerRecord[Any, Any]` rather than `[bytes, bytes]`: `KafkaBroker` exposes `key_deserializer` and `value_deserializer` typed `Callable[[bytes], Any]`, so key and value are only `bytes` when no deserializer is configured, and promising `bytes` would be wrong for everyone who sets one. Applied to the annotations that leak the parameter — the `KafkaBroker`, `KafkaRegistrator`, `KafkaRouter` and FastAPI `KafkaRouter` generic arguments, `KafkaRawMessage`, `KafkaMessage` and `KafkaParserType`.

## Left open on purpose

`ConsumerConnectionParams` and `ProducerConnectionParams` also declare `api_version`, and aiokafka 0.13's consumer and producer do not accept it either — so `KafkaBroker(protocol_version=...)` raises `TypeError` there too. Removing the field would silently drop a user's setting instead of failing, which is a different trade-off from the admin-client case, so that call is left to the maintainer.

## Verification

- `just mypy` — `Success: no issues found in 512 source files`
- `just pyright` — `0 errors, 0 warnings, 0 informations`
- `ruff check` / `ruff format --check` — clean
- full test suite against live brokers — green

One caution: `tests/prometheus/kafka/test_kafka.py::TestPrometheus::test_metrics` failed on two of four full-suite runs, with a *different* parametrisation each time, and passes in isolation. It asserts `event.is_set()` after an `asyncio.wait(timeout=...)` race with a live broker. Pre-existing load-sensitive flakiness, not related to this change.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,6 +72,15 @@ repos:
         verbose: true
         pass_filenames: false
 
+    -   id: pyright
+        name: Pyright
+        entry: just pyright
+        language: python
+        types: [python]
+        require_serial: true
+        verbose: true
+        pass_filenames: false
+
     -   id: bandit
         name: bandit
         entry: just bandit
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -208,7 +208,7 @@ disallow_any_unimported = false
 # the framework surfaces here — but internal annotations are not policed.
 include = ["tests/mypy"]
 pythonVersion = "3.10"
-typeCheckingMode = "basic"
+typeCheckingMode = "strict"
 
 [tool.pytest.ini_options]
 minversion = "7.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,7 +115,8 @@ lint = [
     "codespell==2.4.1",
     "zizmor==1.22.0",
     "mypy==1.19.1",
-    # mypy extensions
+    "pyright==1.1.407",
+    # type checkers extensions
     "types-Deprecated",
     "types-PyYAML",
     "types-setuptools",
@@ -200,6 +201,14 @@ warn_unused_ignores = true
 disallow_incomplete_defs = true
 disallow_untyped_decorators = true
 disallow_any_unimported = false
+
+[tool.pyright]
+# Public API surface only: `tests/mypy` exercises the documented API the way a
+# user does. Pyright still follows imports into `faststream/`, so breakage in
+# the framework surfaces here — but internal annotations are not policed.
+include = ["tests/mypy"]
+pythonVersion = "3.10"
+typeCheckingMode = "basic"
 
 [tool.pytest.ini_options]
 minversion = "7.0"
--- a/faststream/_internal/application.py
+++ b/faststream/_internal/application.py
@@ -61,7 +61,7 @@ T_HookReturn = TypeVar("T_HookReturn")
 class StartAbleApplication:
     def __init__(
         self,
-        *brokers: "BrokerUsecase[Any, Any]",
+        *brokers: "BrokerUsecase[Any, Any, Any]",
         specification: Optional["SpecificationFactory"] = None,
         config: Optional["FastDependsConfig"] = None,
     ) -> None:
@@ -77,13 +77,13 @@ class StartAbleApplication:
 
     def _init_setupable_(  # noqa: PLW3201
         self,
-        *brokers: "BrokerUsecase[Any, Any]",
+        *brokers: "BrokerUsecase[Any, Any, Any]",
         specification: Optional["SpecificationFactory"] = None,
         config: Optional["FastDependsConfig"] = None,
     ) -> None:
         self.config = config or FastDependsConfig()
         self.config.context.set_global("app", self)
-        self.brokers: list[BrokerUsecase[Any, Any]] = []
+        self.brokers: list[BrokerUsecase[Any, Any, Any]] = []
 
         self.schema: SpecificationFactory = specification or AsyncAPI()
 
@@ -96,20 +96,20 @@ class StartAbleApplication:
             await b.start()
 
     @property
-    def broker(self) -> Optional["BrokerUsecase[Any, Any]"]:
+    def broker(self) -> Optional["BrokerUsecase[Any, Any, Any]"]:
         return self.brokers[0] if self.brokers else None
 
     @deprecated(
         "This method is deprecated and will be removed in 0.8.0 Use `add_broker` instead."
     )
-    def set_broker(self, broker: "BrokerUsecase[Any, Any]") -> None:
+    def set_broker(self, broker: "BrokerUsecase[Any, Any, Any]") -> None:
         """Set already existed App object broker.
 
         Useful then you create/init broker in `on_startup` hook.
         """
         self.add_broker(broker)
 
-    def add_broker(self, broker: "BrokerUsecase[Any, Any]") -> None:
+    def add_broker(self, broker: "BrokerUsecase[Any, Any, Any]") -> None:
         if broker in self.brokers:
             msg = f"Broker {broker} is already added"
             raise SetupError(msg)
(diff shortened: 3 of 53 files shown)


pyproject.toml
```diff
@@ -201,6 +202,11 @@ disallow_incomplete_defs = true
 disallow_untyped_decorators = true
 disallow_any_unimported = false
 
+[tool.pyright]
+include = ["faststream", "tests/mypy"]
+pythonVersion = "3.10"
```
sobolevn, line 210: It is not correct to specify the required version. All versions must type-check on all real python runners.

.pre-commit-config.yaml
```diff
@@ -72,6 +72,15 @@ repos:
         verbose: true
         pass_filenames: false
 
+    -   id: pyright
```
sobolevn, line 75: It is better to run this in CI, because `pre-commit` envs differ from regular ones.

pyproject.toml
```diff
@@ -201,6 +202,11 @@ disallow_incomplete_defs = true
...
+typeCheckingMode = "basic"
```
sobolevn: Require strict, otherwise a lot of checks do not work

pyproject.toml
Lancetnik: The pin is deliberate here: checking against the minimum is what guarantees the floor. If the code type-checks under 3.10 semantics it cannot be using anything 3.11+, so every newer runner is covered by construction. `[tool.mypy]` right above already sets `python_version = "3.10"` for the same reason — dropping it for pyright alone would leave the two checkers disagreeing about what they verify.

I also measured it rather than assumed: with `pythonVersion = "3.14"` this config reports the same 0 errors, so a per-version matrix would add five runs for no signal today.

You're right that this is not symmetric, though. Pinning the minimum catches "used something too new"; it does not catch "used something that was removed or changed later", since typeshed guards stdlib on `sys.version_info` — a `distutils`-style removal looks fine at 3.10. On the current scope (`tests/mypy`, seven files of public-API usage) that surface is close to nil, but it's a real gap and worth revisiting if the scope ever widens.

.pre-commit-config.yaml
Lancetnik: It does run in CI — that is how every other static check in this repo reaches it. `.github/workflows/pr_autoupdate.yaml` ("Run linters") triggers on every non-draft PR, installs the `dev` group, installs `just`, and runs `pre-commit run --hook-stage manual --all-files`. mypy, bandit and semgrep have no dedicated job either; this hook is what puts them in CI.

On the env concern specifically: the hook never runs pyright itself. `entry: just pyright` goes to `just _static pyright`, which is `uv run --frozen pyright` — so the tool executes in the project's locked uv environment with the real dependency set. The virtualenv pre-commit builds because of `language: python` is not what pyright resolves imports against.

The hook is also byte-for-byte the same shape as the `mypy` hook directly above it, so this follows the existing convention rather than introducing a new path.

pyproject.toml
Lancetnik: Agreed — switching to `strict`.

For the record on what it costs here: strict reports 31 errors, all of them in `tests/mypy/kafka.py` and all from a single root cause — `aiokafka.ConsumerRecord` is generic (`ConsumerRecord[KT, VT]`) and FastStream uses it bare, so everything downstream is partially unknown.

```
21 reportUnknownArgumentType
 4 reportUnknownParameterType
 4 reportMissingTypeArgument
 2 reportUnknownMemberType
```

So it is one decision rather than 31 fixes, and it reaches the public surface — `KafkaParserType` is `ParserProto["ConsumerRecord"]`. Fixing it properly means saying what key/value types FastStream promises, which is exactly the kind of thing strict is supposed to force. Will push that.

Lancetnik: Pushed — `typeCheckingMode = "strict"`, 0 errors.

All 31 came from the one root cause, so it was a single fix. Parameterised as `ConsumerRecord[Any, Any]` rather than `[bytes, bytes]`: `KafkaBroker` exposes `key_deserializer` and `value_deserializer` typed `Callable[[bytes], Any]`, so key and value are only `bytes` when no deserializer is configured — claiming `bytes` would be wrong for everyone who sets one.

Applied to the annotations that actually leak the parameter (the `KafkaBroker` / `KafkaRegistrator` / `KafkaRouter` / FastAPI `KafkaRouter` generic arguments, `KafkaRawMessage`, `KafkaMessage`, `KafkaParserType`) and to `tests/mypy/kafka.py`, since that is what a user writing a custom parser under strict has to write too. mypy and the full suite are green.

borisalekseev (approved):

PR #3030, opened 2026-08-22
feat: Config values at the App and Broker levels

Lancetnik:
# Description

Adds `Config` — a placeholder written where an endpoint address (topic, queue, subject, stream, group, ...) would go, standing for a value that is not known at declaration time and supplied later at the Broker or the App level:

```python
@router.subscriber(Config("IN_TOPIC"))

broker = KafkaBroker(config={"IN_TOPIC": "orders"})
broker.include_router(router)
```

Addresses #2451.

**This is a draft: the resolution machinery and the `config=` entry points are in place, endpoints do not consume them yet.** `Config` placeholders are not resolved by subscribers/publishers so far — that wiring, plus the AsyncAPI story and the docs, is the remaining work.

## What is in

- `faststream/_internal/config_value.py` — `Config`, `ConfigSource`, and `ConfigResolutionMixin.resolve_option()`, the single place where precedence, defaults and the missing-key error live, so six brokers do not grow six versions of them.
- `config=` on all brokers (Kafka, RabbitMQ, NATS, Redis, MQTT) and on `FastStream` / `AsgiFastStream`.
- Sources compose most-specific-first: Broker-level values win, App-level values fill in the keys the Broker misses. Each Broker under one App resolves the same key to its own value.
- A source is a mapping *or* an arbitrary object (a settings instance), read by item access with an attribute fallback.
- A missing key with no `default` raises `SetupError`; `None` is a real default, not an omission.

## Notable internal renames

`config=` was already taken on the application constructors by the FastDepends config, so:

- `Application.__init__` / `_init_setupable_`: `config` → `fd_config` (the new `config` is the user-facing Config value source, passed down as `config_values`).
- `BrokerUsecase._update_fd_config()` → `_update_config()`, now composing a whole `BrokerConfig` rather than only the FastDepends part, via `ConfigComposition.add_outer_config()`.

Both are private, so this is not a breaking change for users.

## Type of change

## Checklist

- [x] I have added tests to validate the effectiveness of my fix or the functionality of my new feature — `tests/config/` (28 passing)


faststream/_internal/config_value.py
```diff
@@ -0,0 +1,102 @@
...
+
+from collections.abc import Iterator, Mapping
+from typing import Any, TypeAlias, TypeVar, Union, cast
+
+from faststream._internal.constants import EMPTY
+from faststream.exceptions import SetupError
+
+T = TypeVar("T")
+
+ConfigSource: TypeAlias = object
```
sobolevn, line 15: This is not correct :)

faststream/_internal/config_value.py
```diff
@@ -0,0 +1,102 @@
...
+    The one place where level precedence, defaults and the missing-key error
+    live, so that six brokers do not grow six versions of them.
+    """
+
+    @property
+    def config_sources(self) -> Iterator["ConfigSource"]:
+        """Every Config value source in scope, most specific first."""
+        raise NotImplementedError
+
+    def resolve_option(self, option: Union["Config", T]) -> T:
```
sobolevn:
```suggestion
    def resolve_option(self, option: "Config | T") -> T:
```

faststream/_internal/config_value.py
```diff
@@ -0,0 +1,102 @@
...
+    except LookupError:
+        # Item access worked and the source has nothing under that key. Falling back
+        # to attributes here would resolve `Config("items")` to `dict.items`.
+        return EMPTY
+    except TypeError:
+        # Not subscriptable: a settings object rather than a mapping.
+        return getattr(source, key, EMPTY)
+
+
+class ConfigResolutionMixin:
```
sobolevn, line 81: `__slots__ = ()` attr is missing

faststream/_internal/config_value.py
```diff
@@ -0,0 +1,102 @@
...
+
+    try:
+        return cast("Mapping[str, Any]", source)[key]
+    except LookupError:
+        # Item access worked and the source has nothing under that key. Falling back
+        # to attributes here would resolve `Config("items")` to `dict.items`.
+        return EMPTY
+    except TypeError:
+        # Not subscriptable: a settings object rather than a mapping.
+        return getattr(source, key, EMPTY)
```
sobolevn, line 78: This is error handling case is just wrong :)

Issue #2450, opened 2025-08-13
Feature: Publish messages with a pattern destination


Lancetnik:
We would like to support publishing messages with a specific pattern, such as `logs.{level}`.

Here is an example of how this could be implemented:
```python
publisher = broker.publisher("logs.{level}")
await publisher.publish("Hi!", level="info") # should be interpreted as "logs.info" destination
```
Suggested implementation:
```python
async def publish(self, message: SendableMessage, **kwargs: Any) -> None:
    dest = self.destination.format(**kwargs)
    await self._publish(message, dest)
```

As a reference we can use current `compile_path` function using for pattern-based subscribers: https://github.com/search?q=repo%3Aag2ai%2Ffaststream%20compile_path&type=code

Lancetnik:
Original issue: https://github.com/ag2ai/faststream/issues/2356

skv0zsneg:
Well, I guess this issue is mine now :)

Lancetnik:
@skv0zsneg it could be, if you really want to help :)

skv0zsneg:
I'll do it with a pleasure, this issue sounds very interesting! 
Thank you for your trust

skv0zsneg:
@Lancetnik hi again! 

I have a little problem in finding place where this feature must be implement. I'm expected some like base class and found `PublisherUsecase`. Am I right that in this class I must implement `publish` or somewhere else?

Lancetnik:
@skv0zsneg sorry, I was on vacation last week. 😄 You are right – the correct place to look is the `_basic_*` methods of the `BasicPublisher` class: https://github.com/ag2ai/faststream/blob/main/faststream/_internal/endpoint/publisher/usecase.py#L79-L108
You should just patch the `cmd.destination` line as the first line of the method.

skv0zsneg:
>  I was on vacation last week.

It's OK! I will also be unavailable for a while due to moving to another city, but will not forget about this issue :)

> You should just patch the cmd.destination line as the first line of the method.

Thanks! Will do :)

skv0zsneg:
Hello again @Lancetnik!

I return to this issue periodically and still cannot implement this feature because of different difficulties :( 
Can I make a draft PR with my work on this moment so we can solve this issue together? I really want to close this issue.

Lancetnik:
Note for whoever picks this up: #2451 (Config values at the App and Broker levels) will land before this, and it changes two of the assumptions here.

[...]

Issue #2357, opened 2025-07-13
chore: check AsyncAPI render for pattern cases in 0.6


Lancetnik:
We should validate, that all pattern usages cases are rendering correctly in 0.6 branch (for AsyncAPI 2.6 and 3.0 both)

We should validate it for NATS, RabbitMQ and Redis pattern subscribers

Also, after #2356 implementation, we should support such behavior for publishers as well

## Code example

```python
from typing import Annotated
from faststream import Path, FastStream
from faststream.nats import NatsBroker

broker = NatsBroker()

@broker.subscriber("log.{level}")
async def handler(level: Annotated[str, Path()) -> None: ...

app = FastStream(broker)
```

## Related PRs

* https://github.com/ag2ai/faststream/commit/ce705162c9312d45429b9014ba6314c6cda9633e
* https://github.com/ag2ai/faststream/pull/2349/commits/8bcfc7b2430c68967ef6d49adaeab2a077c3ea1b

Lancetnik:
@spataphore1337 have we any updates here? Do you need my help?

Lancetnik:
Looks related to https://github.com/asyncapi/asyncapi-react/issues/1071

Lancetnik:
Seems we can fix it the same time with https://github.com/ag2ai/faststream/issues/2447

PR #3032, opened 2026-08-22
refactor: split the address template from the broker address

Lancetnik:
## The problem

`logs.{level}` written at a declaration site produces **two** values:

- the **address template** — `logs.{level}`, what the developer wrote; the thing that documents the contract and that a publisher would format arguments into
- the **broker address** — `logs.*`, what is actually sent to the infrastructure broker to subscribe with

Today these two compete for one field, and the broker address wins — in three of the four brokers that support templates, by overwriting the template in place:

| Broker | Address template | Broker address |
| --- | --- | --- |
| NATS | `subject` | `clear_subject` |
| Kafka | **lost** — the compiled pattern overwrites `config.pattern` at construction | `pattern` |
| Redis | `PubSub.pattern` — but doubles as the "this is a psubscribe" flag, and `add_prefix` does not prefix it | `PubSub.name` |
| RabbitMQ | **lost** | `RabbitQueue.routing_key` |

Three independently filed pieces of work each hit this same wall: #2450 (`publisher.publish("Hi!", level="info")` needs a template left to format), #2357 (the AsyncAPI render for pattern cases differs across brokers *because* of where compilation happens), and the path-template half of #2451 (the address must be compiled after a Config value resolves, which is impossible while compilation is welded into a value object's constructor). None of the three needs the other two; all three need this.

## What changed

Every address value object exposes both values under distinct names, and every read site names which one it wants:

- **NATS** — `subject` → `subject_template`, `clear_subject` → `broker_subject`, on the subscriber, the publisher and both specifications
- **Kafka** — the compiled pattern no longer overwrites `config.pattern`; `pattern_template` and `broker_pattern` are separate reads, and the shared `compile_kafka_pattern` helper replaces the two duplicated `compile_path` calls
- **Redis** — `PubSub.pattern` becomes purely the "subscribe with a pattern" flag it was already declared as (`pattern: bool` in the constructor), and the template moves to `channel_template`
- **RabbitMQ** — `RabbitQueue` retains the routing key it was given in `routing_key_template`, read through `routing_template()` alongside the existing `routing()`

A Router prefix now reaches the address template as well as the broker address on all four — Redis previously prefixed only the latter, so a Router prefix silently dropped off the template.

No compilation moves and no timing changes here; this is a prefactor, in the same spirit as the RabbitMQ read-layer change.

## Naming convention

- address template → `<noun>_template` (`subject_template`, `pattern_template`, `channel_template`, `routing_template()`)
- broker address → `broker_<noun>` where the read layer owns the name (`broker_subject`, `broker_pattern`), or the value object's established public name where one exists (`PubSub.name`, `RabbitQueue.routing()`)

## Note on public surface

`PubSub.pattern` changes type from `str | None` to `bool`. Every internal read already used it as a truthiness flag; the one place that used its string value (`TestRedisBroker`'s fake message) now reads `channel_template`.

## Deliberately not in scope

A publisher's read stays on the broker address. Redis keeps sending to `logs.*`, and is not aligned with the other three even though that value is self-consistent only by accident: a `psubscribe logs.*` subscriber does match a channel literally named `logs.*`, so the round trip passes FastStream's own tests and only breaks against a real `logs.info`. Aligning it belongs to #2450, which gives the template a meaning, rather than to a prefactor that is supposed to change no behaviour at all.

Also out of scope: any Config handling, compilation-timing change, or AsyncAPI output change.

## Verification

- Existing suites for all four brokers pass untouched — `4005 passed, 32 skipped, 7 xfailed` (non-connected)
- `mypy`: no issues in 505 source files; `ruff check` and `ruff format` clean
- New `test_address_template.py` per broker covers both reads and the Router prefix reaching each


Lancetnik:
Reworked after review: the two values are now held by **one type** rather than two flat reads per endpoint.

`Address` (`faststream/_internal/utils/path.py`) exposes `.template`, `.broker_address` and `.regex`. Every broker's address read answers with one:

| Broker | Read |
| --- | --- |
| NATS | `subscriber.subject` / `publisher.subject` |
| Kafka | `subscriber.pattern` (`None` when none was declared) |
| Redis | `PubSub.address`, with `name` mirroring `.broker_address` for `NameRequired` |
| RabbitMQ | `RabbitQueue.routing_address`, read through `routing()` / `routing_template()` |

The per-broker wildcard spelling becomes one `AddressSyntax` constant each — which is what `compile_path`'s `replace_symbol` + `patch_regex` pair had been travelling as at every call site. `Address.add_prefix` decorates the template and lets the broker address follow, so the Router-prefix requirement holds in one place instead of four.

Compilation stays lazy and is cached per instance: a template can still be decorated after the object is built, which is what #2451 needs in order to resolve a Config value *before* the address compiles.

**Why the change.** The previous revision added `<noun>_template` / `broker_<noun>` pairs to each endpoint. Review found the two values were a data clump — always born, prefixed and read together, spelled four different ways — and that "consistently named reads" was therefore only half-delivered.

**Also addressed from review**

- The four duplicated per-broker test files collapse into `tests/brokers/base/address.py`, inherited by all four, per the repo's testing-patterns rule; `tests/utils/test_path.py` pins the type's own contract.
- `test_pattern_template_renders_as_declared` now covers Kafka's AsyncAPI render of a `{}` template — the existing `test_pattern` uses a literal `events.*` and could not catch the flip.
- NATS's `broker_subject`, duplicated across publisher and subscriber, is gone.
- `compile_kafka_pattern`'s pair-return (each half discarded by a different caller) is gone, along with the per-read recompile.
- The ticket's table wrongly credited Kafka with compiling lazily on read — true on `feat/2451`, not on `main`. Amended in the ticket rather than papered over; Kafka needed the same retain-the-template change as RabbitMQ.

**Public surface.** `PubSub.pattern` changes `str | None` → `bool`. `RabbitQueue.routing_key` and `PubSub.path_regex` become derived properties rather than stored fields, keeping their types. NATS's `subject` and Kafka's `pattern` now answer with an `Address` rather than a `str`.

**Knowingly left.** `TestRedisBroker`'s fake `PubSubMessage.pattern` keeps reporting the template, as before the split; real Redis reports the pattern it was psubscribed with. Nothing reads it except for truthiness, and aligning it belongs with #2450, which gives the template a meaning — noted in a comment at the site.

**Verification.** `4020 passed, 32 skipped, 7 xfailed` (non-connected), no existing test modified; `mypy` clean over 505 files; `ruff check`/`format` clean.

borisalekseev (approved):

faststream/_internal/utils/path.py
```diff
@@ -8,6 +9,76 @@
 PARAM_REGEX = re.compile(r"{([a-zA-Z0-9_]+)}")
 
 
+@dataclass(frozen=True)
```
sobolevn, line 12:
```suggestion
@dataclass(frozen=True, slots=True)
```

faststream/kafka/subscriber/usecase.py
```diff
@@ -31,6 +31,12 @@
     from .config import KafkaSubscriberConfig
 
 
+KAFKA_ADDRESS_SYNTAX = AddressSyntax(
```
sobolevn, line 34: Do you use `Final` for constants?

faststream/rabbit/schemas/queue.py
```diff
@@ -84,17 +91,30 @@ def _hash_dict(d: Any) -> Any:
             ),
         )
 
+    @property
+    def routing_key(self) -> str:
+        """The routing key as it reaches RabbitMQ."""
+        return self.routing_address.broker_address
+
+    @property
+    def path_regex(self) -> Optional["Pattern[str]"]:
```
sobolevn, line 100:
```suggestion
    def path_regex(self) -> "Pattern[str] | None":
```

PR #3038, opened 2026-08-22
feat: the 3.0 `address` field carries the address (#2357)

Lancetnik:
`channels[].address` carries the address the endpoint reads or writes. It is **not** the channel key: the key is a name and carries the handler with it, so the two differ for every subscriber. Before this PR the key was copied into the field, which produced `address: "test:Handle"` — a string nobody publishes to.

Only 3.0 has the field. 2.6's channel object has no such attribute and its key *is* the name; putting an address there would collide two handlers listening on one address.

**Absent, not empty.** Where there is no address the field is dropped entirely, because AsyncAPI reads a missing field as *unknown* and `""` does not say that.

Every row below is a rendered schema, not a reading of the source.

## Kafka

The topic — or the pattern, where one is given.

| Declared as | Channel key | `address` |
|---|---|---|
| `subscriber("orders")` | `orders:Handle` | `orders` |
| `subscriber("orders", "events")` | `orders:Handle` · `events:Handle` | `orders` · `events` |
| `subscriber("orders", batch=True)` | `orders:Handle` | `orders` |
| `subscriber(partitions=[TopicPartition(…)])` | `orders:Handle` | `orders` |
| `subscriber(pattern="logs.{level}")` | `logs.{level}:Handle` | `logs.{level}` |
| `publisher("orders")` | `orders:Publisher` | `orders` |
| `publisher("orders", batch=True)` | `orders:Publisher` | `orders` |

One channel per topic, each with its own address. A partition contributes its topic; the partition number is not part of an address. `pattern=` is the one Kafka argument that is compiled.

## Confluent

As Kafka, minus `pattern=`, which this package has not got.

| Declared as | Channel key | `address` |
|---|---|---|
| `subscriber("orders")` | `orders:Handle` | `orders` |
| `subscriber("orders", "events")` | `orders:Handle` · `events:Handle` | `orders` · `events` |
| `subscriber("orders", batch=True)` | `orders:Handle` | `orders` |
| `subscriber(partitions=[TopicPartition(…)])` | `orders:Handle` | `orders` |
| `publisher("orders")` | `orders:Publisher` | `orders` |
| `publisher("orders", batch=True)` | `orders:Publisher` | `orders` |

## NATS

The subject — unless the consumer filters instead of subscribing.

| Declared as | Channel key | `address` |
|---|---|---|
| `subscriber("logs.{level}")` | `logs.{level}:Handle` | `logs.{level}` |
| `subscriber(…, stream=JStream(…))` | `logs.{level}:Handle` | `logs.{level}` |
| `subscriber(…, pull_sub=PullSub())` | `logs.a:Handle` | `logs.a` |
| `filter_subjects=["logs.a"]` | `logs.a:Handle` | `logs.a` |
| `filter_subjects=["logs.a", "logs.b"]` | `logs.a, logs.b:Handle` | **absent** |
| `kv_watch=KvWatch(…)` | — | **no channel** |
| `obj_watch=ObjWatch()` | — | **no channel** |
| `publisher("logs.{level}")` | `logs.{level}:Publisher` | `logs.{level}` |

A JetStream consumer with no `subject` reaches its stream through `filter_subjects`, and a single one of those is the address. Several are not: no one string is the address, so the document gives none.

`kv_watch` and `obj_watch` resolve to `NotIncludeSpecifation` — `include_in_schema` is `False` and `get_schema()` raises — so they produce no channel and the address question does not arise. Pre-existing, not touched here.

## Redis

The channel, list or stream name — whichever one was declared.

| Declared as | Channel key | `address` |
|---|---|---|
| `subscriber("chan")` | `chan:Handle` | `chan` |
| `subscriber(PubSub("logs.*", pattern=True))` | `logs.*:Handle` | `logs.*` |
| `subscriber(list=ListSub("mylist"))` | `mylist:Handle` | `mylist` |
| `subscriber(list=ListSub(…, batch=True))` | `mylist:Handle` | `mylist` |
| `subscriber(stream=StreamSub("mystream"))` | `mystream:Handle` | `mystream` |
| `subscriber(stream=StreamSub(…, group=…))` | `mystream:Handle` | `mystream` |
| `publisher("chan")` | `chan:Publisher` | `chan` |
| `publisher(list="mylist")` | `mylist:Publisher` | `mylist` |
| `publisher(stream="mystream")` | `mystream:Publisher` | `mystream` |

A list key and a stream name are literals, never compiled. A consumer group is a reading detail, not part of the address.

## RabbitMQ

The routing key — never the queue name, and nothing at all where the exchange ignores keys.

| Declared as | Channel key | `address` |
|---|---|---|
| `subscriber("q")` (default exchange) | `q:_:Handle` | `q` |
| `RabbitQueue("q", routing_key="logs.{level}")` + topic | `q:ex:Handle` | `logs.{level}` |
| `RabbitQueue("q", routing_key="rk")` + direct | `q:ex:Handle` | `rk` |
| `RabbitQueue("q")` + topic | `q:ex:Handle` | `q` |
| `RabbitQueue("q")` + fanout | `q:ex:Handle` | **absent** |
| `RabbitQueue("q")` + headers | `q:ex:Handle` | **absent** |
| `publisher(routing_key=…, exchange=topic)` | `logs.{level}:ex:Publisher` | `logs.{level}` |
| `publisher(RabbitQueue("q", routing_key="rk"), topic)` | `rk:ex:Publisher` | `rk` |
| `publisher(RabbitQueue("q"), fanout)` | `_:ex:Publisher` | **absent** |
| `publisher(…, fanout, routing_key="ignored")` | `ignored:ex:Publisher` | **absent** |

A queue is a place a consumer reads from; the routing key is the string a message is addressed with, and the only one of RabbitMQ's three names that can hold a `{param}`. With no key declared the queue name *is* the binding key, which is why those two rows coincide.

**The exchange decides, not the declaration.** A fanout reaches every queue bound to it and a headers exchange routes on headers, so neither has a key that routes anything — a `routing_key=` handed to a fanout still addresses nothing. Otherwise a publisher would read an address where the subscriber on the far side of it reads none. The document already drops the queue binding for a fanout; naming the queue as the address would put back exactly what that omission removed.

## MQTT

The topic a message arrives on — which is not always the topic you subscribed with.

| Declared as | Channel key | `address` |
|---|---|---|
| `subscriber("logs/{level}")` | `logs.{level}:Handle` | `logs/{level}` |
| `subscriber("test", shared="group")` | `$share.group.test:Handle` | `test` |
| `publisher("logs/{level}")` | `logs.{level}:Publisher` | `logs/{level}` |

The 3.0 channel key spells `/` as `.`, being cleaned of characters a `$ref` cannot hold; the address keeps MQTT's own separator.

`$share/<group>/` asks for a shared subscription and never appears in a message's topic, so the address does not carry it. The group is not lost from the document: the channel name keeps it, and so does `bindings.mqtt.topic`.

## Under a router prefix

The prefix joins the declaration and is compiled with it, so a `{{` of the prefix's own comes off with the rest — the rule #3074 established.

| Broker | Prefix | Declared | `address` |
|---|---|---|---|
| NATS | `app{{v1}}.` | `logs.{level}` | `app{v1}.logs.{level}` |
| Redis | `app{{v1}}.` | `chan` | `app{v1}.chan` |
| MQTT | `app{{v1}}/` | `logs/{level}` | `app{v1}/logs/{level}` |
| RabbitMQ | `app.` | `logs.{level}` | `app.logs.{level}` |
| Kafka | `app{{v1}}.` | `orders` | `app{{v1}}.orders` |
| Confluent | `app{{v1}}.` | `orders` | `app{{v1}}.orders` |

The last two keep their `{{`: a Kafka topic is a literal and never meets the parser, so it has no escape to undo. Known and filed rather than fixed here — see #3076, which also covers a restored `{v1}` being read as a channel parameter.

## 2.6 is untouched

Not one 2.6 snapshot moved. The same Redis declaration that renders `address: "chan"` under 3.0 renders `['bindings', 'publish', 'servers']` under 2.6 — there is nowhere to put an address, and the channel key is the name.

## Deferred

**A subscriber over several addresses should be several channels.** NATS renders two filtered subjects as one channel keyed `logs.a, logs.b`, which is not a subject anybody publishes to — hence the absent address above. Kafka already splits. #3070 does this for NATS and stacks on top of this PR.

**A literal brace cannot be expressed in an AsyncAPI address.** #3076.

### DIFF
--- a/tests/asyncapi/rabbit/v3_0_0/test_address_field.py
+++ b/tests/asyncapi/rabbit/v3_0_0/test_address_field.py
@@ -1,0 +1,34 @@
+import pytest
+
+from faststream.rabbit import ExchangeType, RabbitBroker, RabbitExchange, RabbitQueue
+from tests.asyncapi.base.v3_0_0.basic import get_3_0_0_schema
+
+TOPIC = RabbitExchange("test-ex", type=ExchangeType.TOPIC)
+FANOUT = RabbitExchange("test-ex", type=ExchangeType.FANOUT)
+
+
+def _addresses(broker: RabbitBroker) -> dict[str, str | None]:
+    return {
+        name: channel.get("address")
+        for name, channel in get_3_0_0_schema(broker)["channels"].items()
+    }
+
+
+@pytest.mark.rabbit()
+def test_the_routing_key_is_the_address_not_the_queue() -> None:
+    """A queue is a place to read from; the routing key is what addresses a message.
+
+    It is also the only one of RabbitMQ's three names that can hold a `{param}`,
+    so a queue called one thing and bound by another renders the second.
+    """
+    broker = RabbitBroker()
+
+    @broker.subscriber(RabbitQueue("test-q", routing_key="logs.{level}"), TOPIC)
+    async def handle_logs(body: str) -> None: ...
+
+    broker.publisher(RabbitQueue("test-q", routing_key="cache"), TOPIC)
+
+    assert _addresses(broker) == {
+        "test-q:test-ex:HandleLogs": "logs.{level}",
+        "cache:test-ex:Publisher": "cache",
+    }
--- a/tests/asyncapi/rabbit/v3_0_0/test_address_field.py
+++ b/tests/asyncapi/rabbit/v3_0_0/test_address_field.py
@@ -32,3 +32,38 @@ def test_the_routing_key_is_the_address_not_the_queue() -> None:
         "test-q:test-ex:HandleLogs": "logs.{level}",
         "cache:test-ex:Publisher": "cache",
     }
+
+
+@pytest.mark.rabbit()
+def test_a_fanout_has_no_address() -> None:
+    """A fanout reaches every queue bound to it, so no key routes a message here.
+
+    The document already drops the queue binding for this case; rendering the queue
+    name as the address would put back exactly what that omission removed. Absent
+    rather than empty: AsyncAPI reads a missing address as unknown, `""` does not.
+    """
+    broker = RabbitBroker()
+
+    @broker.subscriber(RabbitQueue("test-q"), FANOUT)
+    async def handle(body: str) -> None: ...
+
+    broker.publisher(RabbitQueue("test-q"), FANOUT)
+
+    assert _addresses(broker) == {
+        "test-q:test-ex:Handle": None,
+        "_:test-ex:Publisher": None,
+    }
+
+
+@pytest.mark.rabbit()
+def test_a_routing_key_declared_on_a_fanout_is_still_no_address() -> None:
+    """The exchange decides, not the declaration.
+
+    A fanout ignores any routing key handed to it, so one declared anyway still
+    addresses nothing — and a publisher must not read an address where the
+    subscriber on the far side of it reads none.
+    """
+    broker = RabbitBroker()
+    broker.publisher(RabbitQueue("test-q"), FANOUT, routing_key="ignored")
+
+    assert _addresses(broker) == {"ignored:test-ex:Publisher": None}
--- a/faststream/confluent/publisher/specification.py
+++ b/faststream/confluent/publisher/specification.py
@@ -26,6 +26,7 @@ class KafkaPublisherSpecification(
 
         return {
             self.name: PublisherSpec(
+                address=self.topic,
                 description=self.config.description_,
                 operation=Operation(
                     message=Message(
--- a/faststream/confluent/subscriber/specification.py
+++ b/faststream/confluent/subscriber/specification.py
@@ -37,6 +37,7 @@ class KafkaSubscriberSpecification(
             handler_name = self.config.title_ or f"{t}:{self.call_name}"
 
             channels[handler_name] = SubscriberSpec(
+                address=t,
                 description=self.description,
                 operation=Operation(
                     message=Message(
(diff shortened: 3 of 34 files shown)


borisalekseev (dismissed):

IvanKirpichnikov (dismissed):

tests/asyncapi/rabbit/v3_0_0/test_address_field.py
```diff
@@ -0,0 +1,69 @@
...
+
+TOPIC = RabbitExchange("test-ex", type=ExchangeType.TOPIC)
+FANOUT = RabbitExchange("test-ex", type=ExchangeType.FANOUT)
+
+
+def _addresses(broker: RabbitBroker) -> dict[str, str | None]:
+    return {
+        name: channel.get("address")
+        for name, channel in get_3_0_0_schema(broker)["channels"].items()
+    }
```
borisalekseev, line 14: Let's avoid using helper functions for assertion comparisons, as this will lead to less implicit behavior reading test. Maybe use [`dirty-equals`](https://github.com/samuelcolvin/dirty-equals) here?
```
assert get_3_0_0_schema(broker)["channels"] == IsPartialDict(...)
```

borisalekseev: Resolved, lets do it independant

borisalekseev (approved):

PR #3045, opened 2026-08-26
Allow MANUAL ack_policy with max_workers on Redis streams

### LINKED ISSUE #3043: Bug: Redis stream subscriber: SetupError for MANUAL ack_policy + max_workers>1 seems unnecessary
**Describe the bug**

  Declaring a Redis stream subscriber with `ack_policy=AckPolicy.MANUAL` and `max_workers > 1` raises `SetupError("Max workers not work with manual no_ack mode.")`. As far as I can tell from the source, this
  restriction has no technical basis for Redis streams.

  **How to reproduce**

  ```python

  broker = RedisBroker()

  @broker.subscriber(
      stream=StreamSub("my-stream", group="my-group", consumer="c1"),
      ack_policy=AckPolicy.MANUAL,
      max_workers=2,
  )  # SetupError: Max workers not work with manual no_ack mode.
      ...
  ```

  **Expected behavior**

  The combination should be allowed. `XACK` is per-entry and has no ordering semantics, so acking manually from concurrent handler tasks cannot conflict — unlike Kafka, where commits are cumulative offsets and the
  analogous Confluent check ("Max workers not work with manual commit mode.") guards a real hazard.

  **Observed behavior**

  `SetupError` raised by `faststream/redis/subscriber/factory.py::_validate_input_for_misconfigure`:

  ```python
  if stream and ack_policy is AckPolicy.MANUAL and max_workers > 1:
      msg = "Max workers not work with manual no_ack mode."
      raise SetupError(msg)
  ```

  Note that this check only validates the `ack_policy` argument passed directly to `subscriber()`. When `AckPolicy.MANUAL` is set as a router (or broker) default, the argument is `EMPTY` at validation time and
  resolves to MANUAL afterwards via `RedisSubscriberConfig.ack_policy` — so the same combination passes setup and actually runs:

  ```python
  router = RedisRouter(ack_policy=AckPolicy.MANUAL)

  @router.subscriber(
      stream=StreamSub("my-stream", group="my-group", consumer="c1"),
      max_workers=2,
  )  # no error; runs concurrently with MANUAL
      ...
  ```

  **Environment**

  - faststream 0.7.3
  - redis-py 8.1.0
  - Python 3.14.6
  - Linux (aarch64)

  **Additional context**

  The message wording ("manual no_ack mode") references the legacy `no_ack` flag, which hints the check was carried over from an older semantic.

00yhj22-debug:
Closes #3043.

## Why the restriction can go

`XACK` is per-entry and carries no ordering semantics, so acknowledging from concurrent handler tasks cannot conflict. The Confluent check this was modelled on (`"Max workers not work with manual commit mode."`) guards cumulative offset commits, which Redis streams have no equivalent of.

The check was also only half-applied. It inspects the `ack_policy` argument passed to `subscriber()`, but that argument is still `EMPTY` when the policy comes from a router or broker default — the `if ack_policy is not EMPTY:` block immediately below it exists for exactly that reason. So the same configuration already ran, depending only on how it was expressed:

```python
# raised SetupError
@broker.subscriber(
    stream=StreamSub("stream", group="group", consumer="consumer"),
    ack_policy=AckPolicy.MANUAL,
    max_workers=2,
)
async def handle(msg: str) -> None: ...

# reached StreamConcurrentSubscriber with AckPolicy.MANUAL and ran
router = RedisRouter(ack_policy=AckPolicy.MANUAL)

@router.subscriber(
    stream=StreamSub("stream", group="group", consumer="consumer"),
    max_workers=2,
)
async def handle(msg: str) -> None: ...
```

I confirmed both paths locally before changing anything: the first raised, the second produced a `StreamConcurrentSubscriber` whose `ack_policy` was `AckPolicy.MANUAL`.

## Tests

Two tests in `tests/brokers/redis/test_misconfigure.py`, one per path, asserting the subscriber type and the resolved policy. The direct-argument test fails on `main`; the router-default one passes there already, which is the inconsistency written down.

No existing test asserted this `SetupError`.

## One note

`max_workers` is now unused inside `_validate_input_for_misconfigure`. I left it in the signature deliberately, because of the TODO a few lines below in the same file:

```python
# TODO: raise warning if max_workers in `_validate_input_for_misconfigure`
```

Happy to drop the parameter instead if that TODO is stale.

### DIFF
--- a/faststream/redis/subscriber/factory.py
+++ b/faststream/redis/subscriber/factory.py
@@ -156,10 +156,6 @@ def _validate_input_for_misconfigure(
 ) -> None:
     validate_options(channel=channel, list=list, stream=stream)
 
-    if stream and ack_policy is AckPolicy.MANUAL and max_workers > 1:
-        msg = "Max workers not work with manual no_ack mode."
-        raise SetupError(msg)
-
     if ack_policy is not EMPTY:
         if channel:
             warnings.warn(
--- a/tests/brokers/redis/test_misconfigure.py
+++ b/tests/brokers/redis/test_misconfigure.py
@@ -1,8 +1,51 @@
 import pytest
 
+from faststream import AckPolicy
 from faststream.exceptions import SetupError
 from faststream.nats import NatsRouter
-from faststream.redis import RedisBroker, RedisRouter
+from faststream.redis import RedisBroker, RedisRouter, StreamSub
+from faststream.redis.subscriber.usecases import StreamConcurrentSubscriber
+
+
+@pytest.mark.redis()
+def test_manual_ack_with_max_workers() -> None:
+    """`XACK` is per-entry, so acking manually from concurrent tasks cannot conflict.
+
+    The combination used to raise on the direct argument only. Declaring the same
+    policy as a router default already reached `StreamConcurrentSubscriber`, because
+    `ack_policy` is still `EMPTY` when the subscriber is validated.
+    """
+    broker = RedisBroker()
+
+    @broker.subscriber(
+        stream=StreamSub("stream", group="group", consumer="consumer"),
+        ack_policy=AckPolicy.MANUAL,
+        max_workers=2,
+    )
+    async def handle(msg: str) -> None: ...
+
+    (subscriber,) = broker.subscribers
+    assert isinstance(subscriber, StreamConcurrentSubscriber)
+    assert subscriber.ack_policy is AckPolicy.MANUAL
+
+
+@pytest.mark.redis()
+def test_manual_ack_with_max_workers_via_router_default() -> None:
+    """The router-default path reaches the same subscriber, and always did."""
+    router = RedisRouter(ack_policy=AckPolicy.MANUAL)
+
+    @router.subscriber(
+        stream=StreamSub("stream", group="group", consumer="consumer"),
+        max_workers=2,
+    )
+    async def handle(msg: str) -> None: ...
+
+    broker = RedisBroker()
+    broker.include_router(router)
+
+    (subscriber,) = broker.subscribers
+    assert isinstance(subscriber, StreamConcurrentSubscriber)
+    assert subscriber.ack_policy is AckPolicy.MANUAL
 
 
 @pytest.mark.redis()


Lancetnik:
Thanks for the patch — the argument is sound and the tests are the right ones.

Two things before we can merge.

We've seen a spike of automated, agent-generated issues and pull requests on this repository over the past week, so we're tightening the bar: **we do not merge contributions from automated accounts**, and we no longer accept behavioural claims that rest on reading the source alone.

That matters here specifically because this PR *removes* a guard rather than adding one. The two new tests prove that `SetupError` is no longer raised — they do not prove the combination is actually safe at runtime.

So: could you confirm this against a real Redis on your side? Concretely — a stream subscriber with `ack_policy=AckPolicy.MANUAL` and `max_workers > 1`, several messages in flight, `XACK` called manually from concurrent handlers — and report what you observe: no entries lost, none processed twice, and the pending list draining as expected. A short log or snippet is enough.

With that confirmation we're happy to merge.

00yhj22-debug:
Thanks for the clear reply, and for saying plainly what the bar is.

Two things back, in the same spirit.

First, the disclosure you're entitled to: this account is mine and I read and stand behind everything I submit, but the patch and its tests were written with AI assistance (Claude Code). Given the policy you just described, you should have that when deciding what to do with it rather than having to infer it.

Second, the substantive point — I can't give you what you asked for. Verifying this properly needs Redis 5.0+ for streams, and I don't have a usable one on this machine: the packaged Redis available here is 3.0.504, and the Memurai install needs an elevation I couldn't complete. So I have no way to run a MANUAL-ack stream subscriber with `max_workers > 1` under real concurrency and report what actually happens to the pending list. You're right that the tests only prove `SetupError` is no longer raised, which is not the same claim as "this is safe", and that gap matters more here than usual because the PR removes a guard.

Rather than leave it open with an unmet request, I'm closing it.

The reasoning is still on the record for whoever picks it up: `max_workers` spawns concurrent handler tasks, and MANUAL ack means the handler acks its own entry, so the guard appears to be rejecting a combination the consumer-group model already accounts for — but "appears to" is exactly the standard you're declining, and correctly so. Anyone with a real Redis can run the check you described and reopen with the evidence.

Sorry for the extra review cycle.

PR #3046, opened 2026-08-26
Deregister fake subscribers when the TestBroker closes

### LINKED ISSUE #2990: Bug: TestBroker keeps only a weak reference to a re-used fake subscriber, causing random SubscriberNotFound
**Describe the bug**
TestBroker._fake_start() registers a fake subscriber for every publisher with no matching subscriber and holds the only strong reference in self._fake_subscribers. _fake_close() drops it (.clear()), but the subscriber stays registered on the broker, because Registrator._subscribers is a WeakSet and the object hasn't been collected yet.

When the next TestBroker starts while that dead-but-uncollected subscriber is still weakly reachable, create_publisher_fake_subscriber() matches it and returns is_real=True. _fake_start() then keeps no strong reference — self._fake_subscribers.append(sub) only runs in the if not is_real: branch — so the publisher's only route to a handler is a weakly-referenced object that is already garbage. The next GC drops it from the WeakSet and publish() raises SubscriberNotFound.

**How to reproduce**
attached py file, outputs:

```python
1. fake subscribers still registered after the context exited: 1
2. strong refs held by this TestRabbitBroker: 0
   fake subscribers registered on the broker: 1
3. fake subscribers registered after gc.collect(): 0
```

**Expected behavior**
patched publisher should always be publishable inside the context; not raise SubscriberNotFound depending on GC timing

**Observed behavior**
raises SubscriberNotFound 

**Environment**
Running FastStream 0.7.3 with CPython 3.14.6 on Linux

**Additional context**
possible fixes: 
- skip persistent=False subscribers when matching
- deregister fakes in _fake_close

00yhj22-debug:
Closes #2990.

## What was happening

Fake subscribers are created with `persistent=False`, which means the broker keeps them only in its `WeakSet` and the caller owns their lifetime. `_fake_close()` released ownership with `self._fake_subscribers.clear()` — but clearing the list does not unregister anything. Until the next collection the fake is still weakly reachable, so `broker.subscribers` still reports it.

A `TestBroker` starting inside that window found it through `create_publisher_fake_subscriber()` and returned `is_real=True`. `_fake_start()` only appends to `self._fake_subscribers` in the `if not is_real:` branch, so the new client kept no strong reference of its own. The next collection then removed the object from the `WeakSet`, and `publish()` raised `SubscriberNotFound` — depending on GC timing rather than on anything the test did.

## The change

Withdraw the fakes from the broker in `_fake_close()` instead of waiting for the collector.

## Reproduction

Running the script from the issue:

```python
                                                          before   after
1. fake subscribers still registered after the context exited:  1       0
2. strong refs held by this TestRabbitBroker:                   0       1
   fake subscribers registered on the broker:                   1       1
3. fake subscribers registered after gc.collect():              0       0
```

Line 2 is the failure: before the change the second client held no reference to the subscriber it was about to rely on. After, it creates its own fake and owns it.

## Tests

Added `test_fake_subscribers_deregistered_without_gc` to `BrokerTestclientTestcase`, so every broker's test client inherits it.

The existing `test_correct_clean_fake_subscribers` calls `gc.collect()` before asserting, which is exactly what hid this — the new test deliberately asserts without collecting first, then collects *inside* the second context to show the fake survives because it is properly owned.

It fails on `main` and passes with this change. The other failures in `tests/brokers/rabbit/test_test_client.py` on my machine are the `_real_` cases that need a live RabbitMQ; they fail identically before and after.

### DIFF
--- a/faststream/_internal/testing/broker.py
+++ b/faststream/_internal/testing/broker.py
@@ -240,6 +240,8 @@ class TestBroker(Generic[Broker, EnterType]):
             if getattr(p, "_fake_handler", None):
                 p.reset_test()
 
+        # Fakes are registered with `persistent=False`, so the broker holds them weakly
+        # and they outlive us until the next collection (see issue #2990).
         for sub in self._fake_subscribers:
             broker._subscribers.discard(sub)
 
--- a/faststream/_internal/testing/broker.py
+++ b/faststream/_internal/testing/broker.py
@@ -240,6 +240,14 @@ class TestBroker(Generic[Broker, EnterType]):
             if getattr(p, "_fake_handler", None):
                 p.reset_test()
 
+        # Dropping our strong reference is not enough: the fakes were registered with
+        # `persistent=False`, so the broker only holds them weakly and they stay
+        # reachable until the next collection. A TestBroker starting in that window
+        # matches one as a real subscriber, keeps no strong reference of its own, and
+        # then loses it to the collector mid-test. Withdraw them here instead.
+        for sub in self._fake_subscribers:
+            broker._subscribers.discard(sub)
+
         self._fake_subscribers.clear()
 
         for sub in broker.subscribers:
--- a/faststream/_internal/testing/broker.py
+++ b/faststream/_internal/testing/broker.py
@@ -240,11 +240,6 @@ class TestBroker(Generic[Broker, EnterType]):
             if getattr(p, "_fake_handler", None):
                 p.reset_test()
 
-        # Dropping our strong reference is not enough: the fakes were registered with
-        # `persistent=False`, so the broker only holds them weakly and they stay
-        # reachable until the next collection. A TestBroker starting in that window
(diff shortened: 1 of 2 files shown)


faststream/_internal/testing/broker.py
```diff
@@ -240,6 +240,14 @@ def _fake_close(
             if getattr(p, "_fake_handler", None):
                 p.reset_test()
 
+        # Dropping our strong reference is not enough: the fakes were registered with
+        # `persistent=False`, so the broker only holds them weakly and they stay
+        # reachable until the next collection. A TestBroker starting in that window
+        # matches one as a real subscriber, keeps no strong reference of its own, and
+        # then loses it to the collector mid-test. Withdraw them here instead.
```
borisalekseev: Please, remove this large comment. The unit test already documents this behavior.

00yhj22-debug: Removed, thanks.

Lancetnik (approved):

PR #3049, opened 2026-08-27
feat(redis): add StreamSub claim_min_idle_time (XREADGROUP CLAIM)

### LINKED ISSUE #3003: Feature: support Redis 8.4 XREADGROUP CLAIM (claim_min_idle_time) in stream subscribers
Currently, `StreamSub` treats new-message consumption and pending-message claiming as mutually exclusive modes: when `min_idle_time` is set, the subscriber polls `XAUTOCLAIM` only and never reads new messages; without it, the subscriber reads new messages via `XREADGROUP` only ([stream_subscriber.py in 0.7.2](https://github.com/ag2ai/faststream/blob/0.7.2/faststream/redis/subscriber/usecases/stream_subscriber.py#L143)). This is also how the docs describe it (the "Dedicated Claiming Worker" pattern).

In practice this means every worker that wants both behaviors has to run **two subscribers per stream** — one for new messages and one dedicated to claiming — with separate consumer names, polling intervals, and ACK handling. That is what we do in production today (a resume-analysis worker and a report-generation worker, each with a new-message subscriber plus a reclaim-only subscriber on the same consumer group).

Redis 8.4 added the `CLAIM min-idle-time` option to `XREADGROUP`, which merges the two into a single command: it first claims entries idle longer than the threshold, then reads new messages, in one round trip ([release blog](https://redis.io/blog/single-shot-reliable-consumers-with-xreadgroup-claim-in-redis-84/), [XREADGROUP docs](https://redis.io/docs/latest/commands/xreadgroup/)). redis-py already exposes it as `xreadgroup(..., claim_min_idle_time=...)`, but FastStream never passes this parameter, so the capability is unreachable from FastStream even on a Redis 8.4 server.

Expose the option on `StreamSub` (name up to you — e.g. `claim_min_idle_time`) so that a single group subscriber can recover orphaned pending messages and consume new ones in one loop:

- When set, the subscriber's `XREADGROUP` calls pass `claim_min_idle_time` through to redis-py.
- It should be opt-in and independent from the existing `min_idle_time` (which would keep its current XAUTOCLAIM-only semantics for backward compatibility), since sending `CLAIM` to a pre-8.4 server is an error.
- Reclaimed entries arrive in the same handler as new ones; `times_delivered` from the PEL still lets the handler distinguish redeliveries and apply poison-message policies.

**Feature code example**
```python

broker = RedisBroker()

# Today: two subscribers are required for the same stream.
@broker.subscriber(stream=StreamSub("orders", group="g", consumer="worker-1"))
async def handle_new(msg): ...

@broker.subscriber(stream=StreamSub(
    "orders", group="g", consumer="worker-1-reclaim",
    min_idle_time=300_000, polling_interval=5_000,
))
async def handle_reclaimed(msg): ...

# Proposed: one subscriber does both (Redis >= 8.4).
@broker.subscriber(stream=StreamSub(
    "orders", group="g", consumer="worker-1",
    claim_min_idle_time=300_000,
))
async def handle(msg): ...
```

- Keeping the current two-subscriber setup (what we run today). It works, but doubles the subscriber wiring per stream, requires manual ACK handling on the claiming side, and the claiming subscriber adds a polling loop whose worst-case recovery latency is `min_idle_time + polling_interval`.
- Bypassing FastStream and calling `redis.xreadgroup(claim_min_idle_time=...)` directly. This gives up the framework's subscriber plumbing (parsing, middlewares, ACK policy), which is most of the reason to use FastStream in the first place.

**Additional context**
- redis-py support: `claim_min_idle_time` parameter on `xreadgroup` (present in redis-py 7.x).
- Verified against a real Redis 8.4.5 server (FastStream 0.7.2, redis-py 7.4.1): with `MONITOR` capturing every command, a plain group subscriber only ever issues `XREADGROUP GROUP ... BLOCK ... STREAMS ...` (no `CLAIM`), and a `min_idle_time` subscriber only issues `XAUTOCLAIM ...` — so the 8.4 capability is unreachable even when the server supports it. `StreamSub(..., claim_min_idle_time=...)` raises `TypeError: unexpected keyword argument`.
- Verified the backward-compatibility concern empirically: calling `xreadgroup(..., claim_min_idle_time=1000)` via redis-py 7.4.1 against a Redis 7.4.10 server raises `redis.exceptions.ResponseError: syntax error` — redis-py does no server-version detection or fallback, which is why the option must be opt-in.

kon28289:
# Description

Redis 8.4 added the [`CLAIM` option](https://redis.io/docs/latest/commands/xreadgroup/#the-claim-option) to `XREADGROUP`, merging pending-message claiming and new-message consumption into a single command. FastStream currently treats the two as mutually exclusive modes (`min_idle_time` switches the subscriber to XAUTOCLAIM-only), so using both requires two subscribers per stream.

This PR exposes the option as `StreamSub(claim_min_idle_time=...)`: a single group subscriber recovers abandoned pending messages and consumes new ones in one loop. Per-entry claim metadata is exposed via `raw_message["idle_times"]` / `raw_message["delivery_counts"]`, enabling retry caps and dead-letter routing without an extra `XPENDING` call.

Fixes #3003

## Design decisions

- **Opt-in and fully backward compatible.** The kwarg is passed to redis-py conditionally, so subscribers without the option behave exactly as before — including on redis-py < 7.1.0, which does not accept the argument at all. The existing `min_idle_time` (XAUTOCLAIM) semantics are unchanged; the two options are mutually exclusive. Note they are complements, not substitutes: XAUTOCLAIM also cleans up deleted (trimmed/`XDEL`-ed) entries from the PEL, which CLAIM leaves in place (documented).
- **Fail-fast validation (`SetupError`)** for combinations that cannot work: redis-py < 7.1.0, `min_idle_time` set, missing group/consumer, `last_id != ">"`, and `no_ack`. Existing options use warnings in some unsupported combinations; `claim_min_idle_time` is stricter because Redis **silently ignores** CLAIM when the id is not `>`, disabling the explicitly requested recovery behavior without any runtime error. The `no_ack` combination is rejected because it causes infinite redelivery (claimed entries stay in the PEL and are never acknowledged — reproduced live).
- **Pre-8.4 servers reject CLAIM with a syntax error.** Instead of the task supervisor hot-restarting the subscriber (~2.8k requests/s measured), the subscriber now stops with a descriptive `StreamClaimUnsupportedError`, mirroring the existing NOGROUP handling.
- **`delivery_counts` counts *previous* deliveries** (0 = new message), matching the reply schema wording — one less than `XPENDING`'s `times_delivered`, which includes the current delivery (documented with a warning).
- **Entry parsing is centralized** in `_parse_stream_entry`; an entry unexpectedly missing CLAIM metadata raises instead of degrading silently.
- **`TestRedisBroker`** attaches new-message metadata (`[0]`) for claim-enabled subscribers so handlers written per the docs also work under the in-memory test client.
- The `type: ignore[misc]` comments at the `xreadgroup` call sites are needed because the types-redis 4.6 stubs predate the argument (redis-py itself accepts it since 7.1.0).

## Known limitation

For **claim-enabled** group reads, `get_one()` and the async iterator now keep the `>` cursor, so repeated calls keep consuming new and reclaimed messages (covered by two-message repeated-call tests for both paths). The cursor drift for plain (non-claim) group reads — `self.last_id` being overwritten with the returned message id — is pre-existing behavior on `main`, now tracked in #3050, and is left unchanged here to keep this PR scoped to the new feature.

## Measurements (local loopback, Redis 8.4.5)

- Full-reclaim sweep, PEL 10,000 with 1% eligible, `COUNT 100` both sides: XAUTOCLAIM 8.98 ms / 10 round trips vs XREADGROUP CLAIM 0.84 ms / 1 round trip (XAUTOCLAIM scans at most `count*10` entries per call, so sweeps scale with PEL size; CLAIM uses the server-side time-ordered index).
- Idle polling cost at a 5 s interval: the two-subscriber setup issues ~17,200 XAUTOCLAIM calls/day/worker/stream even when idle, and with a PEL of 10,000 the `count=1` cursor sweep effectively ignores `polling_interval` (~950 calls/s measured); a single CLAIM loop stays at its BLOCK cadence regardless of PEL size.

## Notes for reviewers

- The connected tests probe `INFO server` and skip below Redis 8.4 — the first server-version gate in the suite, modeled on the cluster capability probes. CI's `redis:alpine` is 8.x, so they run there. A suite-level `skipif` also skips the new tests (instead of failing) on redis-py < 7.1.0, which the package range still allows.
- Redis Cluster: expected to work (a single stream maps to a single node), but untested — the cluster CI image is Redis 7.
- The docs update also fixes the "Pattern 2: All Workers Can Claim" example, whose comment promised new-message consumption that `min_idle_time` alone does not provide.

## Type of change

## Checklist


Labeeb2339 (commented):
The stated `get_one()` limitation looks important enough to address before merge, or at least to track with a linked issue and an explicit scope decision. `get_one()` updates `self.last_id` to the returned message id (lines 295-335), so a second call sends that explicit id to `XREADGROUP` instead of the group cursor `>`; Redis then reads the consumer's history rather than new messages, and the CLAIM option is ignored. This makes repeated `get_one()` consumption—one of the normal uses of this API—stop after the first message. The current test deliberately calls it once, so it cannot catch the regression. Could you add a two-message repeated-call test and either preserve `>` for claim-enabled group reads or document/fix the cursor state before merging?

kon28289:
Thanks for the detailed review and for filing #3050!

Applied your suggestion: claim-enabled group reads now preserve the `>` cursor in both `get_one()` and the async iterator, and I added the two-message repeated-call tests for both paths — they fail without the fix. The plain (non-claim) group-read cursor drift is left unchanged to keep this PR scoped to the feature; it is pre-existing on `main` and now tracked in #3050. The PR description's "Known limitation" section is updated accordingly.

Labeeb2339:
@kon28289 Thanks for applying the cursor-preservation change and adding repeated-call coverage. Leaving the non-claim drift tracked in #3050 keeps #3049 focused on the claim feature. I appreciate the quick follow-up.

faststream/redis/schemas/stream_sub.py
```diff
@@ -3,6 +3,7 @@
 
 from faststream._internal.proto import NameRequired
 from faststream.exceptions import SetupError
+from faststream.redis._compat import REDIS_V710, _REDIS_VERSION
```
IvanKirpichnikov: Let's rename `_REDIS_VERSION` to `REDIS_VERSION` in `_compat.py`.

Lancetnik:
Thanks for this — and for the back-and-forth on the cursor. Agreed with how you split it: this PR stays about the claim option, and the non-claim `get_one()` drift stays tracked in #3050. That one will not hold this up.

Your PR made the use case concrete enough that I went and looked at how you actually run it, in `SW-Maestro-17th-HBB/Kkori-AI`. Two things there are worth saying out loud.

**First — the workaround this PR removes is real, and you documented it better than we did.** From your module docstring:

> 구독자는 둘이다 — 새 메시지용과 방치 메시지 회수용. `StreamSub` 에 `min_idle_time` 을 주면 XREADGROUP 대신 XAUTOCLAIM 을 도는 회수 전용 모드가 되어 갓 들어온 메시지를 못 읽으므로(실측 확인), 한 구독자가 둘 다 할 수 없어 나눈다.

Two subscribers, two consumer names, twice the config — because we made the modes mutually exclusive. That is the cost of the current design stated plainly, and it is a better argument for this change than anything in the issue.

**Second — you worked around two gaps without reporting them, so I filed them:**

- **#3064** — there is no public API for `times_delivered`, so you wrote your own `messaging/pel.py` with an `XPENDING RANGE` call. A give-up-after-N-deliveries rule is the standard shape of a consumer-group worker and should not require reaching into the PEL by hand.
- **#3065** — annotating a handler argument with `redis.asyncio.Redis` silently fails to inject and dies in validation; the working symbol is `faststream.redis.annotations.Redis`, same name, different import. You left a note about it in a docstring and aliased the import to `InjectedRedis`, which tells me it cost you time.

Neither is your problem to fix. I mention it because things you route around quietly are the ones that stay broken for everyone else — if you hit more of them, an issue is genuinely welcome, however rough.

One heads-up while you are in this area: your reclaim subscriber runs `ack_policy=AckPolicy.MANUAL`, and combining that with `max_workers > 1` currently raises `SetupError`. Given that your handler calls out to Bedrock per message, concurrency is the obvious next step and you would hit it immediately. That restriction is being discussed in #3043 / #3045 — worth watching, or weighing in, since you have the production case.

Lancetnik:
Heads-up so this is not sitting in the dark: I'll review this one myself, by the end of this week.

tests/brokers/redis/test_schemas.py
```diff
@@ -12,3 +14,80 @@ def test_stream_group() -> None:
...
+                "consumer": "consumer",
+                "no_ack": True,
+                "claim_min_idle_time": 1000,
+            },
+            id="incompatible with no_ack",
+        ),
+    ),
+)
+def test_stream_claim_min_idle_time_misconfiguration(kwargs: dict) -> None:
+    with pytest.raises(SetupError):
```
borisalekseev: Please, add match argument to fix setup error reason in these tests

tests/brokers/redis/test_stream_claim.py
```diff
@@ -0,0 +1,519 @@
...
+)
+from faststream.redis._compat import REDIS_V710
+from faststream.redis.annotations import RedisBatchStreamMessage, RedisStreamMessage
+
+from .basic import RedisTestcaseConfig
+
+pytestmark = pytest.mark.skipif(
+    not REDIS_V710,
+    reason="`claim_min_idle_time` requires redis-py 7.1.0+",
+)
```
borisalekseev: Please, add skipif directly in each test.

@Lancetnik I think we should specify both the compose and the CI broker versions and update them with Dependabot. This way, we can quickly find out about incompatible changes or failed fixes in broker updates. What do you think?

borisalekseev (changes requested):
Please, fix threads and use read_id from https://github.com/ag2ai/faststream/pull/3080

faststream/redis/subscriber/usecases/stream_subscriber.py
```diff
@@ -204,19 +288,23 @@ async def get_one(
...
         if self.stream_sub.group and self.stream_sub.consumer:
             if self.min_idle_time is None:
-                stream_message = await self._client.xreadgroup(
+                stream_message = await self._client.xreadgroup(  # type: ignore[misc]
                     groupname=self.stream_sub.group,
                     consumername=self.stream_sub.consumer,
                     streams={self.stream_sub.name: self.last_id},
                     block=math.ceil(timeout * 1000),
                     count=1,
+                    **self._claim_kwargs,
```
powersemmi: It would be great to clearly indicate the arguments here to

faststream/redis/subscriber/usecases/stream_subscriber.py
```diff
@@ -150,13 +233,14 @@ async def start(self) -> None:
...
                 ) -> Awaitable[ReadResponse]:
-                    return client.xreadgroup(
+                    return client.xreadgroup(  # type: ignore[misc]
                         groupname=stream.group,
                         consumername=stream.consumer,
                         streams={stream.name: self.read_id},
                         count=stream.max_records,
                         block=stream.polling_interval,
                         noack=stream.no_ack,
+                        **self._claim_kwargs,
```
powersemmi: It would be great to clearly indicate the arguments here

tests/brokers/redis/test_stream_claim.py
Lancetnik: Yes, we should definitely bump our CI Redis dependency up to test a new features.
But I am fine with such skips to stay able to run tests locally with any Redis version

faststream/redis/subscriber/usecases/stream_subscriber.py
powersemmi: Or at least different invariants to make for different versions of redis clearly

powersemmi (changes requested):

Lancetnik:
**Findings**

- `faststream/redis/_compat.py:9` — pre-existing: `REDIS_V720 = _REDIS_MAJOR >= 7 and _REDIS_MINOR >= 2` is `False` on redis-py 8.0.x, so we still send the deprecated `lib_name`/`lib_version` instead of `driver_info` (13 `DeprecationWarning`s in this PR's own test run); worth fixing here as `(_REDIS_MAJOR, _REDIS_MINOR) >= (7, 2)`, since you add the correct idiom two lines below.
- `tests/docs/redis/stream/test_claiming.py:29` — missing the `REDIS_V710` skipif every other new test has: `docs_src/redis/stream/claiming_unified.py` builds `StreamSub(claim_min_idle_time=...)` at import, so on redis-py < 7.1.0 (allowed by `redis>=5.0.0`) the test errors with `SetupError` instead of skipping.
- `faststream/redis/subscriber/usecases/stream_subscriber.py:295` (and `:384`) — `get_one()` / `__aiter__` send CLAIM but don't map the rejection to `StreamClaimUnsupportedError`, so on a pre-8.4 server they surface a bare `ResponseError: ERR syntax error` while `_consume` gets the descriptive one.

kon28289:
- read_id from #3080 (@borisalekseev): adopted — the claim-specific cursor guard from 53ca1da is gone, group reads now go through read_id, and the repeated-call tests assert read_id stays > (26e4e12fd). The tests are kept since #3080's coverage is non-claim only.
- REDIS_V720 (@Lancetnik): fixed as suggested, (_REDIS_MAJOR, _REDIS_MINOR) >= (7, 2) (c4c7478e6). The 13 DeprecationWarnings are gone from the test run.
- docs test skipif: added the missing REDIS_V710 gate to test_stream_claiming_unified (40ca14fda).
- error mapping in get_one() / __aiter__: extracted the _consume mapping into _raise_if_claim_rejected() and applied it to all three paths (a2acdbb7c). One wording change: the message no longer ends with "Stopping subscriber.", since that's not true for the manual-consumption paths. Added mock-based regression tests for both paths — they fail without the fix.
- match in the misconfiguration cases (@borisalekseev): each case is now pinned to its specific rejection reason (0a7e8fa66). Follow-up in 1210723a1: on redis-py < 7.1.0 the version guard raises before these checks ever run, so the match cases are version-gated there — old-version behavior stays covered by the dedicated test.
- per-test skipif (@borisalekseev): done in 7b4256671 via a named mark (requires_claim_support) attached to each test, following the tests/marks.py pattern but kept file-local since the condition is redis-specific.
- explicit arguments (@powersemmi): dropped the _claim_kwargs property in 2cb74ac92 — each call site now branches, the plain call passes only stub-known arguments (no type: ignore there anymore), and the claim-enabled call passes claim_min_idle_time= by name with the version invariant commented right at the call site.

Verified against Redis 8.4.6 (tests/brokers/redis + tests/docs/redis, cluster excluded: 499 passed) and against redis-py 7.0.0 (claim tests skip cleanly, 0 failures). mypy and ruff are clean.

faststream/redis/_compat.py
```diff
@@ -6,4 +6,7 @@
 
 _REDIS_MAJOR, _REDIS_MINOR = int(major), int(minor)
 
-REDIS_V720 = _REDIS_MAJOR >= 7 and _REDIS_MINOR >= 2
+REDIS_V720 = (_REDIS_MAJOR, _REDIS_MINOR) >= (7, 2)
+
+# `xreadgroup(claim_min_idle_time=...)` appeared in redis-py 7.1.0
```
IvanKirpichnikov: The comment is unnecessary.

tests/brokers/redis/test_schemas.py
```diff
@@ -12,3 +14,88 @@ def test_stream_group() -> None:
...
+def test_stream_claim_min_idle_time() -> None:
+    stream = StreamSub(
+        "test",
+        group="group",
+        consumer="consumer",
+        claim_min_idle_time=1000,
+    )
+
+    assert stream.claim_min_idle_time == 1000
+    assert stream.add_prefix("prefix.").claim_min_idle_time == 1000
```
IvanKirpichnikov: What is the purpose of the test?

Lancetnik:
@kon28289 heads-up: I pushed a series of commits on top of yours (maintainer edits) to bring the branch in line with our conventions, and I am happy with where it is now. I have re-requested review from @borisalekseev and @powersemmi so they can clear their earlier "changes requested".

What changed, in short:

- `stream_subscriber.py`: the six `xreadgroup` call sites collapse into one `_xreadgroup` helper that owns the `claim_min_idle_time` branch and the `StreamClaimUnsupportedError` mapping; the claim metadata tuple became a `ClaimMeta` NamedTuple; self-less helpers moved to module level. Behaviour is unchanged.
- Tests: the claim tests are now a `StreamClaimTestcase` shared by the single-node and cluster configs (so the feature is exercised against Redis Cluster in CI as well), trimmed to one test per behaviour, assertions rewritten with `dirty-equals`, no test docstrings, one `require_redis_v710` mark in `tests/marks.py`, and the in-memory docs tests lost their `connected` mark.
- CI, compose and docs pin Redis to `redis:8-alpine` everywhere, so the claim tests run on 8.x by design rather than because `latest` happened to be 8.x.

Two of those commits were amended and force-pushed, so please reset your local branch to the remote before adding anything on top:

```
git fetch origin && git reset --hard origin/feat/redis-xreadgroup-claim
```

Thanks again for the PR and for the quick turnaround on the earlier rounds.

Lancetnik (dismissed):

Lancetnik:
LGTM on my side

powersemmi (dismissed):

faststream/redis/subscriber/usecases/stream_subscriber.py
```diff
@@ -92,7 +107,10 @@ async def _consume(self, *args: Any, start_signal: "Event") -> None:
             try:
                 await self._get_msgs(*args)
 
-            except ResponseError as e:  # noqa: PERF203
+            except StreamClaimUnsupportedError:  # noqa: PERF203
+                raise
```
borisalekseev: The faststream supervisor is entering an infinite error loop. Perhaps we should exit the app here?

kon28289: Tried it against a real pre-8.4 server (Redis 7.4.10, this branch) — here is what actually happens. In a 6.3s window: 13,087 supervisor restarts, 13,086 XREADGROUP attempts reaching the server (~2,070/s), no backoff, and the traceback is printed only once — after that the loop is silent. Server-side MONITOR during the run:

```
1788835453.040266 "PING"
1788835453.040467 "XREADGROUP" "GROUP" "repro-group" "repro-consumer" "BLOCK" "100" "CLAIM" "1000" "STREAMS" "claim-loop-repro" ">"
1788835453.040714 "PING"
1788835453.040906 "XREADGROUP" "GROUP" "repro-group" "repro-consumer" "BLOCK" "100" "CLAIM" "1000" "STREAMS" "claim-loop-repro" ">"
1788835453.041148 "PING"
1788835453.041327 "XREADGROUP" "GROUP" "repro-group" "repro-consumer" "BLOCK" "100" "CLAIM" "1000" "STREAMS" "claim-loop-repro" ">"
```

So I agree: using an option the server version does not support is a configuration error, not a transient failure — retrying cannot fix it, and exiting the app is the right call.

(The existing NOGROUP → StreamGroupNotFoundError path has the same restart-loop hole — reproduced it as well, ~2,000 restarts/s after the stream is deleted. I'll open a separate issue for it rather than widen this PR.)

Lancetnik: @kon28289 thank you! I am totally fine with a new Issue

kon28289:
@Lancetnik Sorry for the churn — I meant to keep the supervisor fix as a separate issue, but pushed it here by mistake. I've reverted it.

borisalekseev (approved):

PR #3066, opened 2026-08-27
feat(redis): expose stream delivery count

### LINKED ISSUE #3064: Feature: expose PEL delivery count on Redis stream messages

There is no public way to ask how many times a Redis stream message has been delivered. `times_delivered` lives in the PEL and is exactly what you need to implement a give-up rule ("after N deliveries, stop retrying and record a failure"), which is the standard shape of a consumer-group worker.

Today users have to reach for the raw client and issue `XPENDING RANGE` themselves.

**Real usage**

A worker in production does precisely this — [`SW-Maestro-17th-HBB/Kkori-AI`](https://github.com/SW-Maestro-17th-HBB/Kkori-AI), a resume-analysis service built on `faststream[redis]` with Redis Stream consumer groups.

They wrote their own PEL module — [`worker/src/messaging/pel.py`](https://github.com/SW-Maestro-17th-HBB/Kkori-AI/blob/abd0b4af9fc6d32c77408412bb445513eeb241b9/worker/src/messaging/pel.py):

```python
async def get_delivery_count(redis: Redis, stream: str, group: str, message_id: str) -> int:
    entries = await redis.xpending_range(
        stream, group, min=message_id, max=message_id, count=1
    )
    return entries[0]["times_delivered"] if entries else 1
```

and call it from the handler, taking the connection out of the broker's private attribute — [`worker/src/main.py#L141-L142`](https://github.com/SW-Maestro-17th-HBB/Kkori-AI/blob/abd0b4af9fc6d32c77408412bb445513eeb241b9/worker/src/main.py#L141-L142):

```python
delivery_count = await _delivery_count_of(broker._connection, msg)
await _process(request, delivery_count, broker._connection)
```

`broker._connection` is private, and the same file uses the public `Redis` annotation elsewhere — so the private access is avoidable. The unavoidable part is the PEL round-trip: whatever handle you get, you still have to write the `XPENDING` call yourself.

Expose the delivery count on `RedisStreamMessage` — something like `await msg.get_delivery_count()`, resolving to the PEL's `times_delivered` and defaulting to `1` when the entry is no longer pending.

Worth noting for whoever picks this up: `XAUTOCLAIM` increments the counter as it hands the message over, so a value read right after a claim already means "this is delivery number N".

**Additional context**

This is the second finding from reading that worker's source; the other is #3065.

@IvanKirpichnikov — Redis is your area, and this one is small and self-contained. Worth a look for 1.0?

lxingy3:
# Description

Redis stream users currently need a manual exact-ID `XPENDING RANGE` query to enforce retry limits.

This adds `await RedisStreamMessage.get_delivery_count(redis, group)`. The caller supplies the Redis client and consumer group, matching the existing `ack` and `delete` message APIs; handlers can receive the client through FastStream's public `Redis` annotation. Each call reads the current PEL `times_delivered` value, returns `1` when the message has no Redis ID or is no longer pending, and propagates Redis errors.

This is intentionally separate from #3049's optional `delivery_counts` raw metadata. That field reports previous deliveries for Redis 8.4 `XREADGROUP CLAIM`, while this method reads the current PEL state for existing consumer-group flows. Batch stream messages remain unchanged, and each call adds one Redis round trip.

Fixes #3064

## Type of change

## Validation

- `pytest tests/brokers/redis/test_delivery_count.py -m connected` — 2 passed against Redis 7.0.15 using real broker subscribers and no Redis client mocks
- `pytest tests/brokers/redis -m "not connected"` — 379 passed, 6 skipped, 1 xfailed
- Ruff format and lint checks
- Mypy — 516 source files, 0 errors
- Pyright — 0 errors
- Bandit and Semgrep — 0 findings
- Changed-file pre-commit hooks, including typos and detect-secrets

The full GitHub test matrix is pending on the current head.

## Checklist

- [x] I have made the necessary documentation changes
- [x] My changes do not generate new warnings
- [x] I have added tests to validate the feature
- [ ] Both new and existing unit tests pass locally with `just test-coverage`
- [x] Static analysis passes
- [x] I have included a usage example

### DIFF
--- a/faststream/redis/message.py
+++ b/faststream/redis/message.py
@@ -12,6 +12,8 @@ from typing_extensions import NotRequired, TypedDict, override
 from faststream.message import StreamMessage as BrokerStreamMessage
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
+
     from redis.asyncio import Redis
 
     from faststream._internal.basic_types import DecodedMessage
@@ -146,7 +148,36 @@ class _RedisStreamMessageMixin(BrokerStreamMessage[_StreamMsgType]):
 
 
 class RedisStreamMessage(_RedisStreamMessageMixin[DefaultStreamMessage]):
-    pass
+    _redis: Optional["Callable[[], Redis[bytes]]"] = None
+    _group: str | None = None
+
+    def _set_delivery_count_context(
+        self,
+        redis: Optional["Callable[[], Redis[bytes]]"],
+        group: str | None,
+    ) -> None:
+        self._redis = redis
+        self._group = group
+
+    async def get_delivery_count(self) -> int:
+        """Return this message's current delivery count from the Redis PEL.
+
+        The count is queried on every call. Messages without lookup context or
+        a pending entry, including acknowledged messages, return ``1``.
+        """
+        message_ids = self.raw_message["message_ids"]
+        if self._redis is None or not self._group or not message_ids:
+            return 1
+
+        message_id = message_ids[0]
+        entries = await self._redis().xpending_range(
+            name=self.raw_message["channel"],
+            groupname=self._group,
+            min=message_id,
+            max=message_id,
+            count=1,
+        )
+        return int(entries[0]["times_delivered"]) if entries else 1
 
 
 class RedisBatchStreamMessage(_RedisStreamMessageMixin[BatchStreamMessage]):
--- a/faststream/redis/message.py
+++ b/faststream/redis/message.py
@@ -12,8 +12,6 @@ from typing_extensions import NotRequired, TypedDict, override
 from faststream.message import StreamMessage as BrokerStreamMessage
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
-
     from redis.asyncio import Redis
 
     from faststream._internal.basic_types import DecodedMessage
@@ -148,31 +146,24 @@ class _RedisStreamMessageMixin(BrokerStreamMessage[_StreamMsgType]):
 
 
 class RedisStreamMessage(_RedisStreamMessageMixin[DefaultStreamMessage]):
-    _redis: Optional["Callable[[], Redis[bytes]]"] = None
-    _group: str | None = None
-
-    def _set_delivery_count_context(
+    async def get_delivery_count(
         self,
-        redis: Optional["Callable[[], Redis[bytes]]"],
-        group: str | None,
-    ) -> None:
-        self._redis = redis
-        self._group = group
-
-    async def get_delivery_count(self) -> int:
+        redis: "Redis[bytes]",
+        group: str,
+    ) -> int:
         """Return this message's current delivery count from the Redis PEL.
 
-        The count is queried on every call. Messages without lookup context or
-        a pending entry, including acknowledged messages, return ``1``.
+        The count is queried on every call. Messages without an ID or a pending
+        entry, including acknowledged messages, return ``1``.
         """
         message_ids = self.raw_message["message_ids"]
-        if self._redis is None or not self._group or not message_ids:
+        if not message_ids:
             return 1
 
         message_id = message_ids[0]
-        entries = await self._redis().xpending_range(
+        entries = await redis.xpending_range(
             name=self.raw_message["channel"],
-            groupname=self._group,
+            groupname=group,
             min=message_id,
             max=message_id,
             count=1,
--- a/tests/brokers/redis/test_delivery_count.py
+++ b/tests/brokers/redis/test_delivery_count.py
@@ -1,0 +1,239 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+from redis.asyncio import Redis
+from redis.exceptions import ConnectionError as RedisConnectionError
+
+from faststream.redis import (
+    RedisBroker,
+    RedisStreamMessage as AnnotatedRedisStreamMessage,
+    StreamSub,
+    TestRedisBroker,
+)
+from faststream.redis.message import DefaultStreamMessage, RedisStreamMessage
+from faststream.redis.parser import BinaryMessageFormatV1
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+pytestmark = [
+    pytest.mark.asyncio,
+    pytest.mark.redis,
+]
+
+
+def make_message(
+    redis: Callable[[], Redis[bytes]] | None,
+    *,
+    group: str | None = "workers",
+    message_ids: list[bytes] | None = None,
+) -> RedisStreamMessage:
+    raw_message = DefaultStreamMessage(
+        type="stream",
+        channel="orders",
+        message_ids=[b"123-0"] if message_ids is None else message_ids,
+        data={},
+    )
+    message = RedisStreamMessage(
+        raw_message=raw_message,
+        body=b"",
+    )
+    message._set_delivery_count_context(redis=redis, group=group)
+    return message
+
+
+async def test_get_delivery_count_queries_exact_message() -> None:
+    client = AsyncMock(spec=Redis)
+    client.xpending_range = AsyncMock(return_value=[{"times_delivered": 3}])
+    get_client = MagicMock(return_value=client)
+
+    message = make_message(get_client)
+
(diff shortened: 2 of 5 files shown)


Lancetnik:
Thanks for picking up #3064 so quickly — the file set is right, and shipping it with tests and a docs page is the right shape.

Before we can look at it properly, two blockers:

1. **The CLA is not signed.** We cannot accept any contribution without it, no exceptions: https://cla-assistant.io/ag2ai/faststream?pullRequest=3066
2. **The branch is behind `main`** and the test matrix has not run on it. Rebase and push, and CI will have something to say.

If the CLA is not signed within three days we will close this. That is not a judgement on the patch — closing is cheap and reopening after signing is trivial; we just do not keep unsignable contributions in the review queue.

faststream/redis/message.py
```diff
@@ -146,7 +148,36 @@ async def delete(self, redis: Optional["Redis[bytes]"]) -> None:
...
+        self._redis = redis
+        self._group = group
+
+    async def get_delivery_count(self) -> int:
+        """Return this message's current delivery count from the Redis PEL.
+
+        The count is queried on every call. Messages without lookup context or
+        a pending entry, including acknowledged messages, return ``1``.
+        """
+        message_ids = self.raw_message["message_ids"]
```
powersemmi, line 159: Are you sure that message_ids is always in raw_message?

powersemmi (changes requested):

faststream/redis/message.py
Lancetnik: Good catch, and I think it points at the underlying problem: if a method needs to reach into `raw_message` to do its job, those fields should be lifted onto the message object itself.

The only place where touching `raw_message` is legitimate is the parser, where we translate the broker payload into our own format. Everything downstream should work with the typed fields of the object, not re-parse the raw dict.

IvanKirpichnikov: > Good catch, and I think it points at the underlying problem: if a method needs to reach into `raw_message` to do its job, those fields should be lifted onto the message object itself.
> 
> The only place where touching `raw_message` is legitimate is the parser, where we translate the broker payload into our own format. Everything downstream should work with the typed fields of the object, not re-parse the raw dict.

I think that in this case, using `raw_message` is quite acceptable, since `raw_message` is used in `_RedisStreamMessageMixin`. 

 If we follow your opinion, we’ll have to refactor `_RedisStreamMessageMixin` as well, and along with it, `RedisStreamMessage` and `RedisBatchStreamMessage`.

faststream/redis/message.py
```diff
@@ -146,7 +148,36 @@ async def delete(self, redis: Optional["Redis[bytes]"]) -> None:
...
+
+    def _set_delivery_count_context(
+        self,
+        redis: Optional["Callable[[], Redis[bytes]]"],
+        group: str | None,
+    ) -> None:
+        self._redis = redis
+        self._group = group
+
+    async def get_delivery_count(self) -> int:
```
IvanKirpichnikov: Let’s stick to the design of `ack`, `nack`, `reject`, `delete`, which takes `redis: Redis | None` and `group: str | None` as parameters.

In general, let’s create a `get_delivery_count` method that takes `redis: Redis` and `group: str` and remove `_set_delivery_count_context`

IvanKirpichnikov (changes requested):

IvanKirpichnikov (changes requested):
do it please https://github.com/ag2ai/faststream/pull/3066/changes#r3908949481

faststream/redis/message.py
lxingy3: Updated in acd7abdb: get_delivery_count now takes redis and group directly, matching the existing message APIs. I removed the parser wrapper and cached context, and updated the docs and type checks to use the public Redis annotation.

IvanKirpichnikov (changes requested):
Let's rewrite the tests in `tests/brokers/redis/test_delivery_count.py` from function-based tests to class-based tests using `RedisTestcaseConfig` and without mocks.

That is, we will need to create a subscriber with a stream

```
broker = RedisBroker()

@broker.subscriber(stream=StreamSub("stream"))
async def hello(msg: RedisMessage):
    ...
```

lxingy3:
Updated in 903e4958. The delivery-count coverage is now a connected RedisTestcaseConfig class with real broker subscribers and no Redis client mocks. It covers first delivery (1), XAUTOCLAIM redelivery (2), and the post-ack fallback (1), including Redis and RedisStreamMessage handler injection. Both tests pass locally against Redis 7.0.15.

tests/brokers/redis/test_delivery_count.py
```diff
@@ -0,0 +1,70 @@
...
+        async def handler(message: RedisStreamMessage, redis: Redis) -> None:
+            counts.append(await message.get_delivery_count(redis, group))
+            event.set()
+
+        async with self.patch_broker(broker) as br:
+            await br.start()
+            await br.publish("hello", stream=queue)
+            await asyncio.wait_for(event.wait(), timeout=self.timeout)
+
+        assert counts == [1]
```
IvanKirpichnikov: replace to?

```py
    async def test_delivery_count_for_new_message(
        self,
        queue: str,
        event: asyncio.Event,
        mock: MagicMock,
    ) -> None:
        group = f"{queue}-workers"
        broker = self.get_broker(apply_types=True)

        @broker.subscriber(stream=StreamSub(queue, group=group, consumer="worker-1"))
        async def handler(message: RedisStreamMessage, redis: Redis) -> None:
            mock(await message.get_delivery_count(redis, group))
            event.set()

        async with self.patch_broker(broker) as br:
            await br.start()
            await asyncio.wait_for(
                [
                    asyncio.create_task(br.publish("hello", stream=queue)),
                    asyncio.create_task(event.wait()),
                ],
                timeout=self.timeout)

        mock.assert_called_once_with(1)
```

tests/brokers/redis/test_delivery_count.py
```diff
@@ -0,0 +1,70 @@
...
+            await br._connection.xreadgroup(
+                groupname=group,
+                consumername="worker-1",
+                streams={queue: ">"},
+                count=1,
+            )
+            await br.start()
+            await asyncio.wait_for(event.wait(), timeout=self.timeout)
+
+        assert counts == [2, 1]
```
IvanKirpichnikov: replace to?

```
    async def test_delivery_count_after_xautoclaim_and_ack(
        self,
        queue: str,
        event: asyncio.Event,
        mock: MagicMock,
        mock2: MagicMock,
    ) -> None:
        group = f"{queue}-workers"
        broker = self.get_broker(apply_types=True)

        @broker.subscriber(
            stream=StreamSub(
                queue,
                group=group,
                consumer="worker-2",
                min_idle_time=0,
            )
        )
        async def handler(message: RedisStreamMessage, redis: Redis) -> None:
            mock(await message.get_delivery_count(redis, group))
            await message.ack(redis=redis, group=group)
            mock2(await message.get_delivery_count(redis, group))
            event.set()

        async with self.patch_broker(broker) as br:
            await br.publish("hello", stream=queue)
            await br._connection.xgroup_create(queue, group, id="0")
            await br._connection.xreadgroup(
                groupname=group,
                consumername="worker-1",
                streams={queue: ">"},
                count=1,
            )
            await br.start()
            await asyncio.wait_for(event.wait(), timeout=self.timeout)

        mock.assert_called_one_with(1)
        mock2.assert_called_one_with(2)
```

IvanKirpichnikov (changes requested):

tests/brokers/redis/test_delivery_count.py
lxingy3: Updated in 8a237ed1 to use the MagicMock fixture. I kept the publish awaited before waiting on the event because syncio.wait_for accepts a single awaitable. The real Redis test passes with the new structure.

tests/brokers/redis/test_delivery_count.py
lxingy3: Updated in 8a237ed1 with separate MagicMock fixtures. The assertions remain 2 before ack and 1 after ack: XAUTOCLAIM increments the pending-entry delivery count to 2, while the acknowledged ID is no longer pending and the API falls back to 1. The real Redis test passes.

IvanKirpichnikov (approved):

powersemmi (approved):

Issue #3036, opened 2026-08-22
feat: emit AsyncAPI channel `parameters` for address templates


Lancetnik:
Follow-up to #2357, which aligns *which string* an address template renders as. This issue is about a field we do not emit at all.

## The problem

`parameters` is a `# TODO` in both `Channel` models (`faststream/specification/asyncapi/v2_6_0/schema/channels.py`, `.../v3_0_0/schema/channels.py`). We have never emitted it.

The AsyncAPI 3.0 Channel Object says of `parameters`:

> A map of the parameters included in the channel address. It MUST be present only when the address contains Channel Address Expressions.

and the Parameters Object:

> This map MUST contain all the parameters used in the parent channel address.

2.6 says the same thing about a channel name containing `{variables}`. So a subscriber on `logs.{level}` has been producing a non-conforming document for as long as address templates have existed.

It went unnoticed because the official rule does not fire: `asyncapi3-channel-parameters` is declared with `required: ['parameters']` in its Spectral input schema, so a channel *without* `parameters` is skipped rather than flagged. Adding the field is therefore not a free win — once present, it must match the address exactly, and both `Missed: …` and `redundant defined "…"` are `error`.

## What to build

Emit `parameters` for subscriber channels whose address contains `{param}`, with metadata from two sources:

**From the handler's type annotation** — the same path payload schemas already take. `Literal["info", "error"]` gives `enum`; `Path(default=...)` gives `default`.

**From new `Path()` keyword arguments** — `Path(description=..., examples=...)`. This is needed because of an asymmetry between the two spec versions:

| | 2.6 Parameter Object | 3.0 Parameter Object |
|---|---|---|
| fields | `description`, `schema`, `location` | `enum`, `default`, `description`, `examples`, `location` |

3.0 has **no** `schema` field — its Parameter Object is documentation, not validation. The two fields that carry meaning there (`description`, `examples`) are exactly the ones a type annotation cannot supply, so annotation-derived metadata alone would leave 3.0 nearly empty. In 2.6 the annotation's JSON Schema goes into `schema`; in 3.0 only `enum` and `default` are expressible and the rest is dropped rather than stuffed into a machine-written `description`.

`location` is not emitted in either version. Its grammar is `$message.(header|payload)[#fragment]` — a Path value comes from the message's address, which that grammar cannot express.

[...]

PR #3070, opened 2026-08-28
fix: a subscriber over several addresses is several channels (#2357)

Lancetnik:
**Top of a three-PR stack**: #3072 → #3038 → this. Review them in that order; this diff is only the four commits above #3038.

Chasing the address through each broker for #3038 turned up four defects in one family: an endpoint that reads more than one address, described as if it read one. They are separated out here because they are about *channels*, not about the `address` field.

## Why this is stacked on top and not underneath

These are bugs in code that is on `main` today, so it is fair to ask why they do not go first and merge on their own.

Because they cannot be *stated* first. Every test here asserts an address — `[c["address"] for c in channels.values()] == ["logs.info", "logs.error"]`, "a title leaves every address alone". Underneath #3038 the `address` field is the channel key with the handler name glued on, so those same tests would have to assert `address == "Titled:first"` — the exact bug #2357 is about — and then flip back one PR later. "One channel per topic" only becomes observable once the address field means something.

`SubscriberSpec` has no `address` field at all before #3038, so the NATS commit below would not even compile underneath it.

## A subscriber over several addresses is several channels

Kafka already emitted one channel per topic. A NATS JetStream consumer reaching a stream through `filter_subjects` did not: it rendered one channel whose binding subject was the subjects joined with commas. That string comes from `LogicSubscriber._resolved_subject_string`, where it labels a log line — `logs.info, logs.error` is not a subject anybody publishes to, and it is not an address. NATS now splits the way Kafka does.

## An explicit `title` was dropping topics

`get_schema` keyed one channel per topic, but with a `title` declared that key was the title itself — constant across the loop, so each topic overwrote the last and only one survived, silently.

A title cannot name several channels, so it now prefixes the address where there is more than one, and still names the channel on its own for a single-topic subscriber, leaving those documents unchanged.

The 3.0 generator keyed *operations* the same way and collapsed them identically, leaving channels that no operation referenced. Fixed in the same commit; its test asserts every channel is operated, which is what the first round of tests missed by only looking at `channels`.

## Channel order was hash-seed dependent

`KafkaSubscriberSpecification.topics` collected into a `set`, and set iteration order for strings is randomised per process, so the same broker produced a different document on every run — nothing anyone could commit or diff in CI. `dict.fromkeys` dedupes the same way and keeps declaration order. Confluent had it identically.

Verified across `PYTHONHASHSEED` 0–2, including a declaration mixing `Topic` objects with plain strings after #3026.

## One home for the naming rule

The last commit is a refactor, not a fix. `_channel_key` had become a copy in Kafka and another in Confluent; it moves to the shared `SubscriberSpecification`, and the per-broker `name` properties — six of them, three of which were dead — become one `channel_labels` list that the shared `name` delegates to.

`name` stays rather than being deleted: it is not in `__all__` and not in the Public API docs, but it is in the generated full API reference and reachable through the public `.specification` attribute, so removing it would break quietly.

`channel_labels` is a list, and not simply the address, because for two brokers they differ: MQTT keeps the `$share/` prefix its address drops, and RabbitMQ names a queue and an exchange rather than the routing key a message travels by.

### DIFF
--- a/faststream/_internal/endpoint/subscriber/specification.py
+++ b/faststream/_internal/endpoint/subscriber/specification.py
@@ -50,6 +50,17 @@ class SubscriberSpecification(Generic[T_BrokerConfig, T_SpecificationConfig]):
     def call_name(self) -> str:
         return self.calls.name or "Subscriber"
 
+    def _channel_key(self, address: str, *, split: bool) -> str:
+        """The key one address's channel goes under.
+
+        A `title` names one channel, so `split` prefixes it with the address to
+        keep several of them unique.
+        """
+        if not self.config.title_:
+            return f"{address}:{self.call_name}"
+
+        return f"{self.config.title_}:{address}" if split else self.config.title_
+
     def get_payloads(self) -> list[tuple["dict[str, Any]", str]]:
         payloads: list[tuple[dict[str, Any], str]] = []
 
--- a/faststream/_internal/endpoint/subscriber/specification.py
+++ b/faststream/_internal/endpoint/subscriber/specification.py
@@ -91,8 +91,20 @@ class SubscriberSpecification(Generic[T_BrokerConfig, T_SpecificationConfig]):
 
     @property
     @abstractmethod
-    def name(self) -> str:
+    def channel_labels(self) -> list[str]:
+        """The label each of this endpoint's channels is keyed by, one per channel.
+
+        Usually the address, but not always: MQTT keeps the `$share/` prefix its
+        address drops, and RabbitMQ names a queue and an exchange rather than the
+        routing key a message travels by.
+        """
         raise NotImplementedError
+
+    @property
+    def name(self) -> str:
+        """The key of the first channel this endpoint names."""
+        labels = self.channel_labels
+        return self._channel_key(labels[0], split=len(labels) > 1)
 
     @abstractmethod
     def get_schema(self) -> dict[str, "SubscriberSpec"]:
--- a/faststream/confluent/subscriber/specification.py
+++ b/faststream/confluent/subscriber/specification.py
@@ -12,15 +12,17 @@ class KafkaSubscriberSpecification(
 ):
     @property
     def topics(self) -> list[str]:
-        topics: set[str] = set()
+        """The topics this endpoint reads, in the order they were declared.
 
-        topics.update(f"{self._outer_config.prefix}{t.name}" for t in self.config.topics)
+        Deduped through a dict rather than a set: set order varies per process and
+        would reach the document as the order of its channels.
+        """
+        prefix = self._outer_config.prefix
 
-        topics.update(
-            f"{self._outer_config.prefix}{p.topic}" for p in self.config.partitions
-        )
+        topics = [f"{prefix}{t.name}" for t in self.config.topics]
+        topics.extend(f"{prefix}{p.topic}" for p in self.config.partitions)
 
-        return list(topics)
+        return list(dict.fromkeys(topics))
 
     @property
     def name(self) -> str:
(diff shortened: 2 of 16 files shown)


Lancetnik:
### The NATS split is already here

I came to propose "a consumer filtering on several subjects should be several channels" as a follow-up, having only looked at #3038. It is `035503a334` on this branch, and it works:

```
2 filters -> {'logs.a:<Lambda>': 'logs.a', 'logs.b:<Lambda>': 'logs.b'}
operations -> ['logs.a:<Lambda>Subscribe', 'logs.b:<Lambda>Subscribe']
```

Both channels are operated, which is the part the first round of tests in #3038 would have missed. `NatsSubscriberSpecification.address` and its `return None` branch are gone with it, so "several filter subjects, therefore no address" stops being a case that needs documenting rather than becoming one. And a subscriber with neither `subject` nor `filter_subjects` raises `SetupError` at declaration, so `subjects` is never empty and the loop never produces a channel-less endpoint.

Nothing to do here. Noting it so the follow-up does not get filed twice.

### What is still open: a literal brace cannot be expressed in an address

This is #3038's field rather than this PR's channels, but it is worth deciding at the top of the stack, because it is the last place the string is still ours.

A router prefix carrying an escaped brace renders like this today:

```python
router = RedisRouter(prefix="app{{v1}}.")

@router.subscriber("logs.{level}")
async def handle(msg: str) -> None: ...
```

```
address    = app{v1}.logs.{level}
parameters = <NOT EMITTED>
```

`{{v1}}` means "a literal `{v1}` on the wire" — that is the entire job of the escape, and #3072 established that it comes off once, where the declaration arrives. But the place it comes off into is a field where `{}` is *grammar*. AsyncAPI 3.0:

> Expressions MUST be composed by a name enclosed in curly braces (`{` and `}`). E.g., `{userId}`.

> **parameters** — A map of the parameters included in the channel address. It MUST be present only when the address contains Channel Address Expressions.

So a conformant parser reads `['v1', 'level']` — two expressions — where we mean one parameter and one literal. There is no escaping mechanism in the specification for a literal brace in an address; I checked the 3.0 reference and it does not define one anywhere.

Two consequences, one of which is already live:

- **Now.** `parameters` is commented out in `Channel` (`# parameters: Optional[Parameter] = None`), so we emit an address containing expressions and no `parameters` at all — for plain `logs.{level}`, before any escaping question arises.
- **After #3036.** `parameters` will declare `level` and not `v1`, because `v1` is not a parameter. A validator then sees an address expression with no matching entry, and the document is wrong in a way that points at the user's escape rather than at us.

The distinction the escape exists to carry has no representation on the far side. Undoing it into `address` is what loses it.

### A shape that fits the rule already established here

#3038 already answers "an address that cannot be stated" by leaving the field out — a fanout, a headers exchange — on the grounds that AsyncAPI reads an absent address as unknown and `""` does not say that. An address holding a literal brace is the same kind of thing: not unknown to *us*, but unstateable in the grammar the field is written in.

So: where a declaration's template contains a literal brace, omit `address` rather than emit one a parser will misread. The channel keeps its name and its broker binding, which is where the real string already lives and where no expression grammar applies — `bindings.redis.channel` and friends carry `app{v1}.logs.{level}` correctly today.

Costs a real address on an uncommon declaration. Buys never emitting one that means something other than what it says. Happy to be argued out of it — the alternative is documenting the limitation and letting `{{` render as a silent lie.

Worth noting the neighbouring case: Kafka and Confluent render this same prefix as `app{{v1}}.orders`, escape intact, because a topic is a literal and never meets the parser. That is not affected by the above, but it is a name Kafka itself refuses — `{` and `}` are outside `[a-zA-Z0-9._-]` — so the document describes an endpoint that cannot start. The commit that would catch it at declaration was split out of #3074 and has not landed.

borisalekseev (approved):

PR #3080, opened 2026-08-31
fix(redis): keep group read cursor in get_one and iterator

### LINKED ISSUE #3050: Bug: repeated Redis StreamSub.get_one() calls stop reading new group messages
**Describe the bug**

For a Redis stream subscriber using a consumer group, repeated calls to `StreamSub.get_one()` do not continue consuming new messages. The first call updates the subscriber's `last_id` to the returned message ID, and the next call sends that explicit ID to `XREADGROUP`. Redis treats that as a history read for the current consumer rather than a request for new group messages (which requires `>`), so the second call returns no new message.

This also affects the new `claim_min_idle_time` mode discussed in #3049: an explicit ID causes Redis to ignore the CLAIM option.

**How to reproduce**

1. Create a `StreamSub` with `group="workers"` and `consumer="worker-1"`.
2. Publish two messages to the stream.
3. Call `await subscriber.get_one()` twice.
4. Observe that the second call does not return the second new message.

The relevant current implementation updates the group-read cursor after a successful call:

- `faststream/redis/subscriber/usecases/stream_subscriber.py` uses `self.last_id` in `streams={...}` for `XREADGROUP`.
- The same method then assigns `self.last_id = message_id.decode()`.

**Expected behavior**

For a group subscriber whose `last_id` is `>`, each repeated `get_one()` call should continue consuming newly available messages. If explicit IDs are intentionally supported for pending/history reads, the implementation should keep separate state for the group selector and the returned message ID.

**Observed behavior**

After the first message, subsequent calls use the returned message ID as the next group selector. Redis then reads the current consumer's history instead of new messages; when no matching history exists, the call returns `None` (or the claiming parser can fail because CLAIM metadata is absent).

**Environment**

- FastStream: `main` at `3dafcc58d981235d9cc05425318e7cd47ca829df`
- Redis: consumer-group stream with `XREADGROUP`
- Python: not material to the cursor behavior

**Additional context**

PR #3049 identifies this behavior as a known limitation and currently tests `get_one()` only once. A regression test that calls `get_one()` repeatedly with two messages would make the behavior explicit while the cursor handling is fixed.

Pizzagami:
# Description

 get_one() and __aiter__() were both passing self.last_id to XREADGROUP instead of self.read_id. Since last_id gets updated to the last delivered message ID after each call, the cursor was stuck and Redis would never deliver the next message in consumer group mode.

Fixes #3050 

## Type of change

## Checklist

### DIFF
--- a/tests/brokers/redis/test_consume.py
+++ b/tests/brokers/redis/test_consume.py
@@ -989,7 +989,7 @@ class TestConsumeStream(RedisTestcaseConfig):
         self,
         queue: str,
     ) -> None:
-        broker = self.get_broker(apply_types=True)
+        broker = self.get_broker()
         subscriber = broker.subscriber(
             stream=StreamSub(queue, group="test_group", consumer="test_consumer")
         )
@@ -1012,7 +1012,7 @@ class TestConsumeStream(RedisTestcaseConfig):
         self,
         queue: str,
     ) -> None:
-        broker = self.get_broker(apply_types=True)
+        broker = self.get_broker()
         subscriber = broker.subscriber(
             stream=StreamSub(queue, group="test_group", consumer="test_consumer")
         )
--- a/tests/brokers/redis/test_consume.py
+++ b/tests/brokers/redis/test_consume.py
@@ -985,6 +985,52 @@ class TestConsumeStream(RedisTestcaseConfig):
             calls = [call(msg) for msg in expected_messages]
             mock.assert_has_calls(calls=calls)
 
+    async def test_get_one_group_cursor_advances(
+        self,
+        queue: str,
+    ) -> None:
+        broker = self.get_broker(apply_types=True)
+        subscriber = broker.subscriber(
+            stream=StreamSub(queue, group="test_group", consumer="test_consumer")
+        )
+
+        async with self.patch_broker(broker) as br:
+            await br.start()
+
+            await br.publish("first", stream=queue)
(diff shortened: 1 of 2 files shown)


tests/brokers/redis/test_consume.py
```diff
@@ -985,6 +985,52 @@ async def consume() -> None:
...
+            assert msg1 is not None
+            assert msg2 is not None
+            assert await msg1.decode() == "first"
+            assert await msg2.decode() == "second"
+
+    async def test_iterator_group_cursor_advances(
+        self,
+        queue: str,
+    ) -> None:
+        broker = self.get_broker(apply_types=True)
```
IvanKirpichnikov: can remove `apply_types=True`

tests/brokers/redis/test_consume.py
```diff
@@ -985,6 +985,52 @@ async def consume() -> None:
             calls = [call(msg) for msg in expected_messages]
             mock.assert_has_calls(calls=calls)
 
+    async def test_get_one_group_cursor_advances(
+        self,
+        queue: str,
+    ) -> None:
+        broker = self.get_broker(apply_types=True)
```
IvanKirpichnikov: can remove `apply_types=True`

IvanKirpichnikov (changes requested):

IvanKirpichnikov (approved):

PR #3088, opened 2026-09-01
feat(redis): use native async cluster pubsub

### LINKED ISSUE #3009: Feature: migrate Redis Cluster pub/sub to native redis-py 8 async API
**Status: ready to implement, held for 1.0.0** — it drops support for Redis Cluster on `redis-py < 8.0.0`, so it needs a major release. Raised by @fbartonoctopus in https://github.com/ag2ai/faststream/pull/2893#issuecomment-5215949247.

Redis Cluster pub/sub runs through a synchronous workaround merged in #2854, because `redis.asyncio.cluster` had no pub/sub at the time. It is still on `main`:

- [`configs/state.py`](https://github.com/ag2ai/faststream/blob/main/faststream/redis/configs/state.py) — `_SyncPubSubProxy`, `_get_sync_cluster()`, `sync_publish()`, and a `ThreadPoolExecutor(max_workers=1)` per connection, plus `_sync_cluster` / `_thread_pool` fields carried on the base `ConnectionState`.
- [`publisher/producer.py`](https://github.com/ag2ai/faststream/blob/main/faststream/redis/publisher/producer.py) — `RedisClusterFastProducer`, whose publish path hops to a thread on every message.
- [`broker/cluster_broker.py`](https://github.com/ag2ai/faststream/blob/main/faststream/redis/broker/cluster_broker.py) — `_make_channel_subscriber()` and a `subscriber()` override that exist only to route around the missing async pub/sub.

The cost is a second Redis client per app, a thread pool in an otherwise async path, and a cluster code path that diverges from the standalone one everywhere it touches pub/sub.

redis-py 8.0.0 added async cluster pub/sub, so the workaround can be deleted and cluster can use the same `RedisFastProducer` and stock `ChannelSubscriber.start()` as standalone.

Verified on `redis-py` 8.0.1:

```python
AsyncCluster.pubsub: True
AsyncCluster.publish: True
AsyncCluster.spublish: True
ClusterPubSub: redis.asyncio.cluster.ClusterPubSub
```

First shipped in 8.0.0 — `class ClusterPubSub` in `redis/asyncio/cluster.py` across tags:

```python
v6.2.0: 0
v7.0.0: 0
v7.4.0: 0
v8.0.0: 2
```

#2893 already implemented this end to end — removal of the sync machinery, producer unification, cluster broker simplification, tests and docs — and was closed rather than rejected. It is the natural starting point.

**Feature code example**

No public API change. Cluster pub/sub keeps its current surface and stops going through a thread:

```python

broker = RedisClusterBroker("redis://localhost:7000")
app = FastStream(broker)


@broker.subscriber("test")


@app.after_startup
async def _() -> None:
    await broker.publish("hi", channel="test")
```

The open decision is how to express the new requirement, and it changes who is affected:

1. **Runtime guard, floor stays at 5.0.0** — what #2893 did: raise a clear error from the cluster connection on `redis-py < 8`. Breaking only for cluster users; standalone, sentinel and non-Redis users are untouched.
2. **Raise the floor to `redis>=8.0.0`** — simpler code and one supported path, but breaking for every Redis user including those who never touch cluster.

Current constraint is `redis>=5.0.0,<9.0.0`; #2952 lifted only the ceiling, so redis-py 8 is permitted but not required today.

**Additional context**

Worth confirming during implementation, since #2893 predates the current `main`:

- `legacy_responses=True` on the cluster connection — redis-py 8 defaults to RESP3, and stream/list parsing assumes RESP2-shaped responses.
- Interaction with #2956, which already replaced `lib_name` / `lib_version` with `driver_info`.
- Sharded pub/sub (`spublish` / `ssubscribe`) is available on the async cluster client and is not exposed today — out of scope here, but it becomes reachable once this lands.

lxingy3:
## Summary

- replace the synchronous Redis Cluster Pub/Sub client and thread-pool proxy with redis-py 8's native async cluster API
- reuse the standard Redis producer and channel subscriber for standalone and clustered Redis
- initialize the cluster client before exposing it so channel-only applications start with a populated slot cache, and close partially initialized clients on failure
- keep the package-wide `redis>=5,<9` range while rejecting cluster connections before client construction when native async Pub/Sub is unavailable
- preserve RESP2-shaped responses, forward and type `credential_provider`, and fix DriverInfo selection across redis-py major versions

## Compatibility

`RedisClusterBroker` now requires redis-py 8.0.0 or newer at connection time. Standalone and Sentinel users retain the existing redis-py 5 floor. Regular Pub/Sub is supported; sharded Pub/Sub remains out of scope.

## Validation

- Redis non-connected suite: 386 passed, 6 skipped, 1 xfailed
- Ruff format/check, codespell, Mypy (516 files), and Pyright
- Bandit and Semgrep (0 findings)
- full MkDocs build
- isolated compatibility probes on redis-py 5.3.1, 7.4.0, 8.0.0, and 8.0.1

The connected Redis Cluster suite was not run locally because no cluster runtime was available; the existing CI suite covers regular Pub/Sub, pattern Pub/Sub, request/reply, cleanup, and restart behavior.

Builds on the direction and prior investigation in #2893. Thanks to @fbartonoctopus for identifying the credential-provider gap.

Fixes #3009

### DIFF
--- a/faststream/redis/broker/cluster_broker.py
+++ b/faststream/redis/broker/cluster_broker.py
@@ -70,6 +70,9 @@ class RedisClusterBroker(RedisBroker):
         if maxlen is not None:
             publish_kwargs["maxlen"] = maxlen
 
+        if not self._cluster_state:
+            await self._connect()
+
         return cast(
             "int | bytes",
             await super().publish(
--- a/faststream/redis/broker/cluster_broker.py
+++ b/faststream/redis/broker/cluster_broker.py
@@ -1,6 +1,6 @@
 import warnings
 from collections.abc import Iterable
-from typing import TYPE_CHECKING, Any, Optional, Union, cast
+from typing import TYPE_CHECKING, Any, Optional, cast
 
 from redis.asyncio.cluster import ClusterNode
 from redis.asyncio.connection import SSLConnection
@@ -8,17 +8,10 @@ from typing_extensions import Unpack
 
 from faststream._internal.constants import EMPTY
 from faststream.redis.broker import RedisBroker
-from faststream.redis.configs.state import (
-    ConnectionState,
-    RedisClusterConnectionState,
-)
-from faststream.redis.publisher.producer import (
-    RedisClusterFastProducer,
-)
+from faststream.redis.configs.state import RedisClusterConnectionState
 from faststream.redis.schemas.types import (
     CLUSTER_INCOMPATIBLE_PARAMS,
 )
-from faststream.redis.subscriber.usecases.basic import LogicSubscriber
 
 if TYPE_CHECKING:
     from types import TracebackType
@@ -26,9 +19,7 @@ if TYPE_CHECKING:
     from redis.asyncio.client import Pipeline
 
     from faststream._internal.basic_types import SendableMessage
-    from faststream.redis.schemas import ListSub, PubSub, StreamSub
     from faststream.redis.schemas.types import RedisClusterParams
-    from faststream.redis.subscriber.usecases import ChannelSubscriber
     from faststream.security import BaseSecurity
 
 
@@ -46,70 +37,12 @@ class RedisClusterBroker(RedisBroker):
         self,
         connection_options: dict[str, Any],
         kwargs: dict[str, Any],
-    ) -> "ConnectionState[Any]":
+    ) -> RedisClusterConnectionState:
         return RedisClusterConnectionState(connection_options)
-
-    def _make_producer(
-        self,
-        connection_state: "ConnectionState[Any]",
-        kwargs: dict[str, Any],
-    ) -> "RedisClusterFastProducer":
-        state = cast("RedisClusterConnectionState", connection_state)
-        return RedisClusterFastProducer(
-            connection=state,
-            cluster_state=state,
-            parser=kwargs.get("parser"),
-            decoder=kwargs.get("decoder"),
-            message_format=self.message_format,
-            serializer=kwargs.get("serializer"),
-        )
 
     @property
     def _cluster_state(self) -> RedisClusterConnectionState:
         return cast("RedisClusterConnectionState", self.config.broker_config.connection)
-
-    def subscriber(  # type: ignore[override]
-        self,
-        channel: Union["PubSub", str, None] = None,
-        *,
-        list: Union["ListSub", str, None] = None,
-        stream: Union["StreamSub", str, None] = None,
-        **kwargs: Any,
-    ) -> "LogicSubscriber":
-        if channel is not None:
-            return self._make_channel_subscriber(channel, **kwargs)
-        return super().subscriber(
-            channel=None,
-            list=list,
-            stream=stream,
-            **kwargs,
-        )
-
-    def _make_channel_subscriber(
-        self,
-        channel: Union["PubSub", str],
-        **kwargs: Any,
-    ) -> "ChannelSubscriber":
-        state = self._cluster_state
-
-        sub = cast(
-            "ChannelSubscriber",
-            super().subscriber(channel=channel, list=None, stream=None, **kwargs),
-        )
-
-        async def _patched_start() -> None:
-            if sub.subscription:
-                return
-            psub = state.pubsub()
-            sub.subscription = psub  # type: ignore[assignment]
-            if sub.channel.pattern:
-                await psub.psubscribe(sub.channel.name)
-            else:
-                await psub.subscribe(sub.channel.name)
-            await LogicSubscriber.start(sub, psub)
-
-        sub.start = _patched_start  # type: ignore[method-assign]
-        return sub
 
     async def publish(  # type: ignore[override]
         self,
--- a/faststream/redis/broker/cluster_broker.py
+++ b/faststream/redis/broker/cluster_broker.py
@@ -19,6 +19,7 @@ if TYPE_CHECKING:
     from redis.asyncio.client import Pipeline
 
     from faststream._internal.basic_types import SendableMessage
+    from faststream.redis.configs.state import ConnectionState
     from faststream.redis.schemas.types import RedisClusterParams
     from faststream.security import BaseSecurity
 
@@ -37,7 +38,7 @@ class RedisClusterBroker(RedisBroker):
         self,
         connection_options: dict[str, Any],
         kwargs: dict[str, Any],
-    ) -> RedisClusterConnectionState:
+    ) -> "ConnectionState[Any]":
         return RedisClusterConnectionState(connection_options)
 
     @property
@@ -70,6 +71,8 @@ class RedisClusterBroker(RedisBroker):
         if maxlen is not None:
             publish_kwargs["maxlen"] = maxlen
 
+        # Direct publish may run before start(), while cluster state exposes a client
+        # only after topology initialization.
         if not self._cluster_state:
             await self._connect()
 
--- a/faststream/redis/configs/state.py
+++ b/faststream/redis/configs/state.py
@@ -1,6 +1,5 @@
 from abc import ABC, abstractmethod
 from collections.abc import Mapping, Sequence
-from contextlib import suppress
 from typing import Any, Generic, TypeVar
 
 from redis.asyncio.client import Redis
@@ -10,7 +9,7 @@ from redis.asyncio.sentinel import Sentinel
 
 from faststream.__about__ import __version__
 from faststream.exceptions import IncorrectState
-from faststream.redis._compat import REDIS_V720
+from faststream.redis._compat import REDIS_V720, REDIS_V800
 
 if REDIS_V720:
     from redis.driver_info import DriverInfo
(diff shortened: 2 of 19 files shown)


lxingy3:
The Redis Cluster job exposed a direct-publish compatibility regression after the producer was unified: an unconnected cluster broker reached the shared producer before the cluster client existed.

Fixed in e552bc5b by connecting the cluster broker before direct publication, matching the existing `publish_batch` lifecycle. The real-cluster regression test remains unchanged, and a focused unit test now covers the disconnected direct-publish path.

Local validation:
- Redis cluster unit tests: 38 passed
- Redis non-connected suite: 386 passed, 6 skipped, 1 xfailed
- Ruff, Mypy (516 files), and Pyright: passed

lxingy3:
Could a maintainer rerun the failed test-confluent-real job? It hit the same unrelated three docs-test failures as #3089 after repeated LeaveGroupRequest timeouts:

- tests/docs/getting_started/publishing/test_broker_context.py::test_broker_context_confluent
- tests/docs/getting_started/subscription/test_real.py::test_handle_confluent
- tests/docs/getting_started/subscription/test_real.py::test_validate_confluent

This PR does not touch Confluent code. The real Redis Cluster job, all Python and OS jobs, lint, CodeQL, and every other broker job passed. I also tried the GitHub rerun command, but contributors do not have the required repository permission.

faststream/redis/broker/cluster_broker.py
```diff
@@ -46,71 +37,13 @@ def _make_connection_state(
         self,
         connection_options: dict[str, Any],
         kwargs: dict[str, Any],
-    ) -> "ConnectionState[Any]":
+    ) -> RedisClusterConnectionState:
```
IvanKirpichnikov: Why was it changed return type hint?

faststream/redis/configs/state.py
```diff
@@ -130,129 +131,26 @@ async def connect(self) -> "Redis[bytes]":
...
+    """Manages a Redis Cluster connection using its native async client."""
 
     async def connect(self) -> "RedisCluster[bytes]":
         if self._connected:
             return self._client  # type: ignore[return-value]
 
+        _ensure_cluster_pubsub_supported()
+
         connection_kwargs = {k: v for k, v in self._options.items() if v is not None}
+        connection_kwargs.setdefault("legacy_responses", True)
```
IvanKirpichnikov, line 143: why?

faststream/redis/configs/state.py
```diff
@@ -130,129 +131,26 @@ async def connect(self) -> "Redis[bytes]":
...
-
-    def __bool__(self) -> bool:
-        return self._connected
+    """Manages a Redis Cluster connection using its native async client."""
 
     async def connect(self) -> "RedisCluster[bytes]":
         if self._connected:
             return self._client  # type: ignore[return-value]
 
+        _ensure_cluster_pubsub_supported()
```
IvanKirpichnikov, line 139: maybe use redis_versoin from `faststream.redis._compat` instead of checking via attributes

faststream/redis/configs/state.py
```diff
@@ -130,129 +131,26 @@ async def connect(self) -> "Redis[bytes]":
...
         connection_kwargs = {k: v for k, v in self._options.items() if v is not None}
+        connection_kwargs.setdefault("legacy_responses", True)
         connection_kwargs |= _get_driver_info()
 
         client: RedisCluster[bytes] = RedisCluster(**connection_kwargs)
+        try:
+            await client.initialize()
+        except BaseException:
+            with suppress(Exception):
+                await client.aclose()  # type: ignore[attr-defined]
```
IvanKirpichnikov: Why call aclose if the initialization failed?

faststream/redis/publisher/producer.py
```diff
@@ -129,10 +121,10 @@ async def __publish(
         connection = cmd.pipeline or self._connection.client
 
         if cmd.destination_type is DestinationType.Channel:
-            return await connection.publish(cmd.destination, msg)
+            return cast("int", await connection.publish(cmd.destination, msg))
 
         if cmd.destination_type is DestinationType.List:
-            return await connection.rpush(cmd.destination, msg)
+            return cast("int", await connection.rpush(cmd.destination, msg))
```
IvanKirpichnikov: why is it used cast?

tests/brokers/redis/cluster/test_cluster.py
```diff
@@ -245,9 +247,12 @@ def test_broker_uses_config_composition(self) -> None:
         broker = RedisClusterBroker()
         assert isinstance(broker.config, ConfigComposition)
 
-    def test_start_stop_lifecycle(self) -> None:
+    def test_start_stop_lifecycle(self, monkeypatch: pytest.MonkeyPatch) -> None:
         import anyio
 
+        initialize = AsyncMock()
+        monkeypatch.setattr(state_module.RedisCluster, "initialize", initialize)
```
IvanKirpichnikov, line 255: maybe use `tests.tools.spy_decorator`?

tests/brokers/redis/cluster/test_cluster_test_client.py
```diff
@@ -18,8 +21,19 @@ async def test_broker_with_real_patches_publishers_and_subscribers(
     ) -> None:
         await super().test_broker_with_real_patches_publishers_and_subscribers(queue)
 
-    async def test_broker_gets_patched_attrs_within_cm(self) -> None:
+    async def test_broker_gets_patched_attrs_within_cm(
+        self,
+        monkeypatch: pytest.MonkeyPatch,
+    ) -> None:
+        initialize = AsyncMock()
+        monkeypatch.setattr(state_module.RedisCluster, "initialize", initialize)
```
IvanKirpichnikov, line 29: maybe use `tests.tools.spy_decorator`?

faststream/redis/broker/cluster_broker.py
```diff
@@ -137,6 +70,9 @@ async def publish(  # type: ignore[override]
         if maxlen is not None:
             publish_kwargs["maxlen"] = maxlen
 
+        if not self._cluster_state:
+            await self._connect()
```
IvanKirpichnikov, line 77: why?

faststream/redis/publisher/producer.py
```diff
@@ -81,20 +77,16 @@ def connect(
...
 class RedisFastProducer(BaseRedisFastProducer):
-    """Producer for a single-node Redis."""
-
-    _connection: "ConnectionState[Redis[bytes]]"
+    """Producer for standalone and clustered Redis clients."""
 
     def __init__(
         self,
-        connection: "ConnectionState[Redis[bytes]]",
+        connection: "ConnectionState[Any]",
```
IvanKirpichnikov: maybe `ConnectionState[Redis[bytes] | RedisCluster[bytes]]`?

faststream/redis/publisher/producer.py
IvanKirpichnikov: Then the cast call is not needed

tests/brokers/redis/cluster/test_cluster_test_client.py
```diff
@@ -18,8 +21,19 @@ async def test_broker_with_real_patches_publishers_and_subscribers(
...
+
         await super().test_broker_gets_patched_attrs_within_cm(FakeProducer)
+        initialize.assert_awaited_once()
+
+    @pytest.mark.connected()
+    async def test_broker_with_real_doesnt_get_patched(self) -> None:
+        await super().test_broker_with_real_doesnt_get_patched()
```
IvanKirpichnikov, line 36: why use `@pytest.mark.connected()`?

lxingy3:
Addressed all review points in `a8520776`:

- Added an explicit redis-py 8 compatibility flag and kept the connection-state return type aligned with the base contract.
- Replaced mocked initialization with real-method spies in connected tests, and removed the redundant close after `RedisCluster.initialize()` failures.
- Narrowed the producer connection union and consolidated the dynamic RedisCluster command surface into one documented cast.
- Kept `legacy_responses=True` because FastStream's parser relies on RESP2-shaped replies, and kept direct-publish initialization because `publish()` is valid before `start()`.

The updated CI is fully green. @IvanKirpichnikov, could you take another look?

faststream/redis/configs/state.py
```diff
@@ -130,129 +130,22 @@ async def connect(self) -> "Redis[bytes]":
...
-    def __bool__(self) -> bool:
-        return self._connected
+    """Manages a Redis Cluster connection using its native async client."""
 
     async def connect(self) -> "RedisCluster[bytes]":
         if self._connected:
-            return self._client  # type: ignore[return-value]
+            return self.client
+
+        _ensure_cluster_pubsub_supported()
```
IvanKirpichnikov, line 139: Maybe we should remove the _ensure_cluster_pubsub_supported function.

faststream/redis/publisher/producer.py
```diff
@@ -151,7 +148,7 @@ async def __publish(
     async def request(self, cmd: "RedisPublishCommand") -> "Any":
         nuid = NUID()
         reply_to = str(nuid.next(), "utf-8")
-        psub = self._connection.client.pubsub()
+        psub = cast("Redis[bytes]", self._connection.client).pubsub()
```
IvanKirpichnikov, line 151: why cast? `RedisCluster` do can in `pubsub`

IvanKirpichnikov (changes requested):
And please answer my questions in the comments.

IvanKirpichnikov:
Hi. I’m closing your pull request because we didn’t receive any feedback and reopened our own one: https://github.com/ag2ai/faststream/pull/3131

PR #2947, opened 2026-07-20
feat: Handle errors on a consume task in the subscriber

pbonneaudiabolocom:
…sk in the subscriber.

# Description

Whene a consume task fails, we can now adapt the behaviour of the system by overloading _handle_task_exception

Fixes #2945

## Type of change

Please delete options that are not relevant.

- [ ] Documentation (typos, code examples, or any documentation updates)
- [ ] Bug fix (a non-breaking change that resolves an issue)
- [X] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a fix or feature that would disrupt existing functionality)
- [ ] This change requires a documentation update

## Checklist

- [x] My code adheres to the style guidelines of this project (`just lint` shows no errors)
- [x] I have conducted a self-review of my own code
- [x] I have made the necessary changes to the documentation
- [x] My changes do not generate any new warnings
- [x] I have added tests to validate the effectiveness of my fix or the functionality of my new feature
- [x] Both new and existing unit tests pass successfully on my local environment by running `just test-coverage`
- [x] I have ensured that static analysis tests are passing by running `just static-analysis`
- [x] I have included code examples to illustrate the modifications

PR #3095, opened 2026-09-04
feat: Allow capturing exception on task execution

### LINKED ISSUE #2945: Bug: group removal handling in Redis Stream not functional
**Describe the bug**
When the redis consume group is removed during the life time of the app, the application loop on task execution.

**How to reproduce**
Include source code:
Run faststream with a redis stream and use consume group.

XGROUP DESTROY that group

You'll get that result

2026-07-20 11:29:08,328 INFO     -                     |            - callback for Task-4419 is being executed...
2026-07-20 11:29:08,330 INFO     -                     |            - callback for Task-4420 is being executed...
2026-07-20 11:29:08,331 INFO     -                     |            - callback for Task-4421 is being executed...

**Environment**
faststream 0.7.2

**Additional context**
MR: https://github.com/ag2ai/faststream/pull/2855

pbonneaudiabolocom:
# Description

It allows to customize the task exception managment + the consume error behaviour

Fixes https://github.com/ag2ai/faststream/issues/2945

## Type of change

## Checklist

### DIFF
--- a/docs/docs/en/getting-started/subscription/index.md
+++ b/docs/docs/en/getting-started/subscription/index.md
@@ -157,6 +157,24 @@ Also, synchronous functions are supported as well:
 
     Such functions run in a ThreadPool using `#!python anyio.to_thread.run_sync()`, so they don't block the event loop.
 
+## Consume-loop errors
+
+Subscribers handle errors raised while fetching messages through the asynchronous
+`#!python handle_consume_error(error)` method. Custom subscriber implementations
+can override this method and use
+`#!python await super().handle_consume_error(error)` to preserve the behavior of
+their parent class.
+
+The default implementation raises the error. Broker-specific subscriber classes
+may provide a different default, such as logging the error and retrying the
+consume loop.
+
+If an error escapes the consume loop, the subscriber calls its public
+`#!python handle_task_exception(...)` method. Its default behavior restarts the
+failed task. Custom subscriber implementations can override this method to
+inspect the exception, decide not to restart, or delegate to
+`#!python super().handle_task_exception(...)` for the default restart behavior.
+
 ## Message Body Serialization
 
 Generally, **FastStream** uses your function type annotation to serialize incoming message body with [**Pydantic**](https://docs.pydantic.dev){.external-link target="_blank"}. This is similar to how [**FastAPI**](https://fastapi.tiangolo.com){.external-link target="_blank"} works (if you are familiar with it).
--- a/docs/docs/en/redis/streams/groups.md
+++ b/docs/docs/en/redis/streams/groups.md
@@ -58,6 +58,15 @@ that handler. It reaches the public `#!python handle_task_exception(...)` hook
 as a `#!python StreamGroupNotFoundError`, so custom subscriber implementations
 can capture it and decide whether the failed task should be restarted.
 
+The following example recreates the consumer group and then restarts the failed
+consume task. It creates the group at `$`, matching the default stream
+subscriber behavior; use `0` instead if the replacement group must read all
+existing stream entries.
+
+```python linenums="1"
+{! docs_src/redis/stream/recreate_group.py !}
+```
+
 ## Publishing a message
 
 Publishing a message is the same as what's defined on [Stream Publishing](./publishing.md).
--- a/docs/docs/en/redis/streams/groups.md
+++ b/docs/docs/en/redis/streams/groups.md
@@ -46,6 +46,18 @@ Define a subscription to a Redis stream with a specific Consumer Group using the
 {! docs_src/redis/stream/group.py [ln:8-10] !}
 ```
 
+## Missing consumer groups
+
+If the stream or its consumer group is deleted while the subscriber is running,
+Redis returns a `#!python NOGROUP` error. FastStream converts this to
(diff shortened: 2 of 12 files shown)


pbonneaudiabolocom:
Pursuing efforts of PR https://github.com/ag2ai/faststream/pull/2947

@IvanKirpichnikov 
@powersemmi 
@Lancetnik

IvanKirpichnikov (changes requested):
Hi. Let’s do it this way: let the subscriber accept `exception_handler: Callable[[BaseException], bool]`.

If exception_handler returns True, the exception is suppressed. If it returns False, the exception is considered unhandled, following the same logic as context managers.

This means that this hook should be available on every subscriber. Not only for redis

IvanKirpichnikov:
I don’t like this approach for several reasons:

Before defining your own handler, you have to save the “old” one:
+ `default_task_exception_handler = subscriber.handle_task_exception`
+ You have to override a method on the subscriber. Not a fan of that.
+ It accepts all sorts of `func`, `func_args`, and `func_kwargs` and uses `add_task`.

pbonneaudiabolocom:
I agree, the issue is that you can't overload subscriber because it's provided by the broker. 
And you do not instantiate it either.

So, passing another argument, exception_handler: Callable[[BaseException], bool], is going to be as difficult as switching a method to replace it by another.

After, my knowledge on how faststream works is limited. I'm a user, not so knowledgable about what's under the hood.

The only object we do have access to is the broker, do we really want the broker to be the entry point for a subscriber exception manager ?

I'm honnestly trying to find a way, but I find many average solution, but no good one when I'm satisfied by the result.

IvanKirpichnikov:
> The only object we do have access to is the broker, do we really want the broker to be the entry point for a subscriber exception manager ?

I mean passing the exception_handler to the subscriber. 

Although, if we take the idea further, I think we could also support setting the exception_handler at the broker level, but the question arises as to what order to call them in.
I think we should first call the exception_handler on the subscriber, and if it returns False, then call the exception_handler on the broker.

pbonneaudiabolocom:
I still don't get how we are supposed to access the subscriber to do that. 
This object is created by the broker, so  you really want to do something such as

broker.subscriber.add_exception_handler(<Callable>)

I mean, it could be done for sure, but then you'll come back quickly to something such as getting the default exception_handler for the subscriber, and replacing it by a function taht do some stuff and call the default one in case needed.

I cannot figure what you want to do...

IvanKirpichnikov:
> I still don't get how we are supposed to access the subscriber to do that. This object is created by the broker, so you really want to do something such as
> 
> broker.subscriber.add_exception_handler()
> 
> I mean, it could be done for sure, but then you'll come back quickly to something such as getting the default exception_handler for the subscriber, and replacing it by a function taht do some stuff and call the default one in case needed.
> 
> I cannot figure what you want to do...

I mean something like that.
```python
def global_exception_handler(exc: BaseException) -> bool:
    if isinstance(exc, GlobalError):
        ...
        return True
    return False

broker = NatsBroker(..., exception_handler=global_exception_handler)


def my_exception_handler(exc: BaseException) -> bool:
    if isinstance(exc, MyError):
        # do custom logic
        ...

        # i'm process error
        return True

    # i'm not process error
    return False

@broker.subscriber("subject", exception_handler=my_exception_handler)
async def handle(body: Body):
    if body.act == 1:
        raise MyError
    else:
        raise GlobalError

```

pbonneaudiabolocom:
ok, so on top of the subscriber exception_handler, you want also a broker exception handler.
Both being injected using constructor.

That could be elegant.

Are we allowing to overide the default exception handling, or to call it ?
With such a system, how could we implement the logic : 
```
If groupError:
Do something custom
Else:
Use regular excetion handling //could be also disable exeption by default.
```

Because I think that's going to be a common usecase. Target an error, do a specific action, but let faststream defaut behaviour in any other usecases.
Implementing the custom_exception_handler outside the subscriber(and before it's creation), we will not have access to the subscriber, nor it's values.

IvanKirpichnikov:
> ok, so on top of the subscriber exception_handler, you want also a broker exception handler. Both being injected using constructor.
> 
> That could be elegant.
> 
> Are we allowing to overide the default exception handling, or to call it ? With such a system, how could we implement the logic :
> 
> ```
> If groupError:
> Do something custom
> Else:
> Use regular excetion handling //could be also disable exeption by default.
> ```
> 
> Because I think that's going to be a common usecase. Target an error, do a specific action, but let faststream defaut behaviour in any other usecases. Implementing the custom_exception_handler outside the subscriber(and before it's creation), we will not have access to the subscriber, nor it's values.

My idea is as follows.

First, the exception_handler is called for the subscriber. If it returns False, then the exception_handler for the broker is executed. If it also returns False, the default error handling is triggered.

Is everything clear to you? If not, I’m ready to answer your questions.

pbonneaudiabolocom:
Hi,

Your idea seems nice, but I'm not sure I'm the right one to implement it. I barely understand faststream logic, so except burning token, my value will be very low here...

I don't know how I could assert anything... especially with such a large scope impact.

IvanKirpichnikov:
It doesn’t seem too difficult. It’s worth a try.

Patience and hard work will conquer all :)

PR #3096, opened 2026-09-04
feat(redis): support cluster publishing pipelines

### LINKED ISSUE #3044: Feature:Support the `pipeline` parameter in RedisClusterBroker publish methods

  With `RedisBroker`, the `pipeline` parameter of `publish()` allows queueing the publish (XADD / RPUSH / PUBLISH) into a redis-py pipeline together with other commands, so that publishing and related state updates
  (e.g. counters) execute atomically in one MULTI.

  With `RedisClusterBroker`, the parameter is ignored with a `RuntimeWarning` ("Pipeline is not supported in Redis Cluster and will be ignored."), so there is no supported way to publish atomically with other
  commands.

  Redis Cluster does support MULTI as long as all keys map to the same slot (hash tags), and redis-py's async `ClusterPipeline(transaction=True)` implements exactly that. Publishing to a stream and updating sibling
  keys that share the stream's hash tag is a valid and common pattern, so the limitation seems implementation-side rather than fundamental.

  Accept a `redis.asyncio.cluster.ClusterPipeline` as the `pipeline` argument on `RedisClusterBroker.publish()` / `publish_batch()`, queueing the command into the pipeline instead of executing it.

  Implementation note: the current producer path does `connection = cmd.pipeline or client` followed by `await connection.xadd(...)`. That works for the non-cluster `Pipeline` (awaiting it is a no-op returning
  itself), but awaiting a queued `ClusterPipeline` triggers `initialize()`, which resets its command stack — presumably the reason the parameter was excluded. A ClusterPipeline-aware branch that queues without
  awaiting should be enough.

  **Feature code example**

  ```python

  broker = RedisClusterBroker()
  redis = await broker.connect()

  # All keys share the "{jobs}" hash tag -> same slot -> MULTI is valid
  async with redis.pipeline(transaction=True) as pipe:
      pipe.incr("{jobs}:total_count")
      await broker.publish(message, stream="{jobs}:QUEUE", pipeline=pipe)
      await pipe.execute()  # INCR + XADD atomically
  ```

  - `publish_batch()` (suggested by the docs) covers batching multiple messages into one list, but not combining a publish with other commands, and provides no transactional semantics.
  - Current workaround: encode the message manually with the broker's message format and queue the XADD directly (`pipe.xadd(stream, {"__data__": payload})`). This works but duplicates the envelope construction that
  `publish()` already owns.

  **Additional context**

  Docs reference: "Pipeline is not supported in Redis Cluster. If you are using RedisClusterBroker, the pipeline parameter is not available. Consider using publish_batch() with individual requests instead."

  If cross-slot usage is a concern, raising `CrossSlotTransactionError` (as redis-py already does) or documenting the same-slot requirement seems preferable to ignoring the parameter entirely.

lxingy3:
# Description

Fixes #3044.

`RedisClusterBroker` currently ignores `pipeline=`, so a state update and a publication cannot share a transaction. Simply forwarding it is not enough: awaiting the `ClusterPipeline` returned by a queued command reinitializes the pipeline and clears its queued commands.

This change forwards the supplied pipeline through broker and publisher calls for lists, streams, and list batches. Queued calls return the pipeline without awaiting it; commands issued after `WATCH` but before `MULTI` still execute immediately. Publishing without a pipeline keeps its existing behavior, including automatic connection for cluster batch publishing.

Channel `PUBLISH` remains blocked by redis-py in cluster pipelines. There is no immediate-publish fallback. Atomic transactions require redis-py 6.2.0 or later and keys in the same hash slot; ordinary pipelines are not atomic.

## Type of change

- [x] New feature
- [x] Documentation update

## Validation

- 110 related tests passed, covering standalone Redis publishing, cluster publishing, producers, pipelines, and the documentation example. Redis 7.0.15 ran as a standalone server and a three-node cluster. Five existing slow tests were excluded.
- The standalone test broker used `127.0.0.1` locally. Windows/WSL's `localhost` IPv6 fallback exceeded existing timeouts; the same RPC timeout reproduced on unchanged main. No production code or test timeout was changed for this environment issue.
- redis-py 6.2.0: all nine focused cluster tests passed, including same-slot transactions, WATCH/MULTI, cross-slot rejection, and the runnable example.
- redis-py 5.0.0: all four non-connected queue-preservation and command-restriction cases passed. Transactions are not claimed for this version.
- All four deterministic regression cases fail on unchanged main.
- Ruff, codespell, Mypy, Pyright, Bandit, and Semgrep passed. The remaining pre-commit hooks passed; local `just` wrappers were checked through their underlying commands.

## Checklist

- [x] Reviewed the implementation and preserved existing no-pipeline behavior
- [x] Added regression tests and public API type checks
- [x] Updated the cluster and pipeline documentation with a tested example
- [x] Ran the affected tests and static analysis

### DIFF
--- a/faststream/redis/broker/broker.py
+++ b/faststream/redis/broker/broker.py
@@ -43,6 +43,7 @@ if TYPE_CHECKING:
     from types import TracebackType
 
     from redis.asyncio.client import Pipeline, Redis
+    from redis.asyncio.cluster import ClusterPipeline
 
     from faststream._internal.basic_types import SendableMessage
     from faststream.redis.message import RedisChannelMessage
@@ -222,6 +223,21 @@ class RedisBroker(
         pipeline: Optional["Pipeline[bytes]"] = None,
     ) -> bytes: ...
 
+    @overload
+    async def publish(
+        self,
+        message: "SendableMessage" = None,
+        channel: str | None = None,
+        *,
+        reply_to: str = "",
+        headers: dict[str, Any] | None = None,
+        correlation_id: str | None = None,
+        list: str | None = None,
+        stream: str | None = None,
+        maxlen: int | None = None,
+        pipeline: "ClusterPipeline[bytes]",
+    ) -> "int | bytes | ClusterPipeline[bytes]": ...
+
     @override
     async def publish(
         self,
@@ -234,8 +250,8 @@ class RedisBroker(
         list: str | None = None,
         stream: str | None = None,
         maxlen: int | None = None,
-        pipeline: Optional["Pipeline[bytes]"] = None,
-    ) -> int | bytes:
+        pipeline: Optional["Pipeline[bytes] | ClusterPipeline[bytes]"] = None,
+    ) -> "int | bytes | ClusterPipeline[bytes]":
         """Publish message directly.
 
         This method allows you to publish a message in a non-AsyncAPI-documented way.
@@ -259,10 +275,11 @@ class RedisBroker(
             maxlen:
                 Redis Stream maxlen publish option. Remove eldest message if maxlen exceeded.
             pipeline:
-                Redis pipeline to use for publishing messages.
+                Redis pipeline to use for publishing messages. Cluster pipelines
+                support lists and streams; queued commands run on pipeline.execute().
 
         Returns:
-            int: The result of the publish operation, typically the number of messages published.
+            The publish result, or the cluster pipeline when the command is queued.
         """
         cmd = RedisPublishCommand(
             message,
@@ -278,7 +295,7 @@ class RedisBroker(
             message_format=self.message_format,
         )
 
-        result: int | bytes = await super()._basic_publish(
+        result: int | bytes | ClusterPipeline[bytes] = await super()._basic_publish(
             cmd,
             producer=self.config.producer,
         )
@@ -315,8 +332,8 @@ class RedisBroker(
         )
         return msg
 
-    @override
-    async def publish_batch(  # type: ignore[override]
+    @overload  # type: ignore[override]
+    async def publish_batch(
         self,
         *messages: "SendableMessage",
         list: str,
@@ -324,7 +341,29 @@ class RedisBroker(
         reply_to: str = "",
         headers: dict[str, Any] | None = None,
         pipeline: Optional["Pipeline[bytes]"] = None,
-    ) -> int:
+    ) -> int: ...
+
+    @overload
+    async def publish_batch(
+        self,
+        *messages: "SendableMessage",
+        list: str,
+        correlation_id: str | None = None,
+        reply_to: str = "",
+        headers: dict[str, Any] | None = None,
+        pipeline: "ClusterPipeline[bytes]",
+    ) -> "int | ClusterPipeline[bytes]": ...
+
+    @override
+    async def publish_batch(
+        self,
+        *messages: "SendableMessage",
+        list: str,
+        correlation_id: str | None = None,
+        reply_to: str = "",
+        headers: dict[str, Any] | None = None,
+        pipeline: Optional["Pipeline[bytes] | ClusterPipeline[bytes]"] = None,
+    ) -> "int | ClusterPipeline[bytes]":
         """Publish multiple messages to Redis List by one request.
 
         Args:
@@ -336,7 +375,7 @@ class RedisBroker(
             pipeline: Redis pipeline to use for publishing messages.
 
         Returns:
-            int: The result of the batch publish operation.
+            The batch publish result, or the cluster pipeline when it is queued.
         """
         cmd = RedisPublishCommand(
             *messages,
@@ -349,7 +388,7 @@ class RedisBroker(
             message_format=self.message_format,
         )
 
-        result: int = await self._basic_publish_batch(
+        result: int | ClusterPipeline[bytes] = await self._basic_publish_batch(
             cmd,
             producer=self.config.producer,
         )
--- a/faststream/redis/broker/broker.py
+++ b/faststream/redis/broker/broker.py
@@ -205,7 +205,7 @@ class RedisBroker(
         list: str | None = None,
         stream: None = None,
         maxlen: int | None = None,
-        pipeline: Optional["Pipeline[bytes]"] = None,
+        pipeline: None = None,
     ) -> int: ...
 
     @overload
@@ -220,7 +220,7 @@ class RedisBroker(
         list: str | None = None,
         stream: str = ...,
         maxlen: int | None = None,
-        pipeline: Optional["Pipeline[bytes]"] = None,
+        pipeline: None = None,
     ) -> bytes: ...
 
     @overload
@@ -235,8 +235,8 @@ class RedisBroker(
         list: str | None = None,
         stream: str | None = None,
         maxlen: int | None = None,
-        pipeline: "ClusterPipeline[bytes]",
-    ) -> "int | bytes | ClusterPipeline[bytes]": ...
+        pipeline: "Pipeline[bytes]",
+    ) -> "Pipeline[bytes]": ...
 
     @override
     async def publish(
@@ -251,7 +251,7 @@ class RedisBroker(
         stream: str | None = None,
         maxlen: int | None = None,
         pipeline: Optional["Pipeline[bytes] | ClusterPipeline[bytes]"] = None,
-    ) -> "int | bytes | ClusterPipeline[bytes]":
+    ) -> "int | bytes | Pipeline[bytes] | ClusterPipeline[bytes]":
         """Publish message directly.
 
         This method allows you to publish a message in a non-AsyncAPI-documented way.
@@ -275,11 +275,11 @@ class RedisBroker(
             maxlen:
                 Redis Stream maxlen publish option. Remove eldest message if maxlen exceeded.
             pipeline:
-                Redis pipeline to use for publishing messages. Cluster pipelines
-                support lists and streams; queued commands run on pipeline.execute().
+                Redis pipeline to use for publishing messages. Queued commands run on
+                pipeline.execute().
 
         Returns:
-            The publish result, or the cluster pipeline when the command is queued.
+            The publish result, or the pipeline when the command is queued.
         """
         cmd = RedisPublishCommand(
             message,
@@ -295,7 +295,9 @@ class RedisBroker(
             message_format=self.message_format,
         )
 
-        result: int | bytes | ClusterPipeline[bytes] = await super()._basic_publish(
+        result: (
+            int | bytes | Pipeline[bytes] | ClusterPipeline[bytes]
+        ) = await super()._basic_publish(
             cmd,
             producer=self.config.producer,
         )
@@ -340,7 +342,7 @@ class RedisBroker(
         correlation_id: str | None = None,
         reply_to: str = "",
         headers: dict[str, Any] | None = None,
-        pipeline: Optional["Pipeline[bytes]"] = None,
+        pipeline: None = None,
     ) -> int: ...
 
     @overload
@@ -351,8 +353,8 @@ class RedisBroker(
         correlation_id: str | None = None,
         reply_to: str = "",
         headers: dict[str, Any] | None = None,
-        pipeline: "ClusterPipeline[bytes]",
-    ) -> "int | ClusterPipeline[bytes]": ...
+        pipeline: "Pipeline[bytes]",
+    ) -> "Pipeline[bytes]": ...
 
     @override
     async def publish_batch(
@@ -363,7 +365,7 @@ class RedisBroker(
         reply_to: str = "",
         headers: dict[str, Any] | None = None,
         pipeline: Optional["Pipeline[bytes] | ClusterPipeline[bytes]"] = None,
-    ) -> "int | ClusterPipeline[bytes]":
+    ) -> "int | Pipeline[bytes] | ClusterPipeline[bytes]":
         """Publish multiple messages to Redis List by one request.
 
         Args:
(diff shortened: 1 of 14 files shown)


lxingy3:
The Redis and Redis Cluster integration jobs passed, including the new transaction tests. The Kafka job finished with 359 passed and one failure in `tests/brokers/kafka/future/test_fastapi.py::TestRouter::test_base_real`.

That case did not receive its message within the existing three-second wait; aiokafka logged that the generated topic was not found in cluster metadata. The same inherited test passed later in the job as `TestKafkaRouter::test_base_real`. This PR does not change Kafka, shared execution code, dependencies, or the workflow.

Could a maintainer rerun the failed jobs in [this run](https://github.com/ag2ai/faststream/actions/runs/33902541144)? I only have read access to the upstream repository, so I cannot trigger the rerun.

IvanKirpichnikov (changes requested):
In FastStream, we support three types of Redis brokers:
+ Default Redis
+ Redis Cluster
+ Redis Sentinel

For regular Redis and Redis Sentinel, the pipeline type is `redis.asyncio.client.Pipeline`, while for Redis Cluster it is `redis.asyncio.cluster.ClusterPipeline`.

We need to add the correct type hints for the pipeline argument in the following methods:
+ `publish`
+ `request`
+ `publish_batch`
+ e.g...

We also need to add overloads for these methods so that the return type depends on whether a pipeline is passed.

The expected behavior is:
+ Without pipeline → keep the existing return type.
+ With pipeline → return the corresponding broker pipeline type, as in redis-py.

In other words, the overloads should reflect the following behavior:
+ For regular Redis and Sentinel: `redis.asyncio.client.Pipeline`
+ For Redis Cluster: `redis.asyncio.cluster.ClusterPipeline`

lxingy3:
Addressed in `679e5152`:

- `RedisBroker` and `RedisSentinelBroker` now expose `Pipeline` overloads, while `RedisClusterBroker` exposes `ClusterPipeline` overloads.
- Calls without a pipeline keep their existing `int`/`bytes` return types; calls with a pipeline return that exact pipeline type. Publisher overloads follow the same rule.
- Runtime identity checks now cover regular and cluster pipelines, and the public types are covered by mypy and Pyright assertions.

I intentionally kept `request` pipeline-free. A request subscribes, publishes, and waits for the reply in the same call; queueing that publish would prevent the caller from executing the pipeline until after `request` returns, so it would time out.

@IvanKirpichnikov, could you take another look?

faststream/redis/broker/broker.py
```diff
@@ -234,8 +250,8 @@ async def publish(
         list: str | None = None,
         stream: str | None = None,
         maxlen: int | None = None,
-        pipeline: Optional["Pipeline[bytes]"] = None,
-    ) -> int | bytes:
+        pipeline: Optional["Pipeline[bytes] | ClusterPipeline[bytes]"] = None,
+    ) -> "int | bytes | Pipeline[bytes] | ClusterPipeline[bytes]":
```
IvanKirpichnikov, line 254: only Pipeline. It can only be used with ClusterRedis

faststream/redis/broker/broker.py
```diff
@@ -278,7 +295,9 @@ async def publish(
             message_format=self.message_format,
         )
 
-        result: int | bytes = await super()._basic_publish(
+        result: (
+            int | bytes | Pipeline[bytes] | ClusterPipeline[bytes]
```
IvanKirpichnikov, line 299: only Pipeline. It can only be used with ClusterRedis

faststream/redis/broker/broker.py
```diff
@@ -315,16 +334,38 @@ async def request(  # type: ignore[override]
...
         self,
         *messages: "SendableMessage",
         list: str,
         correlation_id: str | None = None,
         reply_to: str = "",
         headers: dict[str, Any] | None = None,
-        pipeline: Optional["Pipeline[bytes]"] = None,
-    ) -> int:
+        pipeline: Optional["Pipeline[bytes] | ClusterPipeline[bytes]"] = None,
+    ) -> "int | Pipeline[bytes] | ClusterPipeline[bytes]":
```
IvanKirpichnikov, line 368: only Pipeline. It can only be used with ClusterRedis

faststream/redis/broker/broker.py
```diff
@@ -349,7 +390,9 @@ async def publish_batch(  # type: ignore[override]
             message_format=self.message_format,
         )
 
-        result: int = await self._basic_publish_batch(
+        result: (
+            int | Pipeline[bytes] | ClusterPipeline[bytes]
```
IvanKirpichnikov, line 394: only Pipeline. It can only be used with ClusterRedis

tests/docs/redis/test_pipeline.py
IvanKirpichnikov: Let’s use ClusterRedisTestcaseConfig to write tests.

tests/brokers/redis/cluster/test_cluster_pipeline.py
```diff
@@ -0,0 +1,234 @@
...
+                ] == [b"immediate", b"queued"]
+            finally:
+                await client.delete(counter, stream)
+
+    async def test_cross_slot_transaction_rejected(self, queue: str) -> None:
+        """Fixes https://github.com/ag2ai/faststream/issues/3044.
+
+        A cross-slot transaction must fail without publishing outside it.
+        """
+        from redis.exceptions import CrossSlotTransactionError
```
IvanKirpichnikov, line 215: Make the import at the module level.

tests/brokers/redis/test_publish.py
```diff
@@ -250,7 +253,10 @@ async def test_publish_batch_with_pipeline(
 
         @broker.subscriber(channel=queue)
         async def m(msg: str, pipe: Pipeline) -> None:
-            await broker.publish_batch(*range(5), list=queue + "resp", pipeline=pipe)
+            result = await broker.publish_batch(
+                *range(5), list=queue + "resp", pipeline=pipe
+            )
+            assert result is pipe
```
IvanKirpichnikov, line 259: I don’t think these changes are necessary.

tests/brokers/redis/test_publish.py
```diff
@@ -215,10 +215,13 @@ async def test_publish_with_pipeline(
...
                 # publish 5 messages by publisher
-                await publisher.publish(None, pipeline=pipe)
+                publisher_result = await publisher.publish(None, pipeline=pipe)
 
                 # and 5 by broker
-                await broker.publish(None, **destination, pipeline=pipe)
+                broker_result = await broker.publish(None, **destination, pipeline=pipe)
+
+                assert publisher_result is pipe
+                assert broker_result is pipe
```
IvanKirpichnikov, line 224: I don’t think these changes are necessary.

tests/brokers/redis/cluster/test_cluster_publish.py
```diff
@@ -90,6 +90,26 @@ async def handler(msg) -> None:
...
+                messages = await client.lrange(queue, 0, -1)
+                assert [broker.message_format.parse(msg)[0] for msg in messages] == [
+                    b"one",
+                    b"two",
+                ]
+            finally:
+                await client.delete(queue)
+        finally:
+            await broker.stop()
+
```
IvanKirpichnikov, line 112:
```python
broker = self.get_broker()

async with self.patch_broker(broker):
    client = broker.config.broker_config.connection.client
    raw_messages = await client.lrange(queue, 0, -1)
    messages = [broker.message_format.parse(raw_msg)[0] for raw_msg in raw_messages]

assert messages == [b"one", b"two"]
```

tests/brokers/redis/cluster/test_cluster_pipeline.py
```diff
@@ -0,0 +1,234 @@
...
+                    pipe.incr(counter)
+                    await broker.publish("one", stream=stream, pipeline=pipe)
+                    with pytest.raises(CrossSlotTransactionError):
+                        await pipe.execute()
+
+                assert await client.get(counter) is None
+                assert await client.xlen(stream) == 0
+            finally:
+                await client.delete(counter)
+                await client.delete(stream)
```
IvanKirpichnikov, line 234:
```python
broker = self.get_broker()
counter = f"{{first}}:{queue}:count"
stream = f"{{second}}:{queue}:stream"

async with self.patch_broker(broker):
    client = broker.config.broker_config.connection.client

    async with client.pipeline(transaction=True) as pipe:
        pipe.incr(counter)
        await broker.publish("one", stream=stream, pipeline=pipe)
        with pytest.raises(CrossSlotTransactionError):
            await pipe.execute()

    assert await client.get(counter) is None
    assert await client.xlen(stream) == 0
```

faststream/redis/publisher/producer.py
```diff
@@ -69,8 +73,41 @@ async def publish_batch(self, cmd: "RedisPublishCommand") -> int:
...
+        elif cmd.destination_type is DestinationType.List:
+            result = connection.rpush(cmd.destination, msg)
+        elif cmd.destination_type is DestinationType.Stream:
+            result = connection.xadd(
+                name=cmd.destination,
+                fields={DATA_KEY: msg},
+                maxlen=cmd.maxlen,
+            )
+        else:
+            raise UnreachablePathError
```
IvanKirpichnikov, line 104: maybe use `assert_never(cmd.destination_type)`

tests/brokers/redis/cluster/test_cluster_pipeline.py
```diff
@@ -0,0 +1,234 @@
...
+                assert results[0] == 1
+                assert len(results) == 2
+                assert await client.get(counter) == b"1"
+                entries = await client.xrange(stream)
+                assert [
+                    broker.message_format.parse(fields[b"__data__"])[0]
+                    for _, fields in entries
+                ] == [b"immediate", b"queued"]
+            finally:
+                await client.delete(counter, stream)
```
IvanKirpichnikov, line 208:
```python
broker = self.get_broker()
counter = f"{{{queue}}}:count"
stream = f"{{{queue}}}:stream"
publisher = broker.publisher(stream=stream)

async with self.patch_broker(broker):
    client = broker.config.broker_config.connection.client

    async with client.pipeline(transaction=True) as pipe:
        await pipe.watch(counter)
        result = await broker.publish(
            "immediate", stream=stream, pipeline=pipe
        )

        assert isinstance(result, bytes)
        assert await client.xlen(stream) == 1

        pipe.multi()
        pipe.incr(counter)
        await publisher.publish("queued", pipeline=pipe)

        assert await client.get(counter) is None
        assert await client.xlen(stream) == 1

        results = await pipe.execute()

    assert results[0] == 1
    assert len(results) == 2

    assert await client.get(counter) == b"1"

    entries = await client.xrange(stream)
    assert [
        broker.message_format.parse(fields[b"__data__"])[0]
        for _, fields in entries
    ] == [b"immediate", b"queued"]

```

tests/brokers/redis/cluster/test_cluster_pipeline.py
```diff
@@ -0,0 +1,234 @@
...
+                assert await client.get(counter) == b"1"
+                messages = await client.lrange(destination, 0, -1)
+                assert [broker.message_format.parse(msg)[0] for msg in messages] == [
+                    b"one",
+                    b"two",
+                    b"three",
+                    b"four",
+                ]
+            finally:
+                await client.delete(counter, destination)
```
IvanKirpichnikov, line 168:
```python
broker = self.get_broker()
counter = f"{{{queue}}}:count"
destination = f"{{{queue}}}:list"
publisher = broker.publisher(list=ListSub(destination, batch=True))

async with self.patch_broker(broker):
    client = broker.config.broker_config.connection.client

    async with client.pipeline(transaction=True) as pipe:
        pipe.incr(counter)

        await broker.publish_batch(
            "one", "two", list=destination, pipeline=pipe
        )
        await publisher.publish("three", "four", pipeline=pipe)

        assert await client.exists(counter, destination) == 0
        assert await pipe.execute() == [1, 2, 4]

    assert await client.get(counter) == b"1"

    messages = await client.lrange(destination, 0, -1)
    assert [broker.message_format.parse(msg)[0] for msg in messages] == [
        b"one",
        b"two",
        b"three",
        b"four",
    ]
```

tests/brokers/redis/cluster/test_cluster_pipeline.py
```diff
@@ -0,0 +1,234 @@
...
+                entries = await client.xrange(stream)
+                messages = [
+                    broker.message_format.parse(fields[b"__data__"])
+                    for _, fields in entries
+                ]
+                assert [body for body, _ in messages] == [b"one", b"two"]
+                assert messages[0][1]["correlation_id"] == queue
+                assert messages[0][1]["source"] == "broker"
+            finally:
+                await client.delete(counter, stream)
```
IvanKirpichnikov, line 134:
```python
broker = self.get_broker()
counter = f"{{{queue}}}:count"
stream = f"{{{queue}}}:stream"
publisher = broker.publisher(stream=stream)

async with self.patch_broker(broker):
    client = broker.config.broker_config.connection.client

    async with client.pipeline(transaction=True) as pipe:
        pipe.incr(counter)

        await broker.publish(
            "one",
            stream=stream,
            correlation_id=queue,
            headers={"source": "broker"},
            pipeline=pipe,
        )
        await publisher.publish("two", pipeline=pipe)

        assert await client.exists(counter, stream) == 0
        results = await pipe.execute()

    assert results[0] == 1
    assert len(results) == 3
    assert await client.get(counter) == b"1"

    entries = await client.xrange(stream)
    messages = [
        broker.message_format.parse(fields[b"__data__"])
        for _, fields in entries
    ]

    assert [body for body, _ in messages] == [b"one", b"two"]
    assert messages[0][1]["correlation_id"] == queue
    assert messages[0][1]["source"] == "broker"
```

IvanKirpichnikov (changes requested):

IvanKirpichnikov:
Hi. I’m closing your pull request because I didn’t receive any feedback. I created my own pull request: https://github.com/ag2ai/faststream/pull/3138

PR #3100, opened 2026-09-07
chore: three small clean-ups in one PR

Lancetnik:
# Description

Two clean-ups too small to review on their own, extracted from #3030 so the feature branch stops carrying them:

- **`PATH_CONTEXT_PREFIX`** — the context prefix a `Path()` parameter reads from (`"message.path."`) has exactly one definition, in `faststream/_internal/constants.py`, and `Path()` reads it from there. It was spelled inline before.
- **`parser_for(version)`** in `faststream/mqtt/parser.py` — the MQTT Subscriber and the in-memory MQTT test broker choose the parser class for a Broker version through this one function. Each picked it on its own before, and the two have to agree on which version they speak.

The third item the ticket lists, the `.gitignore` entries for local agent tooling and the graphify output, is already on `main`, so it is not part of this PR.

Nothing a user can see changes. No new tests: the change is one definition replacing two identical ones, and the existing MQTT and `Path()` suites pin the behaviour.

Part of the prefactor plan for #2451.

## Type of change

- [x] Refactor (no functional change)

## Checklist

- [x] Existing tests pass: `tests/utils/context/test_path.py`, `tests/brokers/mqtt/test_path.py`, `tests/brokers/mqtt/test_testclient.py`, `tests/docs/mqtt/message/test_message.py` (`-m "not connected"`; MQTT `connected` tests run in CI only)
- [x] `just mypy` passes

### DIFF
--- a/faststream/_internal/constants.py
+++ b/faststream/_internal/constants.py
@@ -23,3 +23,6 @@ class EmptyPlaceholder:
 
 
 EMPTY: Any = EmptyPlaceholder()
+
+PATH_CONTEXT_PREFIX = "message.path."
+"""Where a `Path()` parameter reads from: the Path parameters an Address captured."""
--- a/faststream/mqtt/broker/broker.py
+++ b/faststream/mqtt/broker/broker.py
@@ -3,14 +3,13 @@ from collections.abc import Awaitable, Callable, Iterable, Sequence
 from typing import (
     TYPE_CHECKING,
     Any,
-    Literal,
     Optional,
 )
 from urllib.parse import urlsplit
 
 import zmqtt
 from fast_depends import Provider, dependency_provider
-from typing_extensions import override
+from typing_extensions import assert_never, override
 
 from faststream._internal.broker import BrokerUsecase
 from faststream._internal.constants import EMPTY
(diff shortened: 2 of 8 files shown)


borisalekseev (changes requested):
Better to narrow the mqtt version string type to `Literal["3.1.1", "5.0"]` and use `typing.assert_never`. This will simplify the work when adding a protocol version, and mypy will highlight places where corrections need to be made.

Lancetnik:
Done in the second commit.

- `MQTTVersion = Literal["3.1.1", "5.0"]` lives once, in `faststream/mqtt/parser.py`, and `MQTTBroker`, `MQTTRouter` (FastAPI), `MQTTBrokerConfig` and the test broker spell their version through it.
- `parser_for` and the producer choice in `MQTTBroker.__init__` switch over it with `typing_extensions.assert_never` (the project supports 3.10, so not `typing.assert_never`). Adding a version is now an edit to the alias, and mypy names both switches.
- The one place a third value exists is the config: `version: MQTTVersion | Literal["unset"]`, "unset" being a Subscriber declared on a Router before a Broker composes its version in. The Subscriber resolves that to 5.0 explicitly, where the comment already explained it, instead of the previous `getattr(..., "5.0")` swallowing it.

`just mypy` and `just pyright` are clean; the MQTT test files pass locally without `connected`.

IvanKirpichnikov (approved):

borisalekseev (approved):

PR #3107, opened 2026-09-08
Add new benchmarks

### LINKED ISSUE #3084: Add new benchmarks
We need to add the folowing bench-cases:

- [x] raw client / raw Faststream
- [x] client + pydantic / Faststream + pydantic
- [ ] client + pydantic + OTEL + prom / FastStream + pydantic + OTEL + prom
- [ ] client + pydantic + OTEL + prom + SELECT (postgres) / FastStream + pydantic + OTEL + prom + SELECT (postgres)

These benchmarks should follow the current design

ApusBerliozi:
# Description

Two new scenarios per broker (Kafka, Confluent, NATS, RabbitMQ, Redis), each as a FastStream-vs-raw-client pair:

1. `test_metrics` — consume/echo loop with observability enabled: `*PrometheusMiddleware` + `*TelemetryMiddleware` on the FastStream side, and the equivalent Prometheus MetricsManager calls + an OpenTelemetry process span/histogram hand-rolled around the raw-client handler. Measures the throughput cost of metrics + tracing.

2. `test_sql` — same loop plus a Postgres lookup (SELECT ... FROM users WHERE name = $1) on every message via asyncpg, sharing one self.sql_pool per case. Measures throughput when the handler does real I/O.

The existing `test_basic / test_pydantic / test_msgspec` cases were also restructured to the same `TestFaststream<Broker><Variant>Case / TestPure<Broker><Variant>Case` layout (the standalone `test_aiokafka / test_nats / test_aiopika / test_confluent` files were folded in), so every scenario now compares FastStream against the underlying client directly.

Supporting pieces:

- `metrics.py` — shared Prometheus CollectorRegistry (scrape endpoint on :8001) and an OTLP tracer provider exporting to http://localhost:4318/v1/traces.
- `sql.py` — find_user_by_name(name, pool) helper and the BENCHMARK_PG_DSN config.
- `seed_db.sql` — creates and seeds the users table (John/Mike + 10k filler rows, indexed on name); auto-runs via the postgres service.
- `docker-compose.yaml` — added postgres (mounts the seed script) and otel-collector services.
- `pyproject.toml` — new benchmarks dependency group with the broker clients, asyncpg, msgspec, OTLP exporter, and test runner.

Each `*_cases/__init__.py ` now re-exports its case classes and a `CASES` dict for iteration.

Fixes #3084

## Type of change

- [X] New feature (a non-breaking change that adds functionality)

## Checklist

- [X] I have ensured that static analysis tests are passing by running `just static-analysis`
- [X] I have included code examples to illustrate the modifications


docker-compose.yaml
borisalekseev, line 1: Please, add custom compose profile or separate compose for benchmarks

Lancetnik (changes requested):
I would not merge new scenarios on top of the current harness: the ping-pong loop measures broker round-trip, not FastStream, and in its current state the branch does not run. `test_sql` as the "everything on" case (pydantic + otel + prometheus + postgres) is the right scenario and stays.

### Blocking

- `benchmarks/kafka_cases/test_basic.py:24`, `:69` and every other case: `setup_method` became `async`. Neither pytest nor `bench.py:69` awaits it, so `self.EVENTS_PROCESSED` and `self.broker` never exist and every case fails with `AttributeError`. The "tests pass" checkbox cannot be true.
- `benchmarks/bench.py:65` imports `rabbit_cases.test_aiopika.TestRabbitCase`, which this PR deletes. The new `*_CASES` dicts are not used anywhere.
- `benchmarks/kafka_cases/test_basic.py:90`: the raw consumer uses `auto_offset_reset="earliest"` with no `group_id`, FastStream defaults to `latest`. After one 10-minute run the topic holds a few hundred thousand echoes, so the raw case starts with all of them in flight and the FastStream case with one. The two numbers are not comparable.
- `benchmarks/README.md:16` still installs `faststream==0.6.0rc0` from PyPI. That benchmarks a release, not the branch.
- `benchmarks/kafka_cases/test_sql.py:11-16` mixes import roots: `from metrics import` needs `benchmarks/` on `sys.path`, `from benchmarks.sql import` needs the repo root.

nit: `test_sql.py:41` has the `comment` copied from `test_metrics`, so the two rows are indistinguishable in the CSV. nit: `metrics.py:10` starts an HTTP server at import time; `otel-collector` is pinned to `latest`.

### Methodology

With one message in flight, `EPS = 1 / RTT`. FastStream, the middlewares and even the Postgres query add a few serial milliseconds on top of a round-trip the broker owns. CPU never becomes the bottleneck, and CPU is where library overhead lives. So the harness needs two families of runs, labelled separately in the results and never compared to each other.

**Drain.** A pre-fill step publishes `N` messages, the consumer drains them, we measure how fast. Storage transports only: Kafka / Confluent topic, NATS **JetStream** stream, **Redis Streams**, durable RabbitMQ queue. `broker_type` says which ("NATS JetStream", "Redis Streams").

- Pre-fill is a separate script, not part of the case. It verifies the count afterwards (end offsets, `nats stream info`, `XLEN`, queue `message_count`) and that number is the run's `N`.
- Start with 200k to 1M and tune per broker. 10M pushes RabbitMQ into paging and flow control, and then we measure the broker under memory pressure.
- Fresh topic / stream / queue per run, unique name or explicit delete. Otherwise the leftover problem from `test_basic.py:90` comes back.
- Every payload carries a sequence number. The run ends when message `N` is processed, not on a timer. The sequence numbers also give us drops and duplicates.
- Same work in the handler on both sides: parse, count, the SQL query in the sql case. No echo publish, or we are measuring the producer again.
- Redis: no `MAXLEN ~` on pre-fill, approximate trimming drops part of the backlog silently. 1M × ~250 B is a few hundred MB of Redis memory, budget it in compose.

**Latency.** The current echo loop stays as a latency benchmark for core NATS and Redis pub/sub, which have no storage, and reports per-message RTT percentiles instead of EPS.

One subscriber, one queue, one process per run. Several subscribers or queues are a separate scenario later (10 subscribers on 10 subjects to see dispatch cost), not part of the baseline.

### What to record per run

| Column | How |
|---|---|
| `messages_per_s` | `N / (t_last - t_first)`, connect and subscribe excluded |
| `cpu_ms_per_1k` | `psutil.Process().cpu_times()` user+system delta between first and last message, divided by `N / 1000`. The most stable overhead signal we can get, and it shows even when broker-bound |
| `peak_rss_mb` | `resource.getrusage(RUSAGE_SELF).ru_maxrss` at the end |
| `rss_mb_median` | `memory_info().rss` sampled once a second, median |
| `rtt_p50_ms`, `rtt_p95_ms`, `rtt_p99_ms` | latency family only: publish timestamp in the payload, `perf_counter()` on receive, all samples kept, percentiles at the end |
| `dropped`, `duplicated` | from sequence numbers |
| `scenario`, `transport`, `implementation`, `N`, `run_idx`, `prefetch`, `batch`, `ack_mode` | so one row is reproducible on its own |

Five repetitions per (scenario, transport, implementation), median and stddev in the summary. Warm-up either as a separate 10k run or by discarding the first ~5% of messages. Consumer pinned to its own core (`psutil.Process().cpu_affinity`), broker, Postgres and otel-collector limited to other cores through compose `cpus`. Right now the `BatchSpanProcessor` export thread and Postgres share the core with the code under test.

### Equal settings on both sides

Defaults differ between FastStream and the raw clients, so every one of these is set explicitly in both implementations:

| Transport | Must match |
|---|---|
| Kafka / Confluent | `auto_offset_reset="earliest"`, fresh `group_id` per run, `max_poll_records`, `fetch_max_bytes` / `max_partition_fetch_bytes`, one partition (or partitions == `max_workers`), commit mode |
| NATS JetStream | consumer type: FastStream is push unless `pull_sub=` is given, for drain use pull on both sides; `DeliverPolicy.All`, `max_ack_pending`, ack policy, pull batch size, fresh durable per run |
| Redis Streams | `XREAD` vs `XREADGROUP`, same on both; start id `0-0`, or group created at `0` (`StreamSub` starts at `$`, or `>` with a group); same `COUNT` (FastStream: `batch=True` + `max_records`) and `BLOCK`; ack per message or none |
| RabbitMQ | `prefetch_count`, queue type (classic / quorum), durable, ack mode (`msg.process()` vs FastStream ack) |

Handler concurrency has to be equal too. aio-pika and FastStream both run handlers concurrently up to prefetch, the Kafka loops are sequential on both sides. In the sql case the asyncpg pool must be at least `prefetch` on both sides, or the pool becomes the bottleneck for whichever side has more concurrency.

Instrumentation as well: the FastStream middlewares also emit publish spans and metrics with context propagation, the hand-rolled raw side only instruments `process`. Without the echo publish most of that difference disappears, but check that the raw side records the same set of spans and metrics per consumed message as the middleware does.

### How to land it

Two PRs. First the harness: pre-fill script, the columns above, one broker (Kafka is the simplest to verify) with both implementations, `bench.py` iterating the cases, README running the branch through the `benchmarks` dependency group. Then port the other four brokers and the sql / metrics scenarios onto it. Doing all of it in this PR is possible, but the diff is already 3.8k lines and the harness change is the part that needs review.

Thanks for taking the benchmarks on. The direction, FastStream against the raw client for every scenario, is the right one.

Lancetnik:
Kafka is on the right method now: pre-fill, drain to `N`, no echo publish, `earliest` on both sides, sequence numbers, the CSV columns. It is not ready to produce numbers yet. Four things stop a run from finishing or make it measure the wrong build, six skew the FastStream / raw comparison, and the timing needs to move into the case. All of it below is Kafka only, at `56dc45ffa2`.

### Blocking

- `kafka_cases/test_basic.py:86`: `asyncio.gather(*(send(i) for i in range(n)))` creates `n` tasks up front. A stripped-down copy of that loop (same payload and semaphore, no aiokafka) peaks at 330 MB for 200k and 1.4 GB for 1M on my machine. `benches.sh:54` runs the container with `--memory="256m"`, so the smallest `N` gets OOM-killed. A plain `for` loop with `await producer.send(...)` and one `flush()` at the end fixes it and keeps `_seq` in order.
- `test_basic.py:45-48`: delete then create the same topic. Deletion is asynchronous in Kafka, and aiokafka 0.13 `create_topics` / `delete_topics` return the response without raising on per-topic errors. So a create that loses the race fails silently, and the `except UnknownTopicOrPartitionError` at `:49` never fires. Use a unique topic per run (`bench-<uuid>`), check the error codes in the response, delete the topic at the end.
- `bench.py:116`: `while target > case.EVENTS_PROCESSED` has no timeout. One lost message and the run hangs forever, which is exactly the case `dropped` exists to report.
- `benches.sh:55-63` mounts only `benchmarks/` and installs `faststream` from PyPI, so the run measures a release and writes its `__version__` into the CSV. The dependency list also exists three times: the `benchmarks` group in `pyproject.toml`, `requirements-bench.txt`, and the heredoc at `benches.sh:22-37`. Mount the repo root, `uv sync --group benchmarks`, drop the other two.

### Comparison skew

- `test_pydantic.py:37`, `test_msgspec.py:42`, `test_metrics.py:62`, `test_sql.py:65`: the FastStream handler runs `json.loads(raw.body.decode())` just to feed `_track_message`, after FastStream already decoded the body. The raw side parses once and reuses `data`. That is one extra `json.loads` per message charged to FastStream only. Put `_seq` into the Kafka message key, both sides read it without touching the body.
- `test_msgspec.py:101`: `Schema(**{...})` is the `Struct` constructor. It validates nothing and leaves `children` as dicts. The raw msgspec path is `msgspec.json.decode(msg.value, type=Schema)`, with no `json.loads` at all. Right now FastStream validates and the raw side does not.
- `test_basic.py:74`: pre-fill writes `name = "John-{i}"`, `seed_db.sql` holds `John`, `Mike` and `user_1..10000`. The query misses on every message but one. `f"user_{i % 10000 + 1}"` makes it a hit, which is what the "real world" case is for.
- `test_basic.py:20-21`: a `set` of up to 1M ints and a list of 1M floats live inside the process whose RSS we report. One partition means ordered delivery, so a single `expected_next` counter is enough: lower is a duplicate, higher is a gap.
- `_pub_ts` and the `rtt_*` columns: in a drain this is time spent sitting in the topic, not latency. Drop them here, they belong to the latency family. `perf_counter()` is also process-local and breaks the moment pre-fill moves to its own process.
- No `group_id` on either side means nothing is ever committed, so `ack_mode = "ack_first"` / `"auto_commit"` describe something that does not happen. The sides are equal, but a real consumer commits. Fresh `group_id` per run on both. Same for `prefetch = None`: set `max_poll_records` and `fetch_max_bytes` explicitly on both sides and write the values into the row.

### Measurement

Timing belongs in the case, not in the polling loop. `cpu_start` is taken before connect (`bench.py:112`). `elapsed_time` is taken after leaving `case.start()` (`bench.py:122`), so it includes broker shutdown with `graceful_timeout=10`. The loop wakes once a second, which on a 10-second run is up to 10% error. The handler should record `perf_counter()` and `cpu_times()` on the first message and on message `N`, then set an `asyncio.Event`; `bench.py` awaits that event with a timeout and keeps only the RSS sampling.

One subprocess per run. `ru_maxrss` is a process-lifetime maximum, and today every case plus every pre-fill runs in one process, so `peak_rss_mb` is wrong for every row after the first. `bench.py` becomes the orchestrator: pre-fill, verify the count against end offsets, spawn the case, read one JSON row back. That also separates the shared `registry`, the global `tracer_provider` and the `start_http_server(8001)` at `metrics.py:10`, and leaves `--cpus` applying to the consumer alone instead of being shared with the producer.

`REPEATS = 1` (`bench.py:27`) gives no median and no stddev. Make it 5, and interleave: the loop order is case → N → repeat, so every FastStream run happens before every raw run and any drift on the host lands on one side. Put the repeat loop outside and alternate implementations inside it. First repeat is the warm-up, or discard the first 5% of messages.

`BatchSpanProcessor` defaults to `max_queue_size=2048` and drops spans silently when the queue is full. Both sides are equal, but at drain speed part of the "tracing cost" is the drop path. Raise the queue or record the dropped count.

### Nits

- `test_basic.py:42`: `bootstrap_servers` argument ignored, the admin client hardcodes `localhost:9092`.
- `return message` in the FastStream handlers with no publisher attached. Remove.
- The pytest scaffolding is dead: `setup_method(self, prefill_messages)` would receive the method object under pytest and is never awaited, and `testpaths = ["tests"]` does not collect these files anyway. Drop the decorators and `test_consume_message`, `bench.py` is the runner.
- Linter is red: trailing whitespace after `prefetch = None` in four files, `print(f"Obrabotano ...")` at `bench.py:117`, Russian strings in `benches.sh` and `test_basic.py:50`. English only, please.
- README still describes the old flow.

Order I would do it in: subprocess per run with pre-fill in the orchestrator (removes the OOM, the RSS problem and the shared registry at once), then timing inside the case, then `_seq` in the key with the counter, then msgspec and the SQL hit, then repeats, linter, README.

PR #3110, opened 2026-09-09
feat(testing): `assert_called_once_with` for subscribers and publishers

Lancetnik:
# Description

Continues #2996 by @IvanKirpichnikov from a branch in this repository, so follow-up work can be stacked on top of it. The review thread and history live in #2996.

Fixes #557

Subscribers and publishers get `assert_called_once_with` under a test broker. It takes the body as a model, a dict or a matcher, and the message fields the endpoint saw: `headers`, `correlation_id`, `reply_to`, `content_type`, `path` and `context`.

```python
await handle.assert_called_once_with(
    AddUser(name="John", age=19),
    headers={"trace-id": "42"},
)
```

Both endpoint kinds share one `CallAssertions` surface backed by a `CallRecorder`. A publisher's recorder mirrors the recorder of the subscriber it is routed to, whether that subscriber is faked or real.

The scoping rule of `handle.mock` changes with it: the mock exists only inside the `Test*Broker` context. After the context exits, `handle.mock` raises `SetupError` instead of answering for calls nobody made. The testing docs describe the new rule and the new assertion.

`typing-extensions` is bumped to 4.14 for `Sentinel`.

## Type of change

- [x] New feature (a non-breaking change that adds functionality)
- [x] This change requires a documentation update

## Checklist

- [x] My code adheres to the style guidelines of this project (`just lint` shows no errors)
- [x] I have conducted a self-review of my own code
- [x] I have made the necessary changes to the documentation
- [x] My changes do not generate any new warnings
- [x] I have added tests to validate the effectiveness of my fix or the functionality of my new feature
- [x] Both new and existing unit tests pass successfully on my local environment by running `just test-coverage`
- [x] I have ensured that static analysis tests are passing by running `just static-analysis`
- [x] I have included code examples to illustrate the modifications

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3111, opened 2026-09-09
feat(testing): `assert_called_with` and `assert_any_call` beside `assert_called_once_with`

Lancetnik:
Stacked on #3110. Ticket 01 of the Call assertions plan.

#3110 gives a handler and a publisher one Call assertion, `assert_called_once_with`. A test that publishes twice had nothing to reach for: `handle.mock` covers counts and bodies, but nothing let a test say "the last message carried these headers" or "one of the messages carried this correlation id".

## What changes

Two more methods on every handler and publisher, named as in `unittest.mock` and taking the same arguments as `assert_called_once_with`:

- `assert_called_with(...)` checks the last Recorded call, as the mock's does.
- `assert_any_call(...)` walks the Recorded calls in order and passes on the first match. When none matches, one `AssertionError` lists every candidate with its own mismatch lines, numbered `call 1:`, `call 2:`.

All three now fail with `` `handle` was not called `` on an endpoint nobody called, from one shared path, instead of the `IndexError` the recorder raised on an empty list. `assert_called_once_with` keeps `mock.assert_called_once()` for the "called more than once" case only.

Internally, `CallAssertions` gets one private entry point with a `once | last | any` selector, and `CallRecorder.assert_last_call` becomes `mismatches(call, ...)`: a comparison over one Recorded call that returns its mismatch lines, so the `any` case can collect them per candidate. ADR-0009 is unchanged: headers as a subset, everything else exact, body through the codec, a batch refuses non-body fields.

## Tests and docs

- `tests/brokers/base/testclient.py` gains one test per behaviour (last call after two publishes, any call on either of two and the numbered failure, not called, publisher sharing the recorder) and the two new methods join the `SetupError` check outside a test broker. Every broker's `test_test_client.py` inherits them; MQTT 3.1.1 skips the three that need headers or correlation ids, as it already does for the existing ones.
- The "Validates Message Fields" section of `subscription/test.md` gains a paragraph and a tab block; `publishing/test.md` names the two methods. The six `advanced_testing.py` snippets gain a `test_several_messages` function, and the docs runner calls it.

### DIFF
--- a/faststream/_internal/testing/calls.py
+++ b/faststream/_internal/testing/calls.py
@@ -1,7 +1,7 @@
 from collections.abc import Mapping
 from copy import copy
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 from unittest.mock import MagicMock
 
 from faststream._internal.constants import EMPTY
@@ -15,7 +15,7 @@ if TYPE_CHECKING:
 
 
 class CallAssertions:
-    """The mock and the assertion an endpoint answers with under a test broker."""
+    """The mock and the Call assertions an endpoint answers with under a test broker."""
 
     __slots__ = ()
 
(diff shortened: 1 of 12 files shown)


faststream/_internal/testing/calls.py
```diff
@@ -58,20 +52,113 @@ async def assert_called_once_with(
...
+
+    def _recorder_under_test(self) -> "CallRecorder":
+        if not self.is_test:
+            msg = (
+                f"`{self._recorder.name}` is not under a test broker: "
+                "wrap the broker with its `Test*Broker` to use the mock "
+                "and the Call assertions."
+            )
+            raise SetupError(msg)
+        return self._recorder
```
IvanKirpichnikov, line 157: maybe make `_recorder_under_test` to property?

borisalekseev (approved):

IvanKirpichnikov (approved):

PR #3127, opened 2026-09-13
fix(redis): stop warning that StreamSub.no_ack has no effect with consumer groups

### LINKED ISSUE #3126: Bug: `StreamSub` warns that `no_ack` has no effect with a consumer group, but the flag is forwarded to XREADGROUP NOACK
**Describe the bug**

```python
```

**How to reproduce**

```python

broker = RedisBroker()


@broker.subscriber(
)
    ...
```

**Expected behavior**

**Observed behavior**

**History**

**Suggested fix**

**Environment**

HarshRajSinghania:
## Summary

`StreamSub(..., group=..., consumer=..., no_ack=True)` no longer emits
`RuntimeWarning: \`no_ack\` has no effect with consumer group` when `last_id` is `>` (the default for a consumer group).

Fixes #3126

## Motivation

The warning is leftover from a time when `no_ack` was not forwarded. It *does* have an effect: `stream_subscriber.py` passes `noack=stream.no_ack` into `XREADGROUP`, so entries never enter the PEL. The warning was incorrect and made a working option look broken.

The existing warning for `no_ack` with `last_id != ">"` is unchanged — that combination is still unsupported.

## Implementation

Removed the `elif no_ack:` `RuntimeWarning` branch in `StreamSub.__init__` that fired whenever a consumer group used the default `last_id` of `>`.

## Testing

- Isolated constructor check (no Redis): `StreamSub(..., group, consumer, no_ack=True)` raises no `RuntimeWarning` and sets `last_id == ">"`.
- Same constructor with `last_id="$"` still warns: `` `no_ack` is not supported by consumer group with last_id other than `>` ``.
- Added `test_stream_sub_no_ack_with_group_default_last_id` in `tests/brokers/redis/test_config.py`.
- Existing `test_stream_sub_with_no_ack_group` (last_id `$`) left as-is.

Could not install the full package in this environment (PyPI 502), so the Redis-marked suite was not run end-to-end here.

### DIFF
--- a/faststream/redis/schemas/stream_sub.py
+++ b/faststream/redis/schemas/stream_sub.py
@@ -115,13 +115,6 @@ class StreamSub(NameRequired):
                         stacklevel=1,
                     )
 
-            elif no_ack:
-                warnings.warn(
-                    message="`no_ack` has no effect with consumer group",
-                    category=RuntimeWarning,
-                    stacklevel=1,
-                )
-
         if claim_min_idle_time is not None:
             if not REDIS_V710:
                 msg = (
--- a/tests/brokers/redis/test_config.py
+++ b/tests/brokers/redis/test_config.py
@@ -158,3 +158,20 @@ def test_sub_overrides_broker_and_router() -> None:
         ack_policy=AckPolicy.ACK_FIRST,
     )
     assert sub.ack_policy is AckPolicy.ACK_FIRST
+
+
+@pytest.mark.redis()
+def test_stream_sub_no_ack_with_group_default_last_id() -> None:
+    """no_ack is forwarded to XREADGROUP NOACK when last_id is `>`."""
+    import warnings
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("error", RuntimeWarning)
+        stream = StreamSub(
+            "test_stream",
+            group="test_group",
+            consumer="test_consumer",
+            no_ack=True,
+        )
+    assert stream.no_ack is True
+    assert stream.last_id == ">"


Lancetnik:
@HarshRajSinghania we can't merge this until the CLA is signed — `license/cla` is still red: https://cla-assistant.io/ag2ai/faststream?pullRequest=3127

On the change itself, I want confirmation from a real run, not from reading the code. The description says the Redis suite wasn't executed here, and the whole PR rests on the claim that `no_ack` does something today. So: a consumer group subscriber with `no_ack=True` and the default `last_id` of `>`, against a live Redis, and `XPENDING <stream> <group>` showing the entries stay out of the PEL — next to the same subscriber with `no_ack=False`, where they land in it. Paste both outputs here.

Worth covering in the same run: `no_ack=True` on a stream sub also forces `AckPolicy.MANUAL` (`faststream/redis/subscriber/config.py:43`), so the two things should line up — nothing in the PEL, and FastStream not acking anything behind the NOACK flag.

Lancetnik:
Also: please drop `test_stream_sub_no_ack_with_group_default_last_id`. It asserts that the warning you deleted no longer fires — that's the diff restated, not a behaviour under test. It says nothing about the actual claim of this PR, that `noack` reaches `XREADGROUP` (`faststream/redis/subscriber/usecases/stream_subscriber.py:173`), and `assert stream.last_id == ">"` is the default `last_id` derivation, which this PR doesn't touch.

So the test goes, and the evidence this change needs is the `XPENDING` run above.

HarshRajSinghania:
@Lancetnik dropped `test_stream_sub_no_ack_with_group_default_last_id` as requested — it only asserted that the deleted warning no longer fires and that `last_id` stays `>`.

Pushed on this branch: `cbae4cda9bfe3d0d228aa0fd36fa381dc7162cf0`.

Still outstanding on my side:
- CLA / `license/cla` (I cannot sign that from here; the commit author also needs to be linked to the GitHub account).
- The live Redis `XPENDING` comparison you asked for (`no_ack=True` vs `False` against a real group, default `last_id` `>`). I do not have a Redis instance in this environment, so I have not pasted those outputs.

PR #1779, opened 2024-09-10
0.6.0

Lancetnik:
# Description

**FastStream 0.6** is a significant technical release that aimed to address many of the current project design issues and unlock further improvements on the path to version 1.0.0. We tried our best to minimize breaking changes, but unfortunately, some aspects were simply not working well. Therefore, we decided to break them in order to move forward.

This release includes:

* Finalized Middleware API
* Finalized Router API
* Introduced dynamic subscribers
* Added support for various serializer backends (such as [Msgspec](https://github.com/jcrist/msgspec))
* Support for AsyncAPI 3.0 specification
* A range of minor refactors and improvements

The primary goal of this release is to unlock the path towards further features. Therefore, we are pleased to announce that after this release, we plan to work on **MQTT** #956 and **SQS** #794 support and move towards version **1.0.0**!

## Breaking changes

Firstly, we have dropped support for **Python 3.8** and **Python 3.9**. **Python 3.9** is [almost at the end of its life](https://devguide.python.org/versions/) cycle, so it's a good time to update our minimum version.

### FastStream object changes

The broker has become a POSITIONAL-ONLY argument. This means that `FastStream(broker=broker)` is no longer valid. You should always pass the broker as a separate positional argument, like `FastStream(brokers)`, to ensure proper usage.

This is a preparatory step for `FastStream(*brokers)` support, which will be introduced in **1.0.0**.

### [AsyncAPI](https://faststream.ag2.ai/latest/getting-started/asyncapi/export/) changes

In **0.6**, you can't directly pass custom AsyncAPI options to the `FastStream` constructor anymore.

```python
app = FastStream(   # doesn't work anymore
    ...,
    title="My App",
    version="1.0.0",
    description="Some desctiption",
)
```

You need to create a `specification` object and pass it manually to the constructor.

```python
from faststream import FastStream, AsyncAPI

FastStream(
    ...
    specification=AsyncAPI(
        title="My App",
        version="1.0.0",
        description="Some desctiption",
    )
)
```

### Retry feature removed

Previously, you were able to configure retry attempts for a handler by using the following option:

```python
@broker.subscriber("in", retry=True)  # was removed
async def handler(): ...
```

Unfortunately, this option was a design mistake. We apologize for any confusion it may have caused. Technically, it was just a shortcut to `message.nack()` on error. We have decided that manual acknowledgement control would be more idiomatic and better for the framework. Therefore, we have provided a new feature in its place: `ack_policy` control.

```python
[...]

PR #3128, opened 2026-09-13
fix(redis): drop stale `no_ack` warning for consumer groups

### LINKED ISSUE #3126: Bug: `StreamSub` warns that `no_ack` has no effect with a consumer group, but the flag is forwarded to XREADGROUP NOACK

Thorsrud22:
## Description

Closes #3126.

`StreamSub(..., group=..., consumer=..., no_ack=True)` emits `RuntimeWarning: no_ack has no effect with consumer group`, but the flag has been forwarded to Redis as `XREADGROUP ... NOACK` since 0.6.0 (#1779), so it does have an effect and is the documented "acknowledge on read" mode. The warning is a leftover from 0.3.x when the flag really was dropped.

**Changes**

- Remove the `elif no_ack:` warning branch in `StreamSub.__init__`. With it gone the surrounding nested `if` collapses to `if group and consumer and last_id != ">":` (ruff SIM102), with no behaviour change for the two remaining warnings.
- `tests/brokers/redis/test_config.py`: constructing a `StreamSub` with a group and `no_ack=True` emits no warning (warnings are turned into errors inside the test).
- `tests/brokers/redis/test_consume.py`: connected test asserting every `XREADGROUP` call from such a subscriber carries `noack=True` and that `XPENDING` stays at 0 afterwards, following the `spy_decorator` pattern from `test_autoclaim.py`.

Left as is, per the issue: the `claim_min_idle_time` + `no_ack` `SetupError`, and the sibling "`no_ack` is not supported by consumer group with last_id other than `>`" warning, which the issue suggests checking separately.

One observation for a possible follow-up, not changed here: `get_one()` calls `_xreadgroup(count=1, block=...)` without passing `noack=self.stream_sub.no_ack`, so the flag is only honoured on the `start()` consume path.

**Verification**

- `ruff format` / `ruff check` clean on the three files; `mypy` clean on `stream_sub.py`.
- `pytest tests/brokers/redis/test_config.py tests/brokers/redis/test_schemas.py tests/brokers/redis/test_misconfigure.py -m "redis and not connected"`: 24 passed.
- The new connected test is collected but I don't have a Redis server locally, so it relies on CI here.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes


Thorsrud22:
Closing this one: I'd rather not sign a CLA, so I'll leave the fix here for anyone who wants to pick it up. The change is the one suggested in the linked issue, with tests. Thanks.

PR #3129, opened 2026-09-13
fix(redis): stop warning that `no_ack` has no effect with a consumer group

### LINKED ISSUE #3126: Bug: `StreamSub` warns that `no_ack` has no effect with a consumer group, but the flag is forwarded to XREADGROUP NOACK

Griger10:
# Description

Removed unused warning with "no_ack" elif branch. Added tests covering new behaviour. Now this doesnt show warning:
```python
StreamSub("s", group="g", consumer="c", no_ack=True)
```

Fixes #3126

## Type of change

## Checklist


Lancetnik:
@Griger10 I pushed one maintainer commit on top of yours and retitled the PR; please `git pull` before adding anything.

The fix itself I kept as you wrote it — flattening the nesting into `if group and consumer and last_id != ">"` is the right shape, and the two remaining warnings are untouched.

What I changed in the tests:

- `test_consume_group_with_no_ack_calls_xreadgroup` became `test_consume_group_with_no_ack_skips_pel`. Asserting `noack=True` in the `xreadgroup` kwargs pins the call signature of redis-py; what we actually promise is in the docs — "`no_ack=True` consumers are never tracked in the PEL at all" — and nothing covered that against a real Redis, which is how this warning survived for so long. The test now asserts `XPENDING` is 0 after the message is consumed. With `noack=stream.no_ack` reverted to `noack=False` it fails with `pending: 1`, so it bites on the real thing.
- The same test injected the `mock` fixture and never asserted on it, and had no `assert event.is_set()`, so it stayed green even if the handler never ran. It now ends with `mock.assert_called_once_with(...)`, which the PEL assertion needs anyway.
- Dropped `test_stream_sub_no_ack_correct_arguments_stored` — it checks that the constructor stored what it was handed, and `last_id == ">"` is the default derivation this PR doesn't touch. `test_stream_sub_no_ack_without_warning` is the regression guard; I verified it goes red when the `elif no_ack` branch is put back.
- `recwarn` got its annotation, the rest of that file annotates fixtures too.

Thanks for the PR, and especially for reaching for a real consume test instead of only asserting the warning is gone. Merging once CI is green.

Lancetnik:
Correction to my last comment: the `test_schemas.py` tests are gone entirely, not reshaped. `test_stream_sub_no_ack_without_warning` asserts that a deleted warning stays deleted, which pins the absence of code; the convention here runs the other way — a warning earns a test that it *fires*, and `test_stream_sub_with_no_ack_group` still does that for the `last_id != ">"` case this PR keeps. `test_schemas.py` is untouched by the PR now.

Final shape: the fix in `stream_sub.py` as you wrote it, plus one consume test asserting `XPENDING` is 0 after the message is delivered.

Two of my three commits cancel each other out — force-pushing over them wasn't available to me, and the PR squash-merges anyway, so the history in `main` stays clean. Please `git pull` before adding anything.

Lancetnik (approved):

PR #3131, opened 2026-09-14
feat(redis)!: native async cluster Pub/Sub on redis-py 8

### LINKED ISSUE #3009: Feature: migrate Redis Cluster pub/sub to native redis-py 8 async API

Lancetnik:
Closes #3009. Targets `1.0.0` rather than `main`, because it raises the `redis` floor.

## What this removes

Redis Cluster Pub/Sub has been running through a synchronous client since #2854, because `redis.asyncio.cluster` had no `publish` / `pubsub` at the time. redis-py 8.0.0 added them, so the workaround goes:

- `_SyncPubSubProxy`, `sync_publish()`, `_get_sync_cluster()` and the `ThreadPoolExecutor` in `configs/state.py` — along with the `_sync_cluster` / `_thread_pool` fields that the *base* `ConnectionState` carried for the cluster's sake
- `RedisClusterFastProducer` in `publisher/producer.py`, whose channel publish hopped to a thread on every message
- `_make_channel_subscriber()` and the `subscriber()` override in `broker/cluster_broker.py`, which existed only to monkeypatch `start()` on channel subscribers

Cluster now uses the same `RedisFastProducer` and the stock `ChannelSubscriber.start()` as standalone. `BaseRedisFastProducer` collapsed into `RedisFastProducer` since there is no longer a second producer to share with.

The `credential_provider` gap @fbartonoctopus reported on #3009 closes as a side effect: the hardcoded option whitelist in `_get_sync_cluster()` is gone, so every connection option reaches Channels, Lists and Streams through the one async client.

## The dependency decision

The issue left this open. This PR takes option 2 — **raise the floor to `redis>=8.0.0`** — rather than the runtime guard #2893 used. One supported path, and it lets `faststream/redis/_compat.py` go entirely: `REDIS_V710` and `REDIS_V720` are now always true, so the `claim_min_idle_time` version guard, the `lib_name`/`lib_version` fallback in `_get_driver_info()`, and the `require_redis_v710` test marker all disappear with it.

This is breaking for every Redis user, not just cluster users — hence `1.0.0`.

## One real bug the live cluster surfaced

`ClusterPubSub.execute_command` reads `nodes_manager.slots_cache` directly instead of going through `RedisCluster.execute_command`, so it never triggers the client's lazy topology discovery. A channel subscriber is usually the *first* cluster operation an app performs, and it died on `KeyError: <slot>`. `RedisClusterConnectionState.connect()` now awaits `client.initialize()`.

That makes `connect()` reach the network, which is what "connect" should mean — and it caught two tests that were only passing because the old connect was a no-op:

- `test_start_stop_lifecycle` ran `RedisClusterBroker()` against the default `localhost:6379` with no cluster there; it is now `connected`-marked and points at the real cluster
- `test_pipeline_warns` published without starting the broker, which worked only because the sync client built itself lazily from `_options`; it now starts the broker, matching what standalone `RedisBroker` already required

## Checked during implementation

The issue flagged three things worth confirming, since #2893 predates current `main`:

- **`legacy_responses=True`** — not needed. redis-py 8 already defaults `RedisCluster(legacy_responses=True)`, so RESP3 still yields RESP2-shaped responses and stream/list parsing is untouched. Passing it explicitly would be noise.
- **#2956 (`driver_info`)** — already landed; `_get_driver_info()` just loses its pre-7.2 branch.
- **Sharded Pub/Sub** — out of scope, as the issue says.

## Validation

- `ruff format` / `ruff check` / `codespell` / `mypy` (498 files) clean
- `pyright faststream/redis`: 136 → 125 errors (all pre-existing)
- `tests/brokers/redis` + `tests/docs/redis`, in-memory: pass
- `tests/brokers/redis -m redis` against a live standalone Redis: pass
- `tests/brokers/redis/cluster -m redis_cluster` against a live 3-node cluster (redis 8, ports 7001-7003): pass


faststream/redis/broker/cluster_broker.py
IvanKirpichnikov, line 124: I think it’s worth eliminating this behavior.

```suggestion
```

faststream/redis/configs/state.py
```diff
@@ -1,41 +1,25 @@
...
-            )
-        }
     return {
-        "lib_name": "faststream",
-        "lib_version": __version__,
+        "driver_info": DriverInfo(
+            name="faststream",
+            lib_version=__version__,
+        )
     }
```
IvanKirpichnikov, line 23: This function can be removed.
and pass DriverInfo as a parameter

IvanKirpichnikov (changes requested):

powersemmi (approved):

IvanKirpichnikov (approved):

PR #3139, opened 2026-09-16
feat: #3044 Add support pipe and remove types-redis

### LINKED ISSUE #3044: Feature:Support the `pipeline` parameter in RedisClusterBroker publish methods

IvanKirpichnikov:
# Description

Fixes #3044

## Type of change

## Checklist


docs/docs/en/getting-started/middlewares/index.md
```diff
@@ -261,11 +261,11 @@ To differentiate between different types of publishers, you can use `cmd.publish
     from faststream.redis import RedisPublishCommand
 
 
-    class RedisPublishMiddleware(BaseMiddleware[RedisPublishCommand]):
+    class RedisPublishMiddleware(BaseMiddleware[RedisPublishCommand[Any]]):
```
powersemmi, line 267: What's the point of Any here?

docs/docs/en/redis/cluster.md
```diff
@@ -53,8 +53,8 @@ broker = RedisClusterBroker(
 |---|---|---|
 | List | ✅ | ✅ |
 | Stream + XAUTOCLAIM | ✅ | ✅ |
-| Pub/Sub | ✅ | ✅ |
-| Pipeline | ✅ | ❌ |
+| Pub/Sub | ✅ | ✅ (via sync cluster) |
+| Pipeline | ✅ | ✅ (lists and streams) |
```
powersemmi, line 57: Pipeline does not work in cluster redis. Or has something changed in the new versions of Redis?

powersemmi: I looked, it is necessary to mention in the documentation that the atamarity of the pipeline in the cluster Redis does not provide in case the pipeline takes place on different slots

powersemmi: At the same time, in the streams, pipelines are guaranteed to work atamarity because the streams always occupy 1 slot

powersemmi (changes requested):

docs/docs/en/getting-started/middlewares/index.md
IvanKirpichnikov: This means that it doesn’t matter to us what type of pipeline we need: `Pipeline`, `ClusterPipeline`

IvanKirpichnikov (commented):
We need to reconsider the generic type with None.

PR #3163, opened 2026-09-19
refactor(asgi): let each TestBroker register itself for try-it-out

Lancetnik:
`TryItOutProcessor` used to import all six broker packages under `suppress(ImportError)` to map a broker to its `TestBroker`, so the core knew about every broker. Now each `TestBroker` subclass names its broker when it is defined:

```python
class TestKafkaBroker(TestBroker[KafkaBroker, EnterType], broker=KafkaBroker): ...
```

`find_test_broker()` resolves by `isinstance`, so `RedisClusterBroker` still gets `TestRedisBroker`. A third-party broker can now opt in to try-it-out by passing the same keyword.

Import cost doesn't change: `faststream.<broker>` already loads its `testing` module because it exports the `TestBroker`, which is where the registration happens.

Breaking the registration makes `tests/asgi/multibroker/test_try_it_out.py` fail.

Bottom of a stack; #3158 builds on it.


faststream/_internal/testing/broker.py
```diff
@@ -268,10 +277,23 @@ async def _fake_connect(self, broker: Broker, *args: Any, **kwargs: Any) -> None
...
 def patch_broker_calls(broker: "BrokerUsecase[Any, Any, Any]") -> None:
     """Patch broker calls."""
     for sub in broker.subscribers:
         sub._build_fastdepends_model()
 
         for h in sub.calls:
             h.handler.set_test()
+
+
+_TEST_BROKERS: dict[type[BrokerUsecase[Any, Any, Any]], type[TestBroker[Any, Any]]] = {}
```
IvanKirpichnikov, line 299: Should we add `Final` to the type hint and move it to the top?

IvanKirpichnikov (changes requested):

faststream/_internal/testing/broker.py
Lancetnik: We don't use `Final` annotations for contsant variables in the repo. I suggest to open a new PR to introduce the practice over all the modules

IvanKirpichnikov (approved):

PR #3178, opened 2026-09-19
chore: carry library typing fixes #3173-#3175 for the mypy stack

Lancetnik:
## Description

**Transit layer of the mypy stack — nothing to review here.**

The three commits are cherry-picks of the library typing fixes that the brokers layer (#3176) needs, each reviewed in its own PR against `main`:

- #3173 — `KafkaPublishMessage` accepted by `publish_batch()` / `BatchPublisher.publish()`
- #3174 — `ssl_*` options of `RedisClusterBroker`
- #3175 — batch subscribers and publishers resolve to one type, not a Union

They live in the stack only so that `mypy tests` and CI are green for the layers above. Once #3173–#3175 are in `main` and the stack is synced, these commits drop out as already applied and this PR closes empty.

## Type of change

- [x] Bug fix (duplicates of #3173, #3174, #3175)


Lancetnik:
Folded back into #3176: the library fixes ride there until #3173-#3175 land in main.

PR #3184, opened 2026-09-19
chore(mypy): enable five more error codes and fix their findings

Lancetnik:
## Description

Turns on five optional mypy error codes and fixes what they report in `faststream/` and `tests/mypy/` (62 findings):

| Code | Findings | What it was |
|---|---|---|
| `unused-awaitable` | 23 | the "forgot `await`" class; here all intentional fire-and-forget |
| `truthy-bool` | 20 | conditions on values that are never falsy |
| `truthy-iterable` | 9 | `if x:` on an `Iterable`, which a generator always passes |
| `possibly-undefined` | 6 | names bound in one branch and read in another |
| `redundant-expr` | 4 | conditions the types already decide |

### Behaviour

Two changes are more than annotations. Both close a latent bug with generators, neither changes a public signature:

- `dependencies=` of a broker or router is stored as a `tuple`. It was kept as passed, so a generator would be spent by the first subscriber and every later one would get no broker dependencies.
- `StartupValidationError` stores `missed_fields` and `invalid_fields` as tuples, for the same reason: `__str__` and the CLI renderer both read them.

`_build_middlewares_stack()` reads `extra_middlewares` into a tuple before `or`: a subscriber passes a generator there, which is truthy even when empty. A subscriber's stack is never empty, so nothing observable changes.

### The rest

- `TasksMixin.add_task()` returns `None`. No caller in `faststream/` read the task, which already sits in `self.tasks`; the four supervisor tests take it from there.
- `tg.start_soon(...)` (6) and `await self.connect()` on Redis (3) become `_ = ...`: anyio returns an awaitable `TaskHandle` and the Redis client is awaitable itself, and both are dropped on purpose.
- `prometheus/container.py`: `cast("Counter", ...) or Counter(...)` claimed the lookup never returns `None`; the cast is now `"Counter | None"`.
- AsyncAPI sqs bindings: `from_sub()` / `from_pub()` never return `None`, so the walrus check around them is gone.
- `RabbitSubscriber.exchange` and `RabbitExchange.validate()` are never `None`: three dead checks removed in `rabbit/testing.py` and `rabbit/subscriber/usecase.py`.
- The kafka and confluent `_validate_input_for_misconfigure()` helpers take `partitions` as `Collection` / `Sequence`, as the public `subscriber()` signatures already do.
- `basic_types.py` gets `ignore[possibly-undefined]`: mypy cannot tie the `HAS_PYDANTIC` / `HAS_MSGSPEC` flags to the guarded imports.

`deprecated` and `explicit-override` are left out on purpose. `ignore-without-code` and `exhaustive-match` report nothing and are enabled in #3177.

## Tests

No new tests: the codes are the check. `tests/mypy/confluent.py` now awaits the `no_confirm=True` future instead of dropping it, as `tests/mypy/kafka.py` does.

## Type of change

- [x] Refactoring (no behaviour change intended beyond the two generator cases above)


faststream/_internal/configs/broker.py
```diff
@@ -179,5 +179,5 @@ def broker_middlewares(self) -> Sequence["BrokerMiddleware[Any]"]:
         return [m for c in self.configs for m in c.broker_middlewares]
 
     @property
-    def broker_dependencies(self) -> Iterable["Dependant"]:
-        return (b for c in self.configs for b in c.broker_dependencies)
+    def broker_dependencies(self) -> Sequence["Dependant"]:
+        return [b for c in self.configs for b in c.broker_dependencies]
```
IvanKirpichnikov, line 183: maybe 
```suggestion
tuple(b for c in self.configs for b in c.broker_dependencies)
```

IvanKirpichnikov (dismissed):

IvanKirpichnikov (approved):

faststream/_internal/configs/broker.py
Lancetnik: I perefer to stay with list comprehensions because of prefomance. It requires a little extra memory, but much faster

PR #3191, opened 2026-09-19
Nats security

### LINKED ISSUE #780: Feature: NATS security support

C3EQUALZz:
# Description

This PR adds NATS security configuration through the `security` argument and closes #780, #2673. Previously, NATS authentication was configured through several independent `NatsBroker` constructor arguments. That API made mutually exclusive authentication mechanisms appear composable and provided no unified security model for connection configuration or AsyncAPI generation.

This change introduces dedicated NATS security strategies:

- `NatsSecurity` for TLS configuration without an authentication mechanism
- `NatsUserPassword` for native NATS username/password authentication
- `NatsToken` for static tokens and synchronous token callbacks
- `NatsNKey` for seed files and in-memory seeds
- `NatsCredentials` for combined credentials, separate JWT/seed files, and raw credentials
- `NatsJWT` for callback-based JWT and nonce-signing authentication

Only one authentication strategy can be selected for a broker. Credentials embedded in a server URL cannot be combined with an object-based authentication strategy.

The existing constructor arguments remain available for backward compatibility, but are now deprecated and will be removed in FastStream 1.0.0. When supplied, a deprecated authentication argument is adapted to the corresponding security object and takes precedence over `security=`.

The PR also:

- adds NATS-specific TLS options such as `tls_hostname` and `tls_handshake_first`;
- keeps `SASLPlaintext` compatibility while documenting that NATS uses native username/password authentication rather than SASL;
- adds AsyncAPI security schemes without exposing tokens, passwords, seeds, credential paths, or callback results;
- keeps `faststream[nats]` on the standard `nats-py>=2.14.0` dependency and reports how to install `nats-py[nkeys]` only when an NKey-backed security strategy is initialized;
- adds broker, AsyncAPI, and documentation tests;
- adds detailed documentation and runnable examples for every supported authentication mechanism.

Fixes #780, #2673

## Type of change

## Checklist


Lancetnik (commented):
Thanks for the PR! Two things before we go further:

1. **Every deprecation message has to name the version the thing goes away in.** For these arguments that's **1.0.0**. Right now the runtime warning in `warn_deprecated_security_args` says "will be removed in FastStream." with no version, and the `deprecated(...)` annotations on the `NatsBroker` arguments (`tls_hostname`, `token`, `signature_cb`, `user_jwt_cb`, `user_credentials`, `nkeys_seed`, `nkeys_seed_str`) don't mention one either. Please add "will be removed in 1.0.0" to all of them. The PR description says 0.9.0, so please fix that too.

2. **Please leave `faststream/nats/fastapi/fastapi.py` alone.** The whole FastAPI integration is already deprecated and goes away in 1.0.0 (see `faststream/nats/fastapi/__init__.py`), so new functionality shouldn't land there. Please revert the `NatsRouter` changes (the new `deprecated(...)` annotations, the legacy-argument warning and the `catch_warnings` block) and keep the feature to `NatsBroker`.

C3EQUALZz:
@Lancetnik fixed problems, my bad

pyproject.toml
```diff
@@ -73,7 +73,7 @@ confluent = [
     "confluent-kafka>=2.6,!=2.8.1,<3; python_version >= '3.13'",
 ]
 
-nats = ["nats-py>=2.12.0,<=3.0.0"]
+nats = ["nats-py[nkeys]>=2.14.0,<=3.0.0"]
```
powersemmi: And there is no such vibe to take it in a separate extra? Like nats and nats-nkeys

Lancetnik: The common pattern in the repo is staying with `nats-py` pure extra and raise an installation error only if user trying to initialize the `nkeys` extra object. See `PyYAML`, `uvicorn` for references

C3EQUALZz: @Lancetnik oki oki

C3EQUALZz: @Lancetnik @powersemmi fixed

faststream/nats/security.py
```diff
@@ -1,19 +1,465 @@
...
+        """Load separate NATS user JWT and seed files."""
+        security = cls(
+            jwt,
+            ssl_context=ssl_context,
+            use_ssl=use_ssl,
+            tls_hostname=tls_hostname,
+            tls_handshake_first=tls_handshake_first,
+        )
+        security.credentials = (str(jwt), str(seed))
+        return security
```
IvanKirpichnikov: why not
```suggestion
        return cls(
            (str(jwt), str(seed)),
            ssl_context=ssl_context,
            use_ssl=use_ssl,
            tls_hostname=tls_hostname,
            tls_handshake_first=tls_handshake_first,
        )
```

faststream/nats/security.py
```diff
@@ -1,19 +1,465 @@
...
+
+        security = cls(
+            "",
+            ssl_context=ssl_context,
+            use_ssl=use_ssl,
+            tls_hostname=tls_hostname,
+            tls_handshake_first=tls_handshake_first,
+        )
+        security.credentials = RawCredentials(credentials)
+        return security
```
IvanKirpichnikov: why not

```suggestion
        return cls(
            RawCredentials(credentials),
            ssl_context=ssl_context,
            use_ssl=use_ssl,
            tls_hostname=tls_hostname,
            tls_handshake_first=tls_handshake_first,
        )
```

faststream/nats/security.py
```diff
@@ -1,19 +1,465 @@
...
+
+        security = cls(
+            seed,
+            ssl_context=ssl_context,
+            use_ssl=use_ssl,
+            tls_hostname=tls_hostname,
+            tls_handshake_first=tls_handshake_first,
+        )
+        security.seed_is_file = False
+        return security
```
IvanKirpichnikov, line 332: Maybe it's better to Inject `seed_is_file` via `__init__`?

```suggestion
        return cls(
            seed,
            ssl_context=ssl_context,
            use_ssl=use_ssl,
            tls_hostname=tls_hostname,
            tls_handshake_first=tls_handshake_first,
            seed_is_file=False,
        )
```

faststream/nats/broker/broker.py
```diff
@@ -194,6 +206,119 @@ class NatsInitKwargs(TypedDict, total=False):
...
+        deprecated_arguments.append("user_jwt_cb")
+
+    if deprecated_arguments:
+        warn_deprecated_security_args(*deprecated_arguments)
+
+    if signature_callback_supplied != jwt_callback_supplied:
+        msg = "`user_jwt_cb` and `signature_cb` must be provided together."
+        raise SetupError(msg)
+    if signature_callback_supplied:
+        authentication_mechanisms.append("JWT callbacks")
```
IvanKirpichnikov:
```suggestion
    if signature_callback_supplied:
        deprecated_arguments.append("signature_cb")
        authentication_mechanisms.append("JWT callbacks")
    if jwt_callback_supplied:
        deprecated_arguments.append("user_jwt_cb")

    if deprecated_arguments:
        warn_deprecated_security_args(*deprecated_arguments)

    if signature_callback_supplied != jwt_callback_supplied:
        msg = "`user_jwt_cb` and `signature_cb` must be provided together."
        raise SetupError(msg)
```

faststream/nats/broker/broker.py
```diff
@@ -194,6 +206,119 @@ class NatsInitKwargs(TypedDict, total=False):
...
+    if len(authentication_mechanisms) > 1:
+        mechanisms = ", ".join(authentication_mechanisms)
+        msg = (
+            "Only one deprecated NATS authentication mechanism can be "
+            f"used at a time, but got: {mechanisms}."
+        )
+        raise SetupError(msg)
+
+    ssl_context = security.ssl_context if security is not None else None
+    use_ssl = security.use_ssl if security is not None else None
```
IvanKirpichnikov:
```suggestion
if security is not None:
    use_ssl = security.use_ssl
    ssl_context = security.ssl_context
else:
    use_ssl = None
    ssl_context = None

```

faststream/nats/broker/broker.py
```diff
@@ -194,6 +206,119 @@ class NatsInitKwargs(TypedDict, total=False):
 )
 
 
+def _adapt_deprecated_security(
```
IvanKirpichnikov, line 240: Let’s decompose this function

Let’s move the validation into a separate function that will be called in `_adapt_deprecated_security`

faststream/nats/security.py
IvanKirpichnikov, line 462: Why not represent these functions as methods of the corresponding classes?

IvanKirpichnikov (changes requested):

faststream/nats/security.py
C3EQUALZz: You're right, it will be better this way than mine, thx

faststream/nats/security.py
C3EQUALZz: Here I was rude myself, after all, writing code at night has a bad effect on attentiveness...

faststream/nats/security.py
C3EQUALZz: Won't we break the semantics, as is done in other brokers? Check please how in other brokers

faststream/nats/broker/broker.py
C3EQUALZz: Does it even make sense to decompose it? This function won't be around for long anyway, especially since we're already migrating to the type I proposed here in 1.0.0. Ultimately, this function will be removed, as we'll be getting rid of the unnecessary types.

faststream/nats/security.py
C3EQUALZz: I gave this some more thought, and I don’t think applying the same refactoring here is worth it. Passing seed_is_file to cls(...) would require exposing an implementation-specific flag in the public NatsNKey constructor solely for use by this factory method. It would also allow callers to create ambiguous combinations manually.                                                                                                                                                                                                                                                                                                                     

The assignment remains encapsulated inside from_seed, while the public constructor continues to clearly represent a file-backed seed. I’d therefore prefer to keep the existing implementation here.

faststream/nats/broker/broker.py
C3EQUALZz: The indentation are messed up here, but I got your point and will fix it right away

faststream/nats/broker/broker.py
C3EQUALZz: When I changed it to your if, ruff crashed. I'll decompose a little later...

C3EQUALZz: @IvanKirpichnikov hi, I refactored that problem, can you check please?

IvanKirpichnikov (approved):

PR #3200, opened 2026-09-20
chore(lint): check that declared __slots__ take effect with slotscheck

Lancetnik:
> Was the top of a stack; #3193, #3194 and #3199 are merged, so this now targets `main` directly. Rebased onto `main` after #3192 landed: one conflict, the import line of `faststream/redis/schemas/list_sub.py` (`cached_property` removed next to the added `Any`). No other change since the approval.

# Description

Adds [slotscheck](https://github.com/ariebovenberg/slotscheck) as a static check and fixes everything it finds, so the check starts green.

**The rule.** A class that declares `__slots__` only loses its `__dict__` if every base up the MRO declares `__slots__` too. One unslotted base and instances keep a `__dict__`: the declaration costs a line and buys nothing. Nothing flagged this shape before.

**What main looked like.** `slotscheck -m faststream` reported 35 × `defines slots but superclass does not` (all 510 modules import in the `dev` environment, no exclusions needed). The 35 classes hang off 10 unslotted bases:

| Unslotted base | Classes it disabled |
|---|---|
| `NameRequired` | `RabbitQueue`, `RabbitExchange`, `JStream`, `KvWatch`, `ListSub`, `PubSub`, `StreamSub`, confluent `Topic` |
| `BaseSecurity` | `SASLPlaintext`, `SASLScram256`, `SASLScram512`, `SASLOAuthBearer`, `SASLGSSAPI`, `RabbitExternalAuth` |
| `TelemetrySettingsProvider` / `MetricsSettingsProvider` (protocols) | 6 + 6 broker settings providers |
| `BrokerMiddleware` (protocol) | `TelemetryMiddleware`, `PrometheusMiddleware` |
| `ProducerState` (kafka), `ConnectionState` (nats) | `EmptyProducerState`, `RealProducer`, `EmptyConnectionState`, `ConnectedState` |
| `RabbitDeclarer`, `ChannelManager`, `Unsubscriptable` (protocols) | `RabbitDeclarerImpl`, `ChannelManagerImpl`, `UnsubscribeAdapter` |

**How it was fixed.** No `__slots__` declaration was deleted to quiet the tool.

- 8 `Protocol` bases got `__slots__ = ()`.
- `NameRequired` got `__slots__ = ("name",)` and `BaseSecurity` got `__slots__ = ("ssl_context", "use_ssl")`, since their `__init__` assigns those. The same names were removed from the subclasses' tuples (a name slotted twice is wasted space, and slotscheck reports it); `SASLOAuthBearer`, `SASLGSSAPI` and `RabbitExternalAuth` end up with `__slots__ = ()`.
- No dataclasses were involved.

Once the `__dict__` was gone it turned out to have been covering for attributes no slot declared. On main `vars(RabbitQueue("q"))` is `{'declare': True}`. These would have been `AttributeError` at construction, so the slot tuples were completed:

- `RabbitQueue`, `RabbitExchange`: added `declare`.
- `JStream`: added `subjects`.
- `RabbitDeclarerImpl`: slots were `"__queues"` / `"__exchanges"`, which mangle to `_RabbitDeclarerImpl__queues`, while the code uses `self._queues` / `self._exchanges`. Slots renamed to match the attributes.
- `ListSub.records` was a `functools.cached_property`, which stores into `__dict__` and raises `TypeError` without one. It is now a plain `property` (it is a one-line expression, and nothing in the repo reads it). One test added for it in `tests/brokers/redis/test_schemas.py`, because neither slotscheck nor any existing test goes red if it is turned back; it fails with that `TypeError` when I do.

Not changed: `KvWatch` declares a `bucket` slot that is never assigned (the value lives in `name`). It behaved the same before, so I left it.

`require-subclass` (the opposite direction: a class that could have slots and does not) reports 190 classes and is left off; e.g. `KafkaTelemetrySettingsProvider` and the other 13 unslotted provider subclasses still carry a `__dict__`. That is a separate decision.

## Behaviour change

These public classes no longer accept arbitrary attributes; `obj.anything = 1` now raises `AttributeError`, and they have no `__dict__` / `__weakref__`:

`RabbitQueue`, `RabbitExchange`, `JStream`, `KvWatch`, `ListSub`, `PubSub`, `StreamSub`, `faststream.confluent.Topic`, `BaseSecurity`, `SASLPlaintext`, `SASLScram256`, `SASLScram512`, `SASLOAuthBearer`, `SASLGSSAPI`, `RabbitExternalAuth`, `TelemetryMiddleware`, `PrometheusMiddleware`.

Also `ListSub.records` can no longer be assigned to, and follows `batch` / `max_records` instead of freezing on first read.

A user subclass that does not declare `__slots__` gets a `__dict__` back and is unaffected. The same holds inside the library: the broker middlewares users actually construct (`KafkaTelemetryMiddleware`, `RabbitPrometheusMiddleware`, ...) subclass without `__slots__`, so they keep theirs. There are no subclasses of these classes in `faststream/`, `tests/`, `docs/docs_src` or `examples/`. Nothing in the repo used `vars()`, `__dict__`, weakrefs or undeclared `setattr` on them. `copy`, `deepcopy` (used by `add_prefix`) and pickling work on the slotted instances; `RabbitQueue` and `PubSub` do not pickle, on main either, because `Address` holds a lambda.

## Wiring

Same shape as `bandit`:

- `pyproject.toml`: `slotscheck==0.21.0` in the `lint` group, `[tool.slotscheck]` with `strict-imports = true` (a module that fails to import is an error, not a skip) and `require-superclass = true`. `uv.lock` gains only `slotscheck`.
- `justfile`: `just slotscheck`, added to `just static-analysis`.
- `.pre-commit-config.yaml`: local `slotscheck` hook. The linters workflow (`pr_autoupdate.yaml`) installs the `dev` group and runs the pre-commit hooks on all files, so CI picks it up with no workflow change.
- One sentence each in the `code-architecture` and `dev-workflow` agent skills.

## Verification

- `just slotscheck`: `All OK! Scanned 510 module(s), 656 class(es)` on Python 3.14, and the same in isolated 3.10 and 3.12 environments.
- Bite test: removing `__slots__` from `NameRequired` gives 8 errors and exit code 1; restored, green again.
- A throwaway script instantiates all 35 flagged classes plus `BaseSecurity` and asserts no `__dict__` and that setting an unknown attribute raises: 35/35, on 3.10, 3.12 and 3.14. It also checks `copy` / `deepcopy` / `add_prefix` / `__eq__` / `__hash__` / `__repr__` on the schema classes.
- `SASLPlaintext` measured with tracemalloc over 10k instances: 128 bytes each on main, 72 on this branch.
- Tests: the 120 named test files that reference the touched classes (brokers, asyncapi, docs, examples, opentelemetry, prometheus), `-m "not connected and not slow"`: 1377 tests, 0 failures, 0 errors, 22 skipped. Plus the connected tests in `tests/brokers/{rabbit,nats,kafka}/test_connect.py` against local brokers: 8 passed.
- `just linter` clean; `mypy`: no issues in 1832 source files; `pyright`: 0 errors; `bandit`: no issues; `semgrep`: 0 findings.

Note: #3193 adds an `import-linter` recipe and hook in the same spots of `justfile` and `.pre-commit-config.yaml`; whichever merges second gets a trivial conflict.

## Type of change

- [x] Breaking change (a fix or feature that would disrupt existing functionality) — only for code that set undeclared attributes on the classes listed above

## Checklist

Follow-up for subclasses of the slotted classes (`require-subclass`): #3201.


powersemmi (dismissed):

faststream/redis/schemas/list_sub.py
```diff
@@ -62,7 +60,8 @@ def __init__(
         self.max_records = max_records
         self.polling_interval = polling_interval
 
-    @cached_property
+    # Not `cached_property`: it stores its value in `__dict__`, which slots remove.
```
IvanKirpichnikov: meaningless comment

IvanKirpichnikov (changes requested):
Why did we specify `__slots__ = ()` for the protocols, but not for those that inherit from these protocols?

By example `ChannelManager` and `FakeChannelManager`

faststream/redis/schemas/list_sub.py
Lancetnik: Removed in 744a7ff6fa.

Lancetnik:
Scope of this PR is "a declared `__slots__` must take effect": a Protocol base without `__slots__ = ()` gives every subclass a `__dict__`, so `ChannelManagerImpl`'s slots were dead. The base is slotted for that reason, not to slot the whole hierarchy.

You're right that the `Fake*` siblings in the same files were left inconsistent — `FakeChannelManager` and `FakeRabbitDeclarer` are slotted in 744a7ff6fa.

The remaining 26 unslotted subclasses of these bases are the telemetry/prometheus middlewares (12) and the settings providers (14). They hold state and sit on the per-message path, so they go separately, with measurements: #3201.

IvanKirpichnikov (approved):

PR #2841, opened 2026-04-22
feat: codec wiring unification

ce1ebrimbor:
# Codec Protocol + Subscriber Wiring

## What

Adds `CodecProto` — a stateful, async codec interface for encoding/decoding message bodies — and wires the decode side into the subscriber pipeline. This enables use cases like Avro + Schema Registry where the codec needs state (registry client), async network calls, and custom wire framing.

WIP #2837 

## Changes

- **`CodecProto`** protocol with `async encode()` and `async decode()` in `faststream/_internal/parser.py`
- **`DefaultCodec`** class wrapping existing `encode_message`/`decode_message` — zero behavior change for existing code
- **`codec=` parameter** on all 6 brokers' `__init__()` and `subscriber()` methods
- **Codec-aware decoder resolution** in `_get_parser_and_decoder()` — when `codec=` is set, `codec.decode` replaces the decoder callable
- **`ValueError`** when both `codec=` and `decoder=` are provided (ambiguous)
- **Broker parser type aliases** (`KafkaParserType`, `NatsParserType`, etc.) exported from each broker's `__init__.py`
- **`CodecTestcase`** base class + tests for all 6 brokers (24 tests total)

## What's NOT in this PR

- `codec.encode()` is defined but not wired into producers yet — production-side is the next PR
- No deprecation of the existing `decoder=` parameter

## Usage

```python
from faststream.kafka import KafkaBroker
from faststream._internal.parser import DefaultCodec

class MyCodec(DefaultCodec):
    async def decode(self, msg):
        # custom decode logic (e.g. Avro schema registry lookup)
        return await super().decode(msg)

broker = KafkaBroker(codec=MyCodec())

# or per-subscriber:
@broker.subscriber("topic", codec=MyCodec())
async def handler(msg): ...
```

PR #3219, opened 2026-09-22
fix: give Route and the publisher object the options the broker takes

Lancetnik:
The Option surface rule (#2871) says an option exists at every level it can reasonably be set. Nothing checked it: mypy over `faststream/` cannot see a *missing* keyword — that is a call-site error, which is why `tests/mypy/*` exists and is written by hand.

## The check

For each broker, compare the parameter names of `broker.subscriber` / `broker.publisher` against the router, the `Route` object and the publisher object, and assert the four results at once (`tests/brokers/test_option_surface.py`).

`Broker.subscriber` and `Router.subscriber` are the same registrator function (`is` holds for all six brokers and both methods), so those two rows cannot fail today. They stay in the assertion as a guard: the moment a router gets a signature of its own that drops an option, the test says so.

## What it found, and this PR fixes

`Route` and the publisher objects are hand-written copies of the registrator arguments, and they were behind:

| broker | added to `Route` | added to the publisher object |
|---|---|---|
| kafka | `codec`, `persistent` | `autoflush`, `persistent` |
| confluent | `codec`, `persistent` | `autoflush`, `persistent` |
| nats | `codec`, `persistent` | `persistent` |
| rabbit | `channel`, `codec`, `persistent` | `persistent` |
| redis | `codec`, `message_format`, `persistent` | `message_format`, `persistent` |
| mqtt | `codec` | — |

`codec` came in with #2841 and `persistent` with the 0.7.0 work; both were added to the registrator and not to the objects beside it. Each parameter keeps the registrator's own type, default and docstring line, and is forwarded to it — nothing else changes.

Before this, `KafkaRoute(handler, "topic", persistent=False)` raised `TypeError`. Now it registers a non-persistent subscriber: with `persistent=False` the registrator keeps 0 subscribers and 0 publishers, with `persistent=True` it keeps 1 and 1 — the same as `broker.subscriber(..., persistent=False)`.

## Checks

The test goes red on a real regression: adding a parameter to the Kafka registrator made the `Route` row grow it.

`mypy` (1850 files), `pyright` (0 errors), `ruff`, `ruff format`, `slotscheck`, both collection gates, and the router / include-router tests of all six brokers plus this file (275 passed).


tests/brokers/test_option_surface.py
IvanKirpichnikov, line 1: Let’s make smaller tests of this type instead of one generic test.

```py
@pytest.mark.nats
def test_every_level_takes_the_options_the_broker_takes() -> None:
    assert_test_every_level_takes_the_options_the_broker_takes(NatsBroker, NatsRouter, NatsRoute, NasPublisher)

def assert_test_every_level_takes_the_options_the_broker_takes(
    broker, router, route, publisher,
) -> None:
    assert _lost(broker.subscriber, router.subscriber) == []
    assert _lost(broker.publisher, router.publisher) == []
    assert _lost(broker.subscriber, route) == []
    assert _lost(broker.publisher, publisher) == []
```

IvanKirpichnikov (changes requested):

tests/brokers/test_option_surface.py
Lancetnik: Done in 5cdafd0: each broker has its own small `test_option_surface.py` next to its other tests, and the shared helper lives in `tests/brokers/base/option_surface.py`. The classes are imported by name now, not looked up by string.

One thing I kept: the helper still compares the four levels in one equality instead of four `assert`s. Separate asserts stop at the first mismatch, while the dict shows every level that lost an option in one failure.

IvanKirpichnikov (approved):

PR #3220, opened 2026-09-22
chore(lint): make every declared __slots__ take effect in subclasses too

### LINKED ISSUE #3201: Keep `__slots__` working in subclasses of slotted classes (middlewares, settings providers)
#3200 made every declared `__slots__` take effect by slotting the bases
(`require-superclass`). The other direction was left out of scope: a subclass of a slotted
class that declares no `__slots__` of its own gets a `__dict__` back, so the saving stops at
the first subclass. The ones that matter are created per message or per broker and sit right
under bases #3200 just slotted:

- broker middlewares: `BaseMiddleware` itself, `AcknowledgementMiddleware`,
  `CriticalLogMiddleware`, every `<Broker>TelemetryMiddleware` and
  `<Broker>PrometheusMiddleware` — a middleware object is built for each consumed message;
- settings providers: the 14 `*TelemetrySettingsProvider` / `*MetricsSettingsProvider`
  subclasses under the two `Protocol` bases that got `__slots__ = ()`.

That is 29 of the findings. slotscheck's switch for this is `require-subclass = true`, but it
is broader than the name says: `object` counts as a slotted base, so it reports **every** class
that *could* have slots — 190 today, all "has no slots, but it could have". By package:
specification 74, `_internal` 31, confluent 18, redis 14, nats 13, rabbit 11, kafka 11, mqtt 6,
asgi 6, the rest 6. Most of those are AsyncAPI schema models, configs and application objects
where slots buy nothing and would only forbid attributes. The raw list is at the bottom.

So the work is a decision first, then a fix:

1. Pick the target set. Proposal: the middleware and provider hierarchies above, plus any other
   class instantiated per message (check `StreamMessage` subclasses, `PublishCommand`s, the
   `Response` family). Everything else stays unslotted on purpose.
2. Express it in `[tool.slotscheck]` — `require-subclass = true` with `include-classes` (or
   `exclude-classes`) as a regex, so the rule holds for new subclasses of the target set and
   stays silent elsewhere. The config carries a two-line comment on why the set is what it is.
3. Slot the target classes. Traps are the ones from #3200: names assigned in `__init__` must be
   in the class's own `__slots__` and not repeated from a base; no class-level default with a
   slot's name; `functools.cached_property` needs a `__dict__`; mixins stay `__slots__ = ()`;
   two bases with non-empty slots conflict.

Public behaviour: a user's own middleware subclass is not checked and keeps its `__dict__`
unless it declares slots, so user code does not break by subclassing. What changes is that the
library's own middleware and provider instances stop accepting arbitrary attributes — list the
classes in the PR under a behaviour-change note, as #3200 did.

Worth measuring before and after, since the claim is a hot-path one: bytes per
`BaseMiddleware` instance and the per-message numbers from the hot-path benchmark.

`KvWatch` declares a `bucket` slot that nothing assigns (found in #3200, left alone there).
Either assign it or drop the name — same PR is fine.

**Blocked by** #3200 (it adds slotscheck and slots the bases this builds on).

## Done when

- [ ] The target set is written down and agreed: which hierarchies are slotted, which are not, and why
- [ ] `[tool.slotscheck]` enables `require-subclass` scoped to that set; `just slotscheck` is green
- [ ] A new unslotted subclass inside the target set turns `just slotscheck` red (bite test, reverted)
- [ ] Instances of every class in the target set have no `__dict__`
- [ ] No class outside the target set gained `__slots__`
- [ ] `KvWatch.bucket` is assigned or removed
- [ ] Named test files for the touched middlewares and providers pass (`opentelemetry`, `prometheus`, acknowledgement and logging middleware tests), `-m "not connected and not slow"`
- [ ] Per-instance size and per-message overhead reported before and after
- [ ] PR lists the library classes that no longer accept arbitrary attributes
- [ ] `just lint`, `just mypy`, `just pyright` clean

Counts measured on the #3200 branch with `uv run --frozen slotscheck -m faststream --require-subclass`.

<details>
<summary>All 190 <code>require-subclass</code> findings</summary>

```python
ERROR: 'faststream._internal.application:StartAbleApplication' has no slots, but it could have.
ERROR: 'faststream._internal.broker.pub_base:BrokerPublishMixin' has no slots, but it could have.
ERROR: 'faststream._internal.broker.registrator:Registrator' has no slots, but it could have.
ERROR: 'faststream._internal.cli.supervisors.asgi_multiprocess:ASGIMultiprocess' has no slots, but it could have.
ERROR: 'faststream._internal.cli.supervisors.basereload:BaseReload' has no slots, but it could have.
ERROR: 'faststream._internal.cli.supervisors.utils:ExitEvent' has no slots, but it could have.
ERROR: 'faststream._internal.configs.broker:BrokerConfig' has no slots, but it could have.
ERROR: 'faststream._internal.configs.broker:ConfigComposition' has no slots, but it could have.
ERROR: 'faststream._internal.configs.endpoint:EndpointConfig' has no slots, but it could have.
ERROR: 'faststream._internal.configs.specification:SpecificationConfig' has no slots, but it could have.
ERROR: 'faststream._internal.constants:EmptyPlaceholder' has no slots, but it could have.
ERROR: 'faststream._internal.context.context_type:Context' has no slots, but it could have.
ERROR: 'faststream._internal.context.repository:ContextRepo' has no slots, but it could have.
ERROR: 'faststream._internal.di.config:BuiltDependant' has no slots, but it could have.
ERROR: 'faststream._internal.di.config:FastDependsConfig' has no slots, but it could have.
ERROR: 'faststream._internal.endpoint.specification:EndpointSpecification' has no slots, but it could have.
ERROR: 'faststream._internal.endpoint.subscriber.utils:MultiLock' has no slots, but it could have.
ERROR: 'faststream._internal.endpoint.usecase:Endpoint' has no slots, but it could have.
ERROR: 'faststream._internal.endpoint.utils:ParserComposition' has no slots, but it could have.
ERROR: 'faststream._internal.fastapi._compat:SolvedDependency' has no slots, but it could have.
ERROR: 'faststream._internal.fastapi.config:FastAPIConfig' has no slots, but it could have.
ERROR: 'faststream._internal.fastapi.get_dependant:_FastStreamDependant' has no slots, but it could have.
ERROR: 'faststream._internal.logger.state:LoggerState' has no slots, but it could have.
ERROR: 'faststream._internal.middlewares:BaseMiddleware' has no slots, but it could have.
ERROR: 'faststream._internal.parser:DefaultCodec' has no slots, but it could have.
ERROR: 'faststream._internal.testing.app:TestApp' has no slots, but it could have.
ERROR: 'faststream._internal.testing.broker:TestBroker' has no slots, but it could have.
ERROR: 'faststream._internal.testing.calls:CallRecorder' has no slots, but it could have.
ERROR: 'faststream._internal.utils.functions:FakeContext' has no slots, but it could have.
ERROR: 'faststream._internal.utils.nuid:NUID' has no slots, but it could have.
ERROR: 'faststream._internal.utils.path:AddressSyntax' has no slots, but it could have.
ERROR: 'faststream.asgi.factories.asyncapi.route:AsyncAPIRoute' has no slots, but it could have.
ERROR: 'faststream.asgi.factories.asyncapi.try_it_out:TryItOutProcessor' has no slots, but it could have.
ERROR: 'faststream.asgi.handlers:HttpHandler' has no slots, but it could have.
ERROR: 'faststream.asgi.request:AsgiRequest' has no slots, but it could have.
ERROR: 'faststream.asgi.response:AsgiResponse' has no slots, but it could have.
ERROR: 'faststream.asgi.websocket:WebSocketClose' has no slots, but it could have.
ERROR: 'faststream.confluent.broker.router:KafkaPublisher' has no slots, but it could have.
ERROR: 'faststream.confluent.broker.router:KafkaRoute' has no slots, but it could have.
ERROR: 'faststream.confluent.configs.broker:ConsumerBuilder' has no slots, but it could have.
ERROR: 'faststream.confluent.helpers.admin:AdminService' has no slots, but it could have.
ERROR: 'faststream.confluent.helpers.admin:CreateResult' has no slots, but it could have.
ERROR: 'faststream.confluent.helpers.client:AsyncConfluentConsumer' has no slots, but it could have.
ERROR: 'faststream.confluent.helpers.client:AsyncConfluentProducer' has no slots, but it could have.
ERROR: 'faststream.confluent.helpers.client:BatchBuilder' has no slots, but it could have.
ERROR: 'faststream.confluent.helpers.config:ConfluentFastConfig' has no slots, but it could have.
ERROR: 'faststream.confluent.message:FakeConsumer' has no slots, but it could have.
ERROR: 'faststream.confluent.opentelemetry.middleware:KafkaTelemetryMiddleware' has no slots, but it could have.
ERROR: 'faststream.confluent.opentelemetry.provider:BatchConfluentTelemetrySettingsProvider' has no slots, but it could have.
ERROR: 'faststream.confluent.opentelemetry.provider:ConfluentTelemetrySettingsProvider' has no slots, but it could have.
ERROR: 'faststream.confluent.parser:AsyncConfluentParser' has no slots, but it could have.
ERROR: 'faststream.confluent.prometheus.middleware:KafkaPrometheusMiddleware' has no slots, but it could have.
ERROR: 'faststream.confluent.prometheus.provider:BatchConfluentMetricsSettingsProvider' has no slots, but it could have.
ERROR: 'faststream.confluent.prometheus.provider:ConfluentMetricsSettingsProvider' has no slots, but it could have.
ERROR: 'faststream.confluent.testing:MockConfluentMessage' has no slots, but it could have.
ERROR: 'faststream.kafka.broker.router:KafkaPublisher' has no slots, but it could have.
ERROR: 'faststream.kafka.broker.router:KafkaRoute' has no slots, but it could have.
ERROR: 'faststream.kafka.message:FakeConsumer' has no slots, but it could have.
ERROR: 'faststream.kafka.opentelemetry.middleware:KafkaTelemetryMiddleware' has no slots, but it could have.
ERROR: 'faststream.kafka.opentelemetry.provider:BatchKafkaTelemetrySettingsProvider' has no slots, but it could have.
ERROR: 'faststream.kafka.opentelemetry.provider:KafkaTelemetrySettingsProvider' has no slots, but it could have.
ERROR: 'faststream.kafka.parser:AioKafkaParser' has no slots, but it could have.
ERROR: 'faststream.kafka.prometheus.middleware:KafkaPrometheusMiddleware' has no slots, but it could have.
ERROR: 'faststream.kafka.prometheus.provider:BatchKafkaMetricsSettingsProvider' has no slots, but it could have.
ERROR: 'faststream.kafka.prometheus.provider:KafkaMetricsSettingsProvider' has no slots, but it could have.
ERROR: 'faststream.kafka.testing:FakeConsumer' has no slots, but it could have.
ERROR: 'faststream.message.message:StreamMessage' has no slots, but it could have.
ERROR: 'faststream.middlewares.acknowledgement.middleware:AcknowledgementMiddleware' has no slots, but it could have.
ERROR: 'faststream.middlewares.logging:CriticalLogMiddleware' has no slots, but it could have.
ERROR: 'faststream.mqtt.broker.router:MQTTPublisher' has no slots, but it could have.
ERROR: 'faststream.mqtt.broker.router:MQTTRoute' has no slots, but it could have.
ERROR: 'faststream.mqtt.opentelemetry.middleware:MQTTTelemetryMiddleware' has no slots, but it could have.
ERROR: 'faststream.mqtt.parser:MQTTBaseParser' has no slots, but it could have.
ERROR: 'faststream.mqtt.prometheus.middleware:MQTTPrometheusMiddleware' has no slots, but it could have.
ERROR: 'faststream.mqtt.testing:_BlockingSubscription' has no slots, but it could have.
ERROR: 'faststream.nats.broker.router:NatsPublisher' has no slots, but it could have.
ERROR: 'faststream.nats.broker.router:NatsRoute' has no slots, but it could have.
ERROR: 'faststream.nats.broker.state:BrokerState' has no slots, but it could have.
ERROR: 'faststream.nats.helpers.bucket_declarer:KVBucketDeclarer' has no slots, but it could have.
ERROR: 'faststream.nats.helpers.obj_storage_declarer:OSBucketDeclarer' has no slots, but it could have.
ERROR: 'faststream.nats.opentelemetry.middleware:NatsTelemetryMiddleware' has no slots, but it could have.
ERROR: 'faststream.nats.opentelemetry.provider:NatsBatchTelemetrySettingsProvider' has no slots, but it could have.
ERROR: 'faststream.nats.opentelemetry.provider:NatsTelemetrySettingsProvider' has no slots, but it could have.
ERROR: 'faststream.nats.parser:NatsBaseParser' has no slots, but it could have.
ERROR: 'faststream.nats.prometheus.middleware:NatsPrometheusMiddleware' has no slots, but it could have.
ERROR: 'faststream.nats.prometheus.provider:BatchNatsMetricsSettingsProvider' has no slots, but it could have.
ERROR: 'faststream.nats.prometheus.provider:NatsMetricsSettingsProvider' has no slots, but it could have.
ERROR: 'faststream.nats.schemas.schedule:Schedule' has no slots, but it could have.
ERROR: 'faststream.opentelemetry.consts:MessageAction' has no slots, but it could have.
ERROR: 'faststream.params.no_cast:NoCastField' has no slots, but it could have.
ERROR: 'faststream.rabbit.broker.router:RabbitPublisher' has no slots, but it could have.
ERROR: 'faststream.rabbit.broker.router:RabbitRoute' has no slots, but it could have.
ERROR: 'faststream.rabbit.configs.base:RabbitConfig' has no slots, but it could have.
ERROR: 'faststream.rabbit.helpers.channel_manager:FakeChannelManager' has no slots, but it could have.
ERROR: 'faststream.rabbit.helpers.declarer:FakeRabbitDeclarer' has no slots, but it could have.
ERROR: 'faststream.rabbit.opentelemetry.middleware:RabbitTelemetryMiddleware' has no slots, but it could have.
ERROR: 'faststream.rabbit.parser:AioPikaParser' has no slots, but it could have.
ERROR: 'faststream.rabbit.prometheus.middleware:RabbitPrometheusMiddleware' has no slots, but it could have.
ERROR: 'faststream.rabbit.publisher.producer:_RPCCallback' has no slots, but it could have.
ERROR: 'faststream.rabbit.schemas.channel:Channel' has no slots, but it could have.
ERROR: 'faststream.rabbit.testing:PatchedMessage' has no slots, but it could have.
ERROR: 'faststream.redis.broker.router:RedisPublisher' has no slots, but it could have.
ERROR: 'faststream.redis.broker.router:RedisRoute' has no slots, but it could have.
ERROR: 'faststream.redis.configs.state:ConnectionState' has no slots, but it could have.
ERROR: 'faststream.redis.configs.state:_SyncPubSubProxy' has no slots, but it could have.
ERROR: 'faststream.redis.opentelemetry.middleware:RedisTelemetryMiddleware' has no slots, but it could have.
ERROR: 'faststream.redis.parser.binary:BinaryMessageFormatV1' has no slots, but it could have.
ERROR: 'faststream.redis.parser.binary:BinaryReader' has no slots, but it could have.
ERROR: 'faststream.redis.parser.binary:BinaryWriter' has no slots, but it could have.
ERROR: 'faststream.redis.parser.parsers:SimpleParser' has no slots, but it could have.
ERROR: 'faststream.redis.prometheus.middleware:RedisPrometheusMiddleware' has no slots, but it could have.
ERROR: 'faststream.redis.prometheus.provider:BatchRedisMetricsSettingsProvider' has no slots, but it could have.
ERROR: 'faststream.redis.prometheus.provider:RedisMetricsSettingsProvider' has no slots, but it could have.
ERROR: 'faststream.redis.testing:Entry' has no slots, but it could have.
ERROR: 'faststream.redis.testing:PEL' has no slots, but it could have.
ERROR: 'faststream.response.response:Response' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.amqp.channel:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.amqp.channel:Exchange' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.amqp.channel:Queue' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.amqp.operation:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.http.operation:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.kafka.channel:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.kafka.operation:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.main.channel:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.main.operation:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.mqtt.channel:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.mqtt.operation:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.nats.channel:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.nats.operation:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.redis.channel:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.redis.operation:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.sqs.channel:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.bindings.sqs.operation:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.channels:Channel' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.components:Components' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.contact:Contact' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.docs:ExternalDocs' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.info:ApplicationInfo' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.license:License' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.message:CorrelationId' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.message:Message' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.operations:Operation' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.schema:ApplicationSchema' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.servers:Server' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.servers:ServerVariable' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.tag:Tag' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.utils:Parameter' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v2_6_0.schema.utils:Reference' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v3_0_0.schema.bindings.amqp.channel:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v3_0_0.schema.bindings.amqp.operation:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v3_0_0.schema.bindings.http:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v3_0_0.schema.bindings.main.channel:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v3_0_0.schema.bindings.main.operation:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v3_0_0.schema.channels:Channel' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v3_0_0.schema.components:Components' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v3_0_0.schema.info:ApplicationInfo' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v3_0_0.schema.operations:Operation' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v3_0_0.schema.schema:ApplicationSchema' has no slots, but it could have.
ERROR: 'faststream.specification.asyncapi.v3_0_0.schema.servers:Server' has no slots, but it could have.
ERROR: 'faststream.specification.base.info:BaseApplicationInfo' has no slots, but it could have.
ERROR: 'faststream.specification.base.schema:BaseApplicationSchema' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.amqp:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.amqp:Exchange' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.amqp:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.amqp:Queue' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.http:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.kafka:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.kafka:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.main:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.main:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.mqtt:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.mqtt:LastWillBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.mqtt:MessageBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.mqtt:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.mqtt:ServerBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.nats:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.nats:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.redis:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.redis:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.sqs:ChannelBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.bindings.sqs:OperationBinding' has no slots, but it could have.
ERROR: 'faststream.specification.schema.broker:BrokerSpec' has no slots, but it could have.
ERROR: 'faststream.specification.schema.extra.contact:Contact' has no slots, but it could have.
ERROR: 'faststream.specification.schema.extra.external_docs:ExternalDocs' has no slots, but it could have.
ERROR: 'faststream.specification.schema.extra.license:License' has no slots, but it could have.
ERROR: 'faststream.specification.schema.extra.tag:Tag' has no slots, but it could have.
ERROR: 'faststream.specification.schema.message.model:Message' has no slots, but it could have.
ERROR: 'faststream.specification.schema.operation.model:Operation' has no slots, but it could have.
ERROR: 'faststream.specification.schema.publisher:PublisherSpec' has no slots, but it could have.
ERROR: 'faststream.specification.schema.subscriber:SubscriberSpec' has no slots, but it could have.
```

</details>

Lancetnik:
Fixes #3201.

#3200 slotted the bases so declared `__slots__` take effect. This is the other
direction: a subclass that declares nothing of its own gets a `__dict__` back, so the
saving stopped at the first subclass. `slotscheck --require-subclass` reported 188 such
classes on `main`. They are all dealt with here — slotted where a class can be,
excluded with a written reason where it cannot.

**427 of the package's 662 classes now carry no `__dict__` at all, up from 87.**

## How CI holds it

`require-subclass` is on globally in `[tool.slotscheck]`, so `just slotscheck` is one
run again. The classes that stay unslotted are listed in `exclude-classes`, a verbose
regex where each entry carries a `#` comment with its reason — the config is the record
of what is deliberately not slotted:

```toml
require-subclass = true
exclude-classes = '''
  ^faststream\.specification\.(asyncapi\.v[0-9_]+\.schema|base|schema\.bindings\.http)
      # pydantic models: the fields live in BaseModel's own `__dict__` slot and the
      # AsyncAPI ones accept extra keys, so declaring `__slots__` would save nothing
| ^faststream\._internal\.application:Application$
      # the user's application object, one per process; our CLI tests patch `app.run`
...
'''
```

Bite tests — each reverted afterwards:

- dropping `__slots__ = ()` from `AioKafkaBatchParser` →
  `ERROR: 'faststream.kafka.parser:AioKafkaBatchParser' has no slots, but it could have.`
- dropping `__slots__` from `_LoggingMiddleware` →
  `ERROR: 'faststream.middlewares.logging:_LoggingMiddleware' has no slots, but it could have.`

No new pytest case: `slotscheck` in CI is the check.

## What got slotted

340 classes across the package (87 declared `__slots__` before, 427 do now). By area:

- **Per message** — the ones the issue is about: `BaseMiddleware` and every library
  middleware under it, the middleware factories, the 12 broker telemetry/prometheus
  middlewares, the 12 leaf settings providers; `StreamMessage` and all 16 broker message
  classes; `Response`, `PublishCommand`, `BatchPublishCommand` and every broker's
  response/command; every parser and codec.
- **Per declaration** — `Endpoint`, `SubscriberUsecase`, `PublisherUsecase`,
  `EndpointSpecification` and every broker's subscriber, publisher, specification and
  producer; `Registrator`, `BrokerRouter`, the router `ArgsContainer`/`SubscriberRoute`
  subclasses.
- **Standalone dataclasses** — the 26 models under `faststream/specification/schema/`,
  plus `AddressSyntax`, `Schedule`, `Channel`, `Entry`, `ConsumerBuilder`,
  `CreateResult`, `FastDependsConfig`, `SolvedDependency` and the rest, via
  `@dataclass(slots=True)`. Dataclasses that *inherit* one do not — see below.
- **Everything else** — `ContextRepo`, `Context`, `LoggerState` and the logger proxies,
  `NUID`, `MultiLock`, `ParserComposition`, `TestApp`, `TestBroker` and the six
  `Test<Broker>Broker`, the CLI supervisors, the ASGI request/response/handlers,
  the NATS and Redis helpers, the testing fakes.

Two changes needed more than a `__slots__` line:

- **`StreamMessage.__copy__`** copied through `self.__dict__`. It now walks the class's
  slots, so a broker's own (`KafkaMessage.consumer`) still come along and a slot never
  assigned stays unset on the copy. `__copy__` is only reached from the test client.
- **`StreamMessage.background`** is a declared slot. The FastAPI plugin parks its
  `BackgroundTasks` on the message for `_BackgroundMiddleware` to run
  (`_internal/fastapi/route.py`, already carrying a `# type: ignore[attr-defined]`).
  Declaring it changes nothing observable: unset, `getattr(msg, "background", None)`
  still returns `None`.
- **`Endpoint` declares `__weakref__`.** `Registrator` keeps subscribers and publishers
  in a `WeakSet`, which a fully slotted class cannot join otherwise.

**Protocols were the hidden half.** 26 `Protocol` classes had no `__slots__ = ()`, which
silently gave every implementor a `__dict__` — and neither slotscheck check reports a
Protocol, so nothing flagged them. `PublisherProto` was the biggest: it alone kept the
whole publisher chain unslotted. They all have it now, except `SpecificationFactory`,
whose five attributes are part of its interface and are assigned through it (a slotted
Protocol makes mypy reject that); `AsyncAPI` under it stays unslotted to match.

## What could not be slotted, and why

Nothing here is silenced — each is in `exclude-classes` with its reason, and the class
itself carries a docstring or comment saying the same.

| class | why |
|---|---|
| the AsyncAPI pydantic models (46) | fields live in `BaseModel`'s own `__dict__` slot, and `extra="allow"` means extra keys are stored dynamically. `__slots__ = ()` there would satisfy the checker and save nothing — measured identical, 256 bytes either way |
| `Application` | the user's application object, one per process; user code and our CLI tests assign to it (`patch.object(app, "run")`) |
| `BrokerUsecase` | one broker per process, and the object every suite mocks: `patch.object(broker, "start")` |
| `ConfigComposition` | a `__getattr__` proxy: reads forward to the inner config, writes land on the proxy itself (`patch.object(broker.config, "declarer")`) |
| `ConcurrentMixin` | mixed into each broker's slotted subscriber; two slotted bases under one `SubscriberUsecase` are an instance lay-out conflict |
| `kafka`/`confluent` `message:FakeConsumer` | the module-level `FAKE_CONSUMER` singleton, whose methods tests patch on the instance; a singleton saves nothing anyway |
| `kafka`/`confluent` `Default`/`BatchPublisher` | `create_publisher(autoflush=True)` wraps `_basic_publish*` on the instance |
| `NatsRegistrator` | `NatsRouter` mixes it with `BrokerRouter`, a slotted sibling under `Registrator` |
| NATS `LogicSubscriber` | `ConcurrentCoreSubscriber` mixes `TasksMixin` in beside it; NATS is the one broker whose subscribers do not inherit `TasksMixin` in a straight line |
| `RabbitConfig` | mixed into two slotted dataclass configs; the lay-outs conflict |
| Redis `ChannelSubscriber` | `RedisClusterBroker._make_channel_subscriber` replaces `start` on the instance |
| the config dataclass chain — `BrokerConfig`, `EndpointConfig`, `SpecificationConfig` and the ~30 broker configs under them | `@dataclass(slots=True)` is not portable across 3.10–3.14 for a dataclass that takes part in inheritance; see below |

The four lay-out conflicts are Python's rule, not a judgement: a class cannot inherit
from two bases that each add slots. Resolving them means restructuring those
hierarchies, which is a separate change.

### Why the config dataclasses are out

An earlier revision of this branch slotted them and CI went red on 3.10, 3.11 and
3.12 while 3.13 and 3.14 passed. Two version-gated `dataclasses` behaviours, both
invisible on the interpreter this was written on:

- **Below 3.13**, `slots=True` rebuilds the class, and a zero-argument `super()` in a
  method still points at the class *before* the rebuild. `KafkaBrokerConfig.__post_init__`
  and `MQTTBrokerConfig` raised `TypeError: super(type, obj): obj must be an instance or
  subtype of type` at import — 15 collection errors, which is what took `unused-fixtures`
  down with `test-basic` (it collects the whole suite on 3.12).
- **On 3.10**, `dataclasses._add_slots` re-emits every inherited field into the
  subclass's `__slots__`:

  ```
  (3, 10) B.__slots__ = ('x', 'y')
  (3, 11) B.__slots__ = ('y',)
  ```

  so each broker config duplicated its base's storage instead of saving it, and
  `slotscheck` on 3.10 reported overlapping slots on 22 classes — an error a
  contributor on 3.10 would hit locally even though CI's pre-commit runs 3.12.

So `slots=True` on a dataclass that takes part in inheritance is not portable across
the supported range. The chain is back to a plain `@dataclass` and its three roots are
in `exclude-classes`; standalone dataclasses keep `slots=True`, since neither problem
reaches them. The exclusion can come out when 3.10 support does.

## Before / after

`hasattr(obj, "__dict__")` and `sys.getsizeof(obj) + sys.getsizeof(obj.__dict__)`, the
same objects built the same way on `origin/main` and on this branch:

| class | `__dict__` before | bytes before | `__dict__` after | bytes after |
|---|---|---|---|---|
| `CriticalLogMiddleware` | True | 344 | False | 40 |
| `_LoggingMiddleware` | True | 344 | False | 64 |
| `AcknowledgementMiddleware` | True | 344 | False | 56 |
| `_AcknowledgementMiddleware` | True | 344 | False | 80 |
| `KafkaTelemetryMiddleware` | True | 376 | False | 64 |
| `BaseTelemetryMiddleware` | True | 344 | False | 96 |
| `KafkaPrometheusMiddleware` | True | 384 | False | 72 |
| `BasePrometheusMiddleware` | True | 344 | False | 72 |
| `KafkaTelemetrySettingsProvider` | True | 344 | False | 40 |
| `KafkaMetricsSettingsProvider` | True | 344 | False | 40 |
| `KafkaMessage` | True | 344 | False | 160 |
| `KafkaResponse` | True | 344 | False | 72 |
| `KafkaPublishCommand` | True | 344 | False | 136 |
| `DefaultSubscriber` | True | 344 | False | 224 |
| `KafkaSubscriberSpecification` | True | 344 | False | 56 |
| `ChannelBinding` (spec dataclass) | True | 344 | False | 56 |
| `DefaultPublisher` (kafka) | True | 344 | **True** | 416 |

Per consumed message that is roughly 1.0 KB less on a plain Kafka broker (message,
publish command, logging and acknowledgement middlewares) and about 1.7 KB less with
OpenTelemetry and Prometheus on.

The last row is the one regression, and it is deliberate: `LogicPublisher` above it is
slotted but the leaf must keep a `__dict__` for autoflush, so that object grows 72 bytes.
A handful of publishers exist per application. Nothing else in the table went up.

There is no hot-path benchmark harness in the repo (`tests/` has no `benchmark`-marked
cases), so the throughput half of #3201's "worth measuring" is not reported here.

## Behaviour change

The slotted library classes no longer accept arbitrary attributes — setting an
undeclared name on an instance raises `AttributeError`. That covers the middlewares and
settings providers, the message classes, the response and publish-command family, the
parsers, the subscriber/publisher/specification usecases, the configs, the specification
dataclasses and the internal helpers listed above.

What is deliberately **not** affected:

- **A user's own subclass.** It declares no `__slots__`, so Python gives it a `__dict__`
  as before. Subclassing `BaseMiddleware`, `StreamMessage` or `Response` and assigning
  freely keeps working.
- **The objects users hold and mock** — the application, the broker, and the publishers
  autoflush wraps — are exactly the ones left unslotted.
- **No public attribute was removed** from any class. `KvWatch.__slots__` lost `"bucket"`,
  a name nothing ever assigned (the bucket name is `NameRequired.name`).

Two tests changed, both for the same reason — they assigned to a subscriber instance,
which a slotted subscriber has nowhere to put:

- `tests/brokers/mqtt/test_consume.py` patched a private method on the instance and
  now patches it on the class.
- `tests/brokers/base/testclient.py::test_broker_with_real_stops_fake_subscribers`
  (a `connected` shared testcase) put its `spy_decorator` on the fake subscriber. The
  spy now goes on the class, for the duration of the client exit — the moment the
  fake is stopped. It still bites: making `TestBroker` skip stopping its fakes turns
  it red on kafka and rabbit with `Expected mock to have been awaited once. Awaited 0
  times.`

`--detect-unused-slots` stays off, as it was. It was green when this branch only touched
the middlewares; it now reports `StreamMessage.background` and `_FastStreamDependant`'s
three, which are assigned from another module — it only reads the class body, so those
are false positives here.

## Checks

All from a clean tree on this branch:

- `just slotscheck` on **3.10, 3.11, 3.12, 3.13 and 3.14** → `All OK! Scanned 512 module(s), 662 class(es).` each
- `uv run --frozen mypy` → `Success: no issues found in 1849 source files`
- `uv run --frozen pyright` → `0 errors, 0 warnings, 0 informations`
- `uv run --frozen ruff check .` → `All checks passed!`; `ruff format --check .` → `1307 files already formatted`
- `uv run --frozen lint-imports` → `Contracts: 8 kept, 0 broken.`
- `just misplaced-marks` → clean
- the two collection gates exactly as CI runs them, on 3.12:
  `python -m pytest --collect-only -q -m "" -p tests.unused_fixtures` → exit 0, and
  the same with `-p tests.misplaced_marks` → exit 0
- `pytest -m "(slow and not connected) or not connected"`, the `test-basic` selection,
  on **3.10, 3.11 and 3.12** → all green (38 snapshots passed each), plus the
  `not slow and not connected` suite on 3.14
- the **connected** suites against local brokers, one per `-real` job's marker
  expression — rabbit, kafka, confluent, mqtt, nats, redis and redis-cluster → all
  green. Two failures along the way passed in isolation and are shared-broker flakes,
  not this change: `test_broker_context_rabbit` (green on a re-run) and one NATS test
  per full run — `origin/main` fails a different NATS one in the same run
  (`test_no_trace_context_create_process_span`), so that suite is flaky on `main` too

The whole in-memory suite is the named set here rather than a handful of files: the
change reaches `BaseMiddleware`, `StreamMessage`, `Response`, `Endpoint`, `Registrator`
and every broker's subscriber and publisher, so nothing narrower would cover it. It was
re-run after each batch of classes; four rounds of failures are what identified the
classes in the "could not be slotted" table.


faststream/specification/base/generator.py
```diff
@@ -10,6 +10,8 @@
 
 
 class SpecificationFactory(Protocol):
+    # No `__slots__ = ()`: the five names below are part of the interface and callers
+    # assign them, which a slotted Protocol would forbid.
```
IvanKirpichnikov, line 14: why?

The code below works smoothly.
```py
from typing import Protocol

class AInterface(Protocol):
    __slots__ = ()

    a: int

class AImpl(AInterface):
    __slots__ = ("a",)

    def __init__(self) -> None:
        self.a = 1

print(AImpl().a) # 1
```

faststream/redis/subscriber/usecases/channel_subscriber.py
```diff
@@ -34,6 +34,12 @@
 
 
 class ChannelSubscriber(LogicSubscriber):
+    """Unslotted on purpose.
+
+    `RedisClusterBroker._make_channel_subscriber` replaces `start` on the instance,
+    which needs a `__dict__` to put the replacement in.
+    """
```
IvanKirpichnikov: It’s worth adding a NOTE stating that `__slots__` can be added once we implement async pubsub.

pyproject.toml
```diff
@@ -503,6 +503,38 @@ skips = ["B101"]
...
+| ^faststream\.(kafka|confluent)\.message:FakeConsumer$
+      # the FAKE_CONSUMER singleton, whose methods tests patch on the instance
+| ^faststream\.(kafka|confluent)\.publisher\.usecase:(Default|Batch)Publisher$
+      # `create_publisher(autoflush=True)` wraps `_basic_publish*` on the instance
+| ^faststream\.nats\.broker\.registrator:NatsRegistrator$
+      # NatsRouter mixes it with BrokerRouter, a slotted sibling under Registrator
+| ^faststream\.nats\.subscriber\.usecases\.basic:LogicSubscriber$
+      # ConcurrentCoreSubscriber mixes TasksMixin in beside it; the lay-outs conflict
+| ^faststream\.rabbit\.configs\.base:RabbitConfig$
+      # mixed into two slotted dataclass configs; the lay-outs conflict
```
IvanKirpichnikov, line 532: Can we delete `RabbitConfig`? I don’t see it being used anywhere in our project except during inheritance.

faststream/specification/base/generator.py
Lancetnik: At runtime, yes. The problem is mypy: once the assignment goes through a variable typed as the Protocol, it rejects it. Your example with a function added:

```py
def set_title(f: AInterface) -> None:
    f.a = 2
```

```
error: Trying to assign name "a" that is not in "__slots__" of type "AInterface"  [misc]
```

We have exactly this in `faststream/_internal/fastapi/router.py:287-291`: `self.schema` is typed `SpecificationFactory`, and `title`, `description`, `version`, `contact` and `license` are assigned through it. So the Protocol stays unslotted.

faststream/redis/subscriber/usecases/channel_subscriber.py
Lancetnik: Added in 3b19aa9. #3131 already removes the instance-level `start` replacement on the 1.0.0 branch, so the NOTE names it: slot the class once that reaches `main`. The `exclude-classes` entry says the same.

pyproject.toml
Lancetnik: It can go, but it would not change anything for slots here: the whole config dataclass chain is excluded until 3.10 support is dropped, so the Rabbit configs stay unslotted with or without the mixin. I would inline `queue`/`exchange` into the two configs in a separate PR rather than this one.

faststream/redis/subscriber/usecases/channel_subscriber.py
Lancetnik: Update: that NOTE is gone again, because the class no longer needs to be an exception. In 8e1ad34, `Endpoint` declares `__dict__`, so every subscriber and publisher keeps one: users patch them on the instance in tests (`patch.object(publisher, "publish")`), and that raised on RabbitMQ/NATS/Redis/MQTT. That also covers the cluster `start` replacement, so `ChannelSubscriber` now has `__slots__ = ()` and its `exclude-classes` entry is gone, same as the Kafka autoflush publishers.

IvanKirpichnikov (approved):

pyproject.toml
Lancetnik: Follow-up: on 1.0 it does matter. With 3.10 gone, the config dataclasses can take `slots=True`, and then a mixin with its own fields cannot sit next to a slotted base. So `RabbitConfig` is removed in #3230 (260c962), and `queue`/`exchange` are declared on the four Rabbit configs directly.

PR #3221, opened 2026-09-22
test: cover the docs_src snippets no test imported

Lancetnik:
`tests/docs/test_snippets_are_tested.py` carried 71 snippets that no test imported. This writes tests for 68 of them, one family per commit, and shrinks `NOT_TESTED_YET` to three entries with a reason beside each. The check itself is untouched.

## What each family's test asserts

**`lifespan/{broker}/basic.py`** (5 brokers, in-memory) — the snippet reads a `.env` through `Settings` in an `on_startup` hook and puts it in the context. `TestApp(app, {"env": ""})` runs the hook and the test reads the host back out of `app.context`.

**`subscription/{broker}/msgspec_fields.py`, `msgspec_struct.py`** (6 brokers, in-memory) — the page's payload (`{"name": "John", "userId": 1}`) is published through `TestBroker`; the handler's own asserts on the renamed `user_id` fail the publish if `MsgSpecSerializer` is not doing the renaming, and the test checks the handler saw the message.

**`prometheus/{broker}.py`, `{broker}_asgi.py`** (6 brokers, in-memory) — a message through the snippet's broker has to reach the registry the snippet handed to the middleware, counted once as `faststream_published_messages_total` and once as `faststream_received_messages_total`. The ASGI variants then serve those counters from `/metrics` through a Starlette `TestClient`.

**`opentelemetry/{broker}_telemetry.py`** (6 brokers, in-memory) — these snippets are page fragments: `tracer_provider` comes from the code block printed above them, so the test executes each file with a real `TracerProvider` in its globals (`runpy.run_path(init_globals=...)`) and asserts the `create` / `publish` / `process` spans a published message leaves in an `InMemorySpanExporter`.

**`cli/{broker}/extra_options.py`** (6 brokers, in-memory) — `faststream run <snippet> --port 5000 --foo bar` goes through the real CLI, with `FastStream.run` swapped for one that starts the app against a `TestBroker`, and the captured output is the page's `Port: 5000` / `Foo: bar`.

**`cli/{broker}/worker_id.py`** (6 brokers, in-memory) — both runs the page describes go through the real CLI: `faststream run <snippet>` prints `Started in a single process`, and `faststream run <snippet> --workers 2` prints `Worker 0 started` / `Worker 1 started`. The supervisor's numbering (`start_process(worker_id=i)` writing `extra_options["worker_id"]`) is the real code; only the spawn is replaced, so the children run in the test process instead of being forked.

**`manual_run/{broker}_base_run.py`** (6 brokers, `connected`) — the page runs `await app.run()` as a regular async function. The test runs `main()`, waits for the app that blocking call started, pings the broker it connected, and calls `exit()` so `main()` returns.

**`subscription/{broker}/dynamic.py`, `dynamic_iter.py`** (5 brokers, `connected`) — the page says TestBroker does not support dynamic subscribers, so these run against real brokers. `dynamic.py` has to return the message it was sent. `dynamic_iter.py` never returns, so the test injects a middleware through the broker class the snippet builds and asserts what its `async for` body actually saw.

**One-offs**

- `asyncapi/serve.py` — the page's five shell commands are run through the CLI: `docs gen` writes the application's document as JSON and as YAML, and `docs serve` serves that same document from the app and from both exported files (`serve_app` is patched to capture what it was handed).
- `multiple_brokers/add_broker.py` — `add_broker` leaves the app holding the brokers a constructor call would have, which is exactly what the page claims.
- `index/dependencies_annotated.py` — the `Annotated` twin of the handler already covered next to it.
- `kafka/security/sasl_oauthbearer.py`, `confluent/security/custom_config.py` — the mechanism and the extra `config` reach the client the broker builds.

## Left in `NOT_TESTED_YET`

- `integrations/no_http_frameworks_integrations/aiogram.py` — `Bot("")` needs a real Telegram token and the module calls `asyncio.run(main())` at import, so it cannot even be imported. It could equally move to the coverage `omit` list beside its `http_frameworks_integrations` siblings, if you prefer that.
- `getting_started/subscription/confluent/dynamic.py` and `dynamic_iter.py` — a group-less Confluent consumer joins one shared default group (`faststream-consumer-group`) behind the broker's 3s initial rebalance delay. Measured locally, `dynamic.py` returned a message in 1 run out of 3 (its two windows are 3s each, and the whole call took 6-16s), and `dynamic_iter.py` in 2 out of 3. Neither can be made reliable without changing the snippets.

## Two docs fixes the tests found

- **`worker_id: int` was the wrong signature.** `faststream run main:app` seeds `extra_options={"worker_id": None, ...}` (`faststream/_internal/cli/main.py:158`) and only a `--workers N` child gets an index, so the hook all six `worker_id.py` snippets show could not run single-process at all — it failed with `Invalid value for '--worker_id': extra option in your application 'lifespan/on_startup' hook has a wrong type.` The annotation is now `int | None` (the signature the CLI actually calls) in all six snippets, and the body says which of the two runs it is in:

  ```python
  @app.on_startup
  async def startup(worker_id: int | None) -> None:
      if worker_id is None:
          print("Started in a single process")
      else:
          print(f"Worker {worker_id} started")
  ```

  `docs/includes/en/worker-id.md` highlights the new line range, and `getting-started/cli.md` gains the single-process command and its output beside the `--workers 3` sample it already had (that sample still prints `Worker 0/1/2 started`, so it is unchanged).
- `docs_src/getting_started/cli/redis/worker_id.py` built a `RabbitBroker` under the page's Redis tab. Writing its test found it; it now builds a `RedisBroker`.

## Verification

```
uv run --frozen pytest tests/docs -m "not slow and not connected"   # 388 passed, 3 skipped
uv run --frozen pytest tests/docs/test_snippets_are_tested.py       # 1 passed
uv run --frozen pytest tests/docs/getting_started/manual_run/test_base_run.py \
    tests/docs/getting_started/subscription/test_dynamic.py -m "not slow"   # 16 passed (real brokers)
uv run --frozen ruff check . && uv run --frozen ruff format --check .
uv run --frozen mypy tests/docs && uv run --frozen mypy docs/docs_src
uv run --frozen python -m pytest --collect-only -q -m "" -p tests.unused_fixtures   # exit 0
uv run --frozen python -m pytest --collect-only -q -m "" -p tests.misplaced_marks   # exit 0
```


docs/docs_src/getting_started/subscription/confluent/dynamic.py
```diff
@@ -2,7 +2,7 @@
 
 async def main():
     async with KafkaBroker() as broker:  # connect the broker
-        subscriber = broker.subscriber("test-topic", persistent=False)
+        subscriber = broker.subscriber("dynamic-topic-confluent", persistent=False)
```
IvanKirpichnikov, line 5: why `-confluent` suffix?

docs/docs_src/getting_started/subscription/confluent/dynamic_iter.py
```diff
@@ -2,7 +2,7 @@
 
 async def main():
     async with KafkaBroker() as broker:
-        subscriber = broker.subscriber("test-topic", persistent=False)
+        subscriber = broker.subscriber("dynamic-topic-confluent", persistent=False)
```
IvanKirpichnikov, line 5: why `-confluent` suffix?

IvanKirpichnikov (changes requested):

docs/docs_src/getting_started/subscription/confluent/dynamic_iter.py
Lancetnik: To not conflict with aiokafka ones

IvanKirpichnikov (approved):

